PR #49, opened 2022-05-15
feat: add user role field

boojack:
Introducing site owner: when you first enter your website, you will be asked to sign up as the owner. So the other user cannot be login without an account. And site owner creation of user accounts will be in a separate PR.
image

Related issue: https://github.com/justmemos/memos/issues/48#issuecomment-1126783427


boojack:
FYI @monlor

PR #53, opened 2022-05-15
feat: import data with json file

boojack:
In the previous memos, we provided an "Export data as JSON". Use the exported `.json` file, so that we can import the data into newest memos.
image

Related issue: https://github.com/justmemos/memos/issues/46, https://github.com/justmemos/memos/issues/36


rebron1900:
66666

PR #86, opened 2022-06-23
fix: content link style

whoistiger:
Before:
image

After:
image

### DIFF
--- a/web/src/less/memo-content.less
+++ b/web/src/less/memo-content.less
@@ -17,8 +17,7 @@
   }
 
   .link {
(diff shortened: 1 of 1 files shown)


web/src/less/memo-content.less
```diff
@@ -18,6 +18,7 @@
 
   .link {
     @apply inline-block text-blue-600 cursor-pointer underline hover:opacity-80;
+    width: 100%;
```
boojack, line 21: This might be better
```suggestion
    @apply inline-block text-blue-600 cursor-pointer underline break-all hover:opacity-80;
```

boojack (approved):

PR #104, opened 2022-07-06
fix: banner text click not work

hyoban:


web/src/services/locationService.ts
```diff
@@ -44,7 +44,14 @@ const locationService = {
   },
 
   clearQuery: () => {
-    store.dispatch(setQuery({}));
+    store.dispatch(
+      setQuery({
+        tag: undefined,
+        type: undefined,
```
boojack, line 50: There is also a `duration`
```suggestion
        type: undefined,
        duration: undefined,
```

boojack (commented):
LG otherwise.

boojack (approved):
LGTM. Thanks a lot.

PR #105, opened 2022-07-06
feat: personal memos page

hyoban:
showing memos without logging in can let us use it like a personal Twitter.

### DIFF
--- a/web/src/components/UserBanner.tsx
+++ b/web/src/components/UserBanner.tsx
@@ -1,6 +1,7 @@
 import * as api from "../helpers/api";
 import { useCallback, useEffect, useState } from "react";
 import MenuBtnsPopup from "./MenuBtnsPopup";
+import { getUserIdFromPath } from "../services/userService";
 import { locationService } from "../services";
 import toastHelper from "./Toast";
 import { useAppSelector } from "../store";
@@ -31,7 +32,7 @@ const UserBanner: React.FC<Props> = () => {
         });
       } else {
         api
-          .getUserNameById(Number(locationService.getState().pathname.slice(3)))
+          .getUserNameById(Number(getUserIdFromPath()))
           .then(({ data }) => {
             const { data: username } = data;
             setUsername(username);


boojack:
I think it's better to create **a new page** to show someone's **public** memos. Just like https://jike.city/xidong, wdyt?

hyoban:
My idea is to make it like Github, others can see what's public by default, and at the same time, we can hide some memos.

boojack:
But the current main page UI isn't good for visitor because of the editor and some action buttons. A new pure page to display public memos and some extra information might be better for viewer.

boojack:
> But the current main page UI isn't good for visitor because of the editor and some action buttons. A new pure page to display public memos and some extra information might be better for viewer.

After taking a deep look, I think it's good for viewing when hiding some extra elements.

image

boojack:
LGTM. Another question: how do we view other member's memos?

hyoban:
This is also a more troublesome issue that I have considered. At present, I let it display the owner's memos by default.

hyoban:
I think most people will choose to build their own memos. 
Allowing multiple users to be created doesn't seem very useful.

hyoban:
ok, I'll try to handle each user's homepage

hyoban:
@boojack maybe you can review the code when you are free

server/user.go
```diff
@@ -51,6 +51,26 @@ func (s *Server) registerUserRoutes(g *echo.Group) {
 		return nil
 	})
 
+	g.GET("/username/:id", func(c echo.Context) error {
```
boojack, line 54: `/user/:id` might be better

server/user.go
```diff
@@ -51,6 +51,26 @@ func (s *Server) registerUserRoutes(g *echo.Group) {
...
+
+		user, err := s.Store.FindUser(&api.UserFind{
+			ID: &id,
+		})
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to fetch user").SetInternal(err)
+		}
+
+		c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSONCharsetUTF8)
+		if err := json.NewEncoder(c.Response().Writer).Encode(composeResponse(user.Name)); err != nil {
```
boojack, line 68: Need to check the `user` is `nil` firstly.

store/db/seed/10001__user.sql
```diff
@@ -17,3 +17,23 @@ VALUES
...
+VALUES
+  (
+    102, 
+    'demo2@usememos.com',
+    'USER',
+    'Demo User',
+    'demo_open_id2',
+    -- raw password: secret
+    '$2a$14$ajq8Q7fbtFRQvXpdCq7Jcuy.Rx1h/L4J60Otx.gyNLbAYctGMJ9tK'
+  );
```
boojack, line 39:
```suggestion
  (
    102, 
    'jack@usememos.com',
    'USER',
    'Jack',
    'jack_open_id',
    -- raw password: secret
    '$2a$14$ajq8Q7fbtFRQvXpdCq7Jcuy.Rx1h/L4J60Otx.gyNLbAYctGMJ9tK'
  );
```

web/src/components/Memo.tsx
```diff
@@ -10,6 +10,7 @@ import showMemoCardDialog from "./MemoCardDialog";
 import showShareMemoImageDialog from "./ShareMemoImageDialog";
 import toastHelper from "./Toast";
 import "../less/memo.less";
+import { useAppSelector } from "../store";
```
boojack, line 13: The import order needs update, `.less` should be put in last.

web/src/helpers/api.ts
```diff
@@ -44,16 +44,20 @@ export function getUserList() {
   return axios.get<ResponseObject<User[]>>("/api/user");
 }
 
+export function getUserNameById(id: number) {
```
boojack, line 47: We can just write a `getUserById`, maybe some fields of user might be useful later not only his name.

web/src/helpers/api.ts
```diff
@@ -44,16 +44,20 @@ export function getUserList() {
...
+  return axios.get<ResponseObject<string>>(`/api/username/${id}`);
+}
+
 export function patchUser(userPatch: UserPatch) {
   return axios.patch<ResponseObject<User>>("/api/user/me", userPatch);
 }
 
-export function getMemoList() {
-  return axios.get<ResponseObject<Memo[]>>("/api/memo");
+export function getMemoList(userID?: number) {
```
boojack, line 55: In frontend, the id should be named in `camelCase`. So it should be `userId: UserId`.

web/src/helpers/api.ts
```diff
@@ -44,16 +44,20 @@ export function getUserList() {
...
+  return axios.get<ResponseObject<Memo[]>>(`/api/memo?userID=${userID ? userID : ""}`);
```
boojack, line 56:
```suggestion
  return axios.get<ResponseObject<Memo[]>>(`/api/memo${userID ? "?userID=" + userID : ""}`);
```

web/src/components/Memo.tsx
```diff
@@ -158,38 +160,40 @@ const Memo: React.FC<Props> = (props: Props) => {
...
-                  <span className="tip-text">{memo.pinned ? "Unpin" : "Pin"}</span>
-                </div>
-                <div className="btn" onClick={handleEditMemoClick}>
-                  <img className="icon-img" src="/icons/edit.svg" alt="" />
-                  <span className="tip-text">Edit</span>
-                </div>
-                <div className="btn" onClick={handleGenMemoImageBtnClick}>
-                  <img className="icon-img" src="/icons/share.svg" alt="" />
-                  <span className="tip-text">Share</span>
+        {user && (
```
boojack, line 163: We have a `userService` with some common functions. So I think we can use it to check the userId(in param) and the current userId(in backend session) is same or different. For example, create a new function `isVisitor(): boolean` in `userService` that will check the login and visitor status. With it, these checks will looks better.

web/src/components/MemoList.tsx
```diff
@@ -79,11 +79,26 @@ const MemoList: React.FC<Props> = () => {
...
+    const userId = !isNaN(Number(path)) ? Number(path) : undefined;
     memoService
-      .fetchAllMemos()
-      .then(() => {
-        setFetchStatus(false);
-        memoService.updateTagsState();
+      .fetchAllMemos(userId)
+      .then((data) => {
+        if (data.length === 0) {
+          locationService.replaceHistory("/");
```
boojack, line 88: Why needs to redirect to `/`?

web/src/components/MemosHeader.tsx
```diff
@@ -30,7 +30,9 @@ const MemosHeader: React.FC<Props> = () => {
     const now = Date.now();
     if (now - prevRequestTimestamp > 10 * 1000) {
       prevRequestTimestamp = now;
-      memoService.fetchAllMemos().catch(() => {
+      const path = locationService.getState().pathname.slice(1);
+      const userId = !isNaN(Number(path)) ? Number(path) : undefined;
```
boojack, line 34: Do these check in `userService`.

server/user.go
boojack, line 54: GET user by id would be more useful.

boojack (changes requested):

server/user.go
boojack: GET user by id would be more useful. But needs to remove the some sensitize fields. e.g. `OpenID`

hyoban: That's what I named it at first, but most of the user's information shouldn't be made public

boojack: Or `/user/:id/name`?

server/memo.go
```diff
@@ -60,7 +60,26 @@ func (s *Server) registerMemoRoutes(g *echo.Group) {
 	})
 
 	g.GET("/memo", func(c echo.Context) error {
-		userID := c.Get(getUserIDContextKey()).(int)
+		userID, ok := c.Get(getUserIDContextKey()).(int)
+		if !ok {
+			if c.QueryParam("userID") != "" {
+				var err error
+				userID, err = strconv.Atoi(c.QueryParam("userID"))
```
boojack, line 67:
```suggestion
				userID, err := strconv.Atoi(c.QueryParam("userID"))
```

server/memo.go
```diff
@@ -60,7 +60,26 @@ func (s *Server) registerMemoRoutes(g *echo.Group) {
...
+				}
+			} else {
+				ownerUserType := api.Owner
+				ownerUser, err := s.Store.FindUser(&api.UserFind{
+					Role: &ownerUserType,
+				})
+				if err != nil {
+					return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find owner user").SetInternal(err)
+				}
+				userID = ownerUser.ID
```
boojack, line 79: needs to check ownerUser is whether `nil`

server/user.go
```diff
@@ -51,6 +51,26 @@ func (s *Server) registerUserRoutes(g *echo.Group) {
...
+	g.GET("/user/:id/name", func(c echo.Context) error {
+		id, err := strconv.Atoi(c.Param("id"))
+		if err != nil {
+			return echo.NewHTTPError(http.StatusBadRequest, "Malformatted user id").SetInternal(err)
+		}
+
+		user, err := s.Store.FindUser(&api.UserFind{
+			ID: &id,
+		})
+		if err != nil || user == nil {
```
boojack, line 63: if user==nil, we should not return `http.StatusInternalServerError` but `user not found`.

web/src/components/Sidebar.tsx
```diff
@@ -8,6 +8,7 @@ import UsageHeatMap from "./UsageHeatMap";
 import ShortcutList from "./ShortcutList";
 import TagList from "./TagList";
 import "../less/siderbar.less";
+import { userService } from "../services";
```
boojack, line 11: The import order needs update.

web/src/helpers/api.ts
```diff
@@ -44,16 +44,20 @@ export function getUserList() {
...
-export function getMemoList() {
-  return axios.get<ResponseObject<Memo[]>>("/api/memo");
+export function getMemoList(userId?: number) {
+  return axios.get<ResponseObject<Memo[]>>(`/api/memo${userId ? "?userID=" + userId : ""}`);
 }
 
-export function getArchivedMemoList() {
-  return axios.get<ResponseObject<Memo[]>>("/api/memo?rowStatus=ARCHIVED");
+export function getArchivedMemoList(userId?: number) {
+  return axios.get<ResponseObject<Memo[]>>(`/api/memo?rowStatus=ARCHIVED${userId ? "?userID=" + userId : ""}`);
```
boojack, line 60:
```suggestion
  return axios.get<ResponseObject<Memo[]>>(`/api/memo?rowStatus=ARCHIVED${userId ? "&userID=" + userId : ""}`);
```

web/src/services/memoService.ts
```diff
@@ -10,21 +11,26 @@ const convertResponseModelMemo = (memo: Memo): Memo => {
   };
 };
 
+const getUserIdFromPath = () => {
```
boojack, line 14: Put this function into userService might be better

boojack (changes requested):

server/memo.go
hyoban: I can't modify it like this, the userID will be overwritten in the if area

web/src/services/memoService.ts
```diff
@@ -10,21 +11,26 @@ const convertResponseModelMemo = (memo: Memo): Memo => {
...
+  const path = locationService.getState().pathname.slice(1);
```
boojack, line 15: I think we could update the route to `/u/:userId`. It's more clearly, and we could create a new route `/m/:memoId` for viewing a memo with id in the future.

server/memo.go
boojack: Sorry, my bad

web/src/services/memoService.ts
boojack: And use a regex `/^\/u\/(\d+)/` to match the `:userId`

boojack: My meaning is about the frontend route not the backend API. The current looks like is `/:userId`, and the `/u/:userId` will be more suitable.

hyoban: sorry, i will correct it soon

hyoban:
@boojack maybe we need an eslint config for imports sort, or [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import)

web/src/components/UserBanner.tsx
```diff
@@ -1,26 +1,48 @@
-import { useCallback, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { useAppSelector } from "../store";
 import { locationService } from "../services";
 import MenuBtnsPopup from "./MenuBtnsPopup";
 import "../less/user-banner.less";
+import * as api from "../helpers/api";
+import toastHelper from "./Toast";
```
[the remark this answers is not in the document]
hyoban: maybe we need an eslint config for imports sort, or [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import)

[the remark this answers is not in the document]
boojack: LGTM, maybe you can use `yarn lint` in your local env.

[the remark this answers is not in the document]
hyoban: Personally, I will use [@antfu/eslint-config](https://github.com/antfu/eslint-config)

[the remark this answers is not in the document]
hyoban: if we only use eslint config, we can not auto-fix the order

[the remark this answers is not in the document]
boojack: I will take a look at eslint.

hyoban:
This is my first contribution to a little feat, I have learned a lot, thank you

PR #132, opened 2022-07-24
chore: update `memo` visibility field

boojack:


boojack:
Update `memo.visibility` value in database schema level. FYI @Kellermaan

PR #164, opened 2022-08-23
chore: update `word-break` style

boojack:


boojack:
@limdingwen FYI, updates the `word-break` styles globally.

PR #218, opened 2022-09-16
feat: add vietnamese

f97:


boojack:
Thanks for your contribution! xin chào việt nam 👋

PR #270, opened 2022-10-08
chore: update i18n for shortcut filter

YuJianghao:
Add i18n for resources and shortcuts.

For i18n with constant, I use i18n mark in constants, and render to text with `t` function in component. Wondering if there's a better solution.


boojack (approved):
LGTM

PR #309, opened 2022-10-18
feat: editor tab support

f97:
chrome-capture-2022-9-18

### DIFF
--- a/web/src/components/MemoEditor.tsx
+++ b/web/src/components/MemoEditor.tsx
@@ -79,11 +79,16 @@ const MemoEditor: React.FC = () => {
   const handleKeyDown = (event: React.KeyboardEvent) => {
     if (event.key === "Escape" && state.fullscreen) {
       handleFullscreenBtnClick();
-    } else if (event.key === "Enter" && (event.ctrlKey || event.metaKey)) {
+      return;
+    }
+    if (event.key === "Enter" && (event.ctrlKey || event.metaKey)) {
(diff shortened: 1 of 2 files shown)


web/src/components/MemoEditor.tsx
```diff
@@ -77,13 +77,16 @@ const MemoEditor: React.FC = () => {
...
-        handleFullscreenBtnClick();
-      }
-    } else if (event.key === "Enter" && (event.ctrlKey || event.metaKey)) {
+    if (event.key === "Escape" && state.fullscreen) {
+      handleFullscreenBtnClick();
+    }
+    if (event.key === "Enter" && (event.ctrlKey || event.metaKey)) {
       handleSaveBtnClick();
     }
+    if (event.key === "Tab") {
```
boojack, line 86: `if-else` is more understandable, wdyt?

```suggestion
    } else if (event.key === "Enter" && (event.ctrlKey || event.metaKey)) {
      handleSaveBtnClick();
    } else if (event.key === "Tab") {
```

hyoban: We need else if or return to avoid going into two branches

f97:
I still think if is easier to understand, but it's fine for me.

hyoban (approved):
Sorry for taking your time, more people think `if` would be better

image

f97:
@hyoban , Yeahhhh, I also just remembered I was missing `return`. Thanks....

PR #312, opened 2022-10-18
feat: 添加Memo列表按更新时间排序

Zeng1998:


boojack (changes requested):

web/src/components/Memo.tsx
```diff
@@ -187,7 +190,7 @@ const Memo: React.FC<Props> = (props: Props) => {
       <div className="memo-top-wrapper">
         <div className="status-text-container">
           <span className="time-text" onClick={handleShowMemoStoryDialog}>
-            {createdAtStr}
+            {user?.setting.memoSortOption === "created_ts" ? createdAtStr : `${t("common.update-on")} ${updatedAtStr}`}
```
hyoban, line 193: Not good for reading, maybe creating a new variable would be better

web/src/components/MemoList.tsx
```diff
@@ -70,6 +72,10 @@ const MemoList = () => {
 
   const pinnedMemos = shownMemos.filter((m) => m.pinned);
   const unpinnedMemos = shownMemos.filter((m) => !m.pinned);
+  pinnedMemos.sort((m1, m2) => (user?.setting.memoSortOption === "created_ts" ? m2.createdTs - m1.createdTs : m2.updatedTs - m1.updatedTs));
+  unpinnedMemos.sort((m1, m2) =>
+    user?.setting.memoSortOption === "created_ts" ? m2.createdTs - m1.createdTs : m2.updatedTs - m1.updatedTs
```
hyoban, line 77: Would it be better to extract the same logic into a function?

boojack (approved):
Thanks for your contribution!

PR #321, opened 2022-10-20
feat: inline code within link

### LINKED ISSUE #317: code nested in link text

no

Support code nested in link text, like what github do

e.g. 

```md
[Doc for `docker save`](https://docs.docker.com/engine/reference/commandline/save/)
```

render into:

[Doc for `docker save`](https://docs.docker.com/engine/reference/commandline/save/)

### Additional context

_No response_

YuJianghao:
close #317 

link with code:
image
pure link:
image

---

use `decoration-1` with tailwind (`text-decoration-thickness: 1px`)
without `decoration-1`:
image

### DIFF
--- a/web/src/labs/marked/index.ts
+++ b/web/src/labs/marked/index.ts
@@ -26,7 +26,7 @@ export const marked = (markdownStr: string, blockParsers = blockElementParserLis
   let matchedInlineParser = undefined;
   let matchedIndex = -1;
 
-  for (const parser of inlineElementParserList) {
+  for (const parser of inlineParsers) {
     if (parser.name === "plain text" && matchedInlineParser !== undefined) {
       continue;
     }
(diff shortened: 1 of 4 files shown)


web/src/labs/marked/index.ts
```diff
@@ -26,7 +26,7 @@ export const marked = (markdownStr: string, blockParsers = blockElementParserLis
   let matchedInlineParser = undefined;
   let matchedIndex = -1;
 
-  for (const parser of inlineElementParserList) {
+  for (const parser of inlineParsers) {
```
YuJianghao, line 29: @boojack Is this a mistake? If we use `inlineElementParserList` here, `inlineParsers` (which are passed in many files, like [Bold parser](https://github.com/usememos/memos/blob/main/web/src/labs/marked/parser/Bold.ts#L13)) will never work.

boojack: Thanks, good catch!

boojack (approved):
LGTM

PR #347, opened 2022-10-27
feat: case-insensitive search

boojack:


hdlineage:
can we have this back?

PR #356, opened 2022-10-28
feat: resource dialog enhancements

Zeng1998:
I have done two things: 

- increase the width of the dialog. 
- add a tooltip to show the full resource name/type while hovering.

before:

after:


boojack (approved):
LGTM

boojack:
I think we can hide the resource type field in the list view. WDYT?

PR #357, opened 2022-10-28
fix: typo

Zeng1998:


Zeng1998:
I'm not good at git... I want to split them into two PRs, is it possible?

boojack:
> I'm not good at git... I want to split them into two PRs, is it possible?

Merge main branch firstly.

PR #384, opened 2022-10-31
fix: empty memo list

Zeng1998:
image
image

### DIFF
--- a/web/src/services/memoService.ts
+++ b/web/src/services/memoService.ts
@@ -31,7 +31,10 @@ const memoService = {
       memoFind.creatorId = userService.getUserIdFromPath();
     }
     const { data } = (await api.getMemoList(memoFind)).data;
-    const fetchedMemos = data.map((m) => convertResponseModelMemo(m));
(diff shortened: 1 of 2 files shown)


web/src/services/memoService.ts
```diff
@@ -31,7 +31,10 @@ const memoService = {
       memoFind.creatorId = userService.getUserIdFromPath();
     }
     const { data } = (await api.getMemoList(memoFind)).data;
-    const fetchedMemos = data.map((m) => convertResponseModelMemo(m));
+    let fetchedMemos: Memo[] = [];
+    if (data) {
```
boojack, line 35: You mean the `data` could be undefined?

Zeng1998: Yes... I'm not sure if it's a expected behavior?

boojack: This is certainly an unwanted value. But I don't know how it happened, it should pass error instead of returning undefined.

boojack:
Fixed in https://github.com/usememos/memos/pull/390

PR #399, opened 2022-11-01
chore: pin button color

YuJianghao:
||pinned|unpinned|
|---|---|---|
|new|image|image|
|old|image|image|

The old unpinned Memo's pin button looks like a disabled button. Maybe we can change the color a bit.


boojack (approved):
LGTM

PR #411, opened 2022-11-04
chore: allow skip version

YuJianghao:
image
image
image

### DIFF
--- a/web/src/pages/Home.tsx
+++ b/web/src/pages/Home.tsx
@@ -1,7 +1,6 @@
-import { useEffect, useState } from "react";
+import { useEffect } from "react";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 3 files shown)


web/src/pages/Home.tsx
```diff
@@ -68,7 +35,9 @@ function Home() {
 
   return (
     <section className="page-wrapper home">
-      <div className="banner-wrapper">{state.shouldShowUpdateVersionBanner && <UpdateVersionBanner />}</div>
+      <div className="banner-wrapper">
+        <UpdateVersionBanner />
```
YuJianghao, line 39: I think banners should control themselves to show or not. Otherwise, the `Home` component will be full of banner-related logic.

PR #415, opened 2022-11-05
fix: delete related resources

Zeng1998:
An issue still exists: previous memo deletions did not modify the table `memo_resource` , so, I think those related resources should be deleted manually.


Zeng1998:
> An issue still exists: previous memo deletions did not modify the table `memo_resource` , so, I think those related resources should be deleted manually.

Or an automated script is required.

boojack:
I'd like to delete the related records in store level but not server.

Zeng1998:
> I'd like to delete the related records in store level but not server.

How is it now?

boojack (approved):

PR #426, opened 2022-11-07
fix: marked regex for full-width space

Zeng1998:
#425

### DIFF
--- a/web/src/labs/marked/parser/Blockquote.ts
+++ b/web/src/labs/marked/parser/Blockquote.ts
@@ -1,6 +1,6 @@
 import { escape } from "lodash";
 
-export const BLOCKQUOTE_REG = /^>\s+([\S ]+)(\n?)/;
+export const BLOCKQUOTE_REG = /^>\s+([\S \u3000]+)(\n?)/;
 
 const renderer = (rawStr: string): string => {
   const matchResult = rawStr.match(BLOCKQUOTE_REG);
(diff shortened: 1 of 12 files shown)


web/src/labs/marked/parser/Blockquote.ts
```diff
@@ -1,6 +1,6 @@
 import { escape } from "lodash";
 
-export const BLOCKQUOTE_REG = /^>\s+([\S ]+)(\n?)/;
+export const BLOCKQUOTE_REG = /^>\s+([\S \u3000]+)(\n?)/;
```
boojack, line 3: I think a better way is that we should replace all of the full width space `"　"` with normal space before rendering memo content. So that we don't need to handle a special character in parser level.

Zeng1998: This might be a way, but if we consider this issue #425 , I think he wants to use full-width spaces to align his text (Chinese characters).

So maybe not the best for him if we replace them with normal spaces.

boojack: Then we could use two half width space to replace the special full width space. WDYT

Zeng1998: In fact, in Chinese typography, the width of two half-width spaces is not equal to the width of one full-width space, because there in space between two characters. 

Like this 
image

Maybe it's a small problem, but not perfect I think.

PR #453, opened 2022-11-12
fix: image url extraction

Zeng1998:


Zeng1998:
- change to the global mode. `/\((.*?)\)/g`
- get the last match one. `slice(-1)[0]`
- remove the parentheses. `slice(-1,1)`

boojack (approved):

PR #461, opened 2022-11-13
feat: add visibility filter

Zeng1998:


web/src/less/create-shortcut-dialog.less
```diff
@@ -2,7 +2,7 @@
   @apply px-4;
 
   > .dialog-container {
-    @apply w-128 max-w-full;
+    @apply w-160 max-w-full;
```
boojack, line 5: Please take a look into the mobile display.

Zeng1998: I changed it back to `w-128`. 

I checked the mobile style and found another bug, I'm going to open another PR to deal with it.

image
image

boojack (approved):
LGTM

PR #523, opened 2022-11-21
feat: extend validation config max length

### LINKED ISSUE #519: Password too long?
### Describe the bug

Why is it a problem if my password is on the longer side? Shouldn't it be hashed anyways? It was only 35 characters long.

### Steps to reproduce

Try to change your password into the following string bitwarden gave me: `product-untitled-lure-dingy-scrabble`

_No response_

### LINKED ISSUE #522: Email: too long
Tried installing it couple of minutes ago because this seems promissing.
Docker intall worked like a charm.
Getting to register as host, but there is the error: Email: too long.

My mail isn't short but not that long either.  27chars at max.

Can this be fixed?

yardenshoham:
The previous limit was too short

* Closes #522
* Closes #519

### DIFF
--- a/web/src/pages/Auth.tsx
+++ b/web/src/pages/Auth.tsx
@@ -11,7 +11,7 @@ import "../less/auth.less";
 
 const validateConfig: ValidatorConfig = {
   minLength: 4,
-  maxLength: 100,
+  maxLength: 320,
   noSpace: true,
   noChinese: true,
 };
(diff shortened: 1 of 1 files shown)


web/src/pages/Auth.tsx
```diff
@@ -11,7 +11,7 @@ import "../less/auth.less";
 
 const validateConfig: ValidatorConfig = {
   minLength: 4,
-  maxLength: 24,
+  maxLength: 100,
```
boojack, line 14: How about `320`?

image

yardenshoham: Yes, that makes sense

boojack (approved):
LGTM

PR #541, opened 2022-11-22
changed few translations

dabenzel:
Hi,
I changed a few translations so that they would fit better.
I'm not a native english speaker.

If this PR will be merged, I'll may start translating into german.

### DIFF
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -3,7 +3,7 @@
     "about": "About",
     "email": "Email",
     "password": "Password",
-    "new-password": "New passworld",
+    "new-password": "New password",
     "repeat-new-password": "Repeat the new password",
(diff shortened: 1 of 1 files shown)


web/src/locales/en.json
```diff
@@ -35,11 +35,11 @@
     "title": "Title",
     "filter": "Filter",
     "tags": "Tags",
-    "yourself": "Yourself",
+    "yourself": "You",
```
boojack, line 38: We should use the same key and value in locale. So let's add a new field `"you": "You"` instead of changing `"yourself"` in a separate PR. 

So it's better to fix the existed typo in this PR. e.g. `passworld`

boojack:
Please add German translation in a separate PR. And you can take a look into #546

PR #567, opened 2022-11-25
fix: make tooltip text no-selecting

dmitriisurmin:
With a double click on edit/pin button, you can get selection on tooltip text. It doesn't look good.

https://user-images.githubusercontent.com/118022040/203890748-5901c054-6f83-4834-94c7-17ee84e4d404.mov

### DIFF
--- a/web/src/less/memo.less
+++ b/web/src/less/memo.less
@@ -62,6 +62,7 @@
 
             > .btn {
               @apply relative w-6 h-6 p-1 text-gray-600 cursor-pointer;
+              user-select: none;
 
               &:hover > .tip-text {
                 @apply block;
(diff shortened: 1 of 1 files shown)


web/src/less/memo.less
```diff
@@ -62,6 +62,7 @@
 
             > .btn {
               @apply relative w-6 h-6 p-1 text-gray-600 cursor-pointer;
+              user-select: none;
```
hyoban, line 65: We are using tailwind, so

```suggestion
              @apply relative w-6 h-6 p-1 text-gray-600 cursor-pointer select-none;
```

hyoban (changes requested):

boojack (approved):
LGTM

PR #581, opened 2022-11-26
fix: get markdown image from backend

Zeng1998:
src of the img tag is  `/o/get/image?url=http://localhost:3000/o/r/1/image.png`
but 
src in the markdown content is `['http://localhost:3000/o/r/1/image.png']`
 
So `findIndex` function cannot find the correct index.


boojack (approved):
LGTM

PR #593, opened 2022-11-26
fix: search bar dropdown disappearing

dmitriisurmin:
Dropdown was disappearing when the mouse was still on it.
Common tools bar has the same z-level, so it overlaps the dropdown.

https://user-images.githubusercontent.com/118022040/204094779-827e679b-4c3d-47a3-96c8-411a531bbccc.mov


boojack (approved):
Nice catch, LGTM

Issue #521, opened 2022-11-21
Dark Mode


ItsNoted:
### Is your feature request related to a problem?

White is bright.

### Describe the solution you'd like

Make a dark mode for better visual experience when browsing in the dark or for those with sensitive eyes.

Zeng1998:
Dark mode isn't supported yet, and I'm not sure if it is a high-priority need.

For now, you can try the following additional css (shared from the community)

```css
@media (prefers-color-scheme: dark) {
    html {
        filter: invert(90%) hue-rotate(180deg);
    }

    img, video, div[class*="language-"] {
        filter: invert(110%) hue-rotate(180deg);
        opacity: .8;
    }
}
```

ItsNoted:
> Dark mode isn't supported yet, and I'm not sure if it is a high-priority need.
> 
> For now, you can try the following additional css (shared from the community)
> 
> ```css
> @media (prefers-color-scheme: dark) {
>     html {
>         filter: invert(90%) hue-rotate(180deg);
>     }
> 
>     img, video, div[class*="language-"] {
>         filter: invert(110%) hue-rotate(180deg);
>         opacity: .8;
>     }
> }
> ```

It will do for now. It also darkens emojis and other things too. Strange but thank you!

Zeng1998:
I think the dark mode is in development and coming soon.

ItsNoted:
> I think the dark mode is in development and coming soon.

Great to hear, thank you!

tapionx:
Would be nice to have ☺️

mariushosting:
+1 for Dark mode switch.

DJScias:
As of #637 a lot of the app now works in dark theme.

However as of the main page update the fullscreen option is unstyled for dark theme.
I suppose that was to be expected, unless that was meant to be part of what one would define as "main pages"?

Regardless, little issue since this is still the dev mode. But figured I'd mention it just in case. 😊

![image](https://user-images.githubusercontent.com/439655/204565903-725b79b1-0c40-48cf-adc9-c2a449d9cc2e.png)

PR #604, opened 2022-11-27
feat: dark mode support for memo detail

hyoban:
part of #521

### DIFF
--- a/web/src/css/tailwind.css
+++ b/web/src/css/tailwind.css
@@ -34,8 +34,4 @@
   .input-text {
     @apply w-full px-3 py-2 leading-6 text-sm border rounded;
   }
-
-  .card-memo {
-    @apply flex flex-col justify-start items-start w-full p-4 pt-3 mt-2 bg-white dark:bg-zinc-800 rounded-lg border border-white dark:border-zinc-800 hover:border-gray-200 hover:dark:border-zinc-700;
-  }
 }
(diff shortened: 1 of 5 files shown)


web/src/css/tailwind.css
```diff
@@ -14,22 +14,28 @@
...
+    @apply btn-normal text-gray-600 border-none hover:shadow-none;
+  }
+  
+  .input-text {
+    @apply w-full px-3 py-2 leading-6 text-sm border rounded;
+  }
 
-.input-text {
-  @apply w-full px-3 py-2 leading-6 text-sm border rounded;
+  .card-memo {
```
boojack, line 38: As it's used in only two places, it doesn't look like a common util class. 

https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction

boojack (commented):
LGTM

web/src/css/tailwind.css
hyoban: But it looks like both need to be styled the same way, and also the cards in the home page are styled the same

boojack: We can create a less/module.css file including the same layout css codes. Then import it in the component files. 

The tailwind.css should save the basic class definitions which will be used in multiple files.

boojack (approved):
LGTM

PR #608, opened 2022-11-27
fix: shoutcut tag filter handle mutiple tags

Wujiao233:
when apply a shortcut using tag filter, only the first tag in content can be found.

For example..
截屏2022-11-27 下午8 57 35
截屏2022-11-27 下午8 58 06

result is 
截屏2022-11-27 下午8 58 19

it should be 
截屏2022-11-27 下午9 04 06

### DIFF
--- a/web/src/helpers/filter.ts
+++ b/web/src/helpers/filter.ts
@@ -156,7 +156,7 @@ export const checkShouldShowMemo = (memo: Memo, filter: Filter) => {
   if (type === "TAG") {
     let contained = true;
     const tagsSet = new Set<string>();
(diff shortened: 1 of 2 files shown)


Zeng1998:
why you create a shortcut name `Test2` and filter the memos by  shortcut `Test1`?

Zeng1998:
and why is it related to the parser?

Wujiao233:
sry for the wrong image... the shortcut name should be `Test2`

Instead of edit parser, I change to edit `checkShouldShowMemo`

boojack (approved):
LGTM

PR #611, opened 2022-11-27
fix: clear shortcut filter when delete this shortcut

Wujiao233:
when delete a shortcut which is apply for the filter, the filter remain but filter item disappear, and the list still remain filtered

截屏2022-11-27 下午11 44 34


boojack (approved):
LGTM

PR #619, opened 2022-11-28
fix: selector dropdown position in fullscreen mod

dmitriisurmin:
Before
Screenshot 2022-11-28 at 07 33 21

After
Screenshot 2022-11-28 at 07 46 52


boojack (approved):
LGTM

PR #649, opened 2022-12-02
fix: resource-container overflow

hoi-lau:
## Version

0.8.0

## Steps to reproduce

Upload a file with a long name

## Device

iphone SE

## What is expected?

I can remove this file

## What is actually happening?

file-name-overflow

## Solution

```css
.resource-container {
  max-width: 100%;
}
```


boojack (approved):

PR #667, opened 2022-12-03
chore: remove emoji picker

boojack:


ksdme:
Any context on why the emoji picker was removed?

PR #668, opened 2022-12-03
feat: add swedish/svenska translation

Mozzo1000:
This adds `swedish / svenska` translation for all current translatable strings.


boojack (approved):
LGTM. Welcome aboard

PR #670, opened 2022-12-04
feat: support follow system appearance

boojack:


web/src/App.tsx
```diff
@@ -20,6 +21,15 @@ function App() {
...
   }, []);
 
+  useEffect(() => {
+    window.matchMedia("(prefers-color-scheme: dark)").addEventListener("change", (e) => {
+      if (globalService.getState().appearance === "system") {
+        const mode = e.matches ? "dark" : "light";
+        setMode(mode);
+      }
+    });
+  }, []);
```
hyoban, line 31: Use the `removeEventListener` method to remove the event listener when the component unmounts.

Issue #678, opened 2022-12-05
Automatic continuation of current list in editor


mauricebauer:
### Is your feature request related to a problem?

Its frustrating for me to write long todo lists, because I always have to type `- [ ] ` or click the todo-Item Button...
Other editors (e.g. Joplin with their new editor on mobile) automatically continue a list with the same type...

### Describe the solution you'd like

If I am currently in a line that starts with `- ` or `- [ ] ` then I expect to create the prefix automatically after I switch to a newline... A simple deletion of one character or another newline could completely remove the automatic prefix (to enable fast continuation of typing after finishing the list).

### Additional context

Similar feature request for the mobile apps of Joplin: https://discourse.joplinapp.org/t/feature-request-list-continuation-on-mobile/1742

mauricebauer:
@Zeng1998 @boojack The new feature works fine on my iPad Safari with the basic list ˋ- ˋ. But the continuation of checklists has a bug… The newline is not visible until I type something new and then a basic ˋ- ˋ will be appended and after that a ˋ- [ ] ˋ… Can anyone confirm this with the new 0.8.2?

PR #689, opened 2022-12-06
feat: auto continuation list in editor

Zeng1998:
https://user-images.githubusercontent.com/31177490/205796165-8a6a615d-30d9-4a03-8f5d-ab661e057704.mp4

### DIFF
--- a/web/src/components/MemoEditor.tsx
+++ b/web/src/components/MemoEditor.tsx
@@ -91,6 +91,23 @@ const MemoEditor = () => {
   }, [editorState.editMemoId]);
 
   const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === "Enter") {
+      if (!editorRef.current) {
+        return;
+      }
+      const cursorPosition = editorRef.current.getCursorPosition();
+      const prevValue = editorRef.current.getContent().slice(0, cursorPosition);
+      const prevRows = prevValue.split("\n");
+      const curRowIndex = prevRows.length;
+      const prevRowValue = prevRows[curRowIndex - 1];
+      if (prevRowValue.startsWith("- [ ] ") || prevRowValue.startsWith("- [x] ") || prevRowValue.startsWith("- [X] ")) {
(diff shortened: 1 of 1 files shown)


mauricebauer:
Thank you for your work so far. That's exactly what i meant in  #678 🥳

mauricebauer:
Another idea for improved writing experience: after making 2 new lines (so 1 for a new list entry and another on the empty list entry) then the list could automatically stop the continuation... This would allow writing text after the list without manually removing all the `- [ ] ` characters...
In short: first hit of the return key = new list entry (continuation)
second hit of the return key (on the empty list item) = removal of the empty list entry (to allow other text outside the list)

Zeng1998:
> Another idea for improved writing experience: after making 2 new lines (so 1 for a new list entry and another on the empty list entry) then the list could automatically stop the continuation... This would allow writing text after the list without manually removing all the `- [ ] ` characters... In short: first hit of the return key = new list entry (continuation) second hit of the return key (on the empty list item) = removal of the empty list entry (to allow other text outside the list)

Yep, actually it's also my usage habit(automatically stop).  I am trying to do this now.

Zeng1998:
I've update the code and the video.

web/src/components/MemoEditor.tsx
```diff
@@ -91,6 +91,28 @@ const MemoEditor = () => {
...
+      } else {
+        if (prevRowValue.startsWith("- [ ] ") || prevRowValue.startsWith("- [x] ") || prevRowValue.startsWith("- [X] ")) {
+          event.preventDefault();
+          editorRef.current.insertText("", "\n- [ ] ");
+        } else if (prevRowValue.startsWith("- ")) {
+          event.preventDefault();
+          editorRef.current.insertText("", "\n- ");
+        }
+      }
+    }
```
boojack, line 115:
```suggestion
          event.preventDefault();
          editorRef.current.insertText("", "\n- ");
        }
      }
      return;
    }
```

Issue #545, opened 2022-11-23
Übersetzung German


kaihir:
### Is your feature request related to a problem?

I would like to see the integration of several languages in the project. For me now German. Do we already have Christmas :)

### Describe the solution you'd like

Since the Project runs in Docker, integration in the container would be best!

boojack:
@kaihir Feel free to contribute with German i18n:) 

You can follow this PR which added the French i18n item.

PR #691, opened 2022-12-07
chore: update German translation

mauricebauer:
Switched to plural to make the difference between TAG(S) and TAG(E) clearer... Refers to #545


boojack (approved):
LGTM

PR #694, opened 2022-12-07
feat: vacuum database in setting

### LINKED ISSUE #684: DB size persists after clearing unused Resources
### Describe the bug

The size of the DB continues to persist in size even after clearing unused resources. I was hoping to shrink the size of the DB by removing memos and clearing unused resources. This is particularly important because it's SQLite and the more data used over time the slower the site will become.

### Steps to reproduce

Make several posts with images then permanently clear them and the used resources. The DB stays the same size which seems a bit counter productive for SQLite. It seems it's not actually permanently removing the data from the DB.

_No response_

Zeng1998:
Closes #684

### DIFF
--- a/server/system.go
+++ b/server/system.go
@@ -162,7 +162,7 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
 			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find user").SetInternal(err)
 		}
 		if user == nil || user.Role != api.Host {
-			// throw error
+			return echo.NewHTTPError(http.StatusUnauthorized, "Unauthorized")
 		}
 		if err := s.Store.Vacuum(ctx); err != nil {
 			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to vacuum database").SetInternal(err)
--- a/server/system.go
+++ b/server/system.go
@@ -164,6 +164,9 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
 		if user == nil || user.Role != api.Host {
 			// throw error
 		}
+		if err := s.Store.Vacuum(ctx); err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to vacuum database").SetInternal(err)
+		}
 		c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSONCharsetUTF8)
 		return nil
 	})
--- a/server/system.go
+++ b/server/system.go
@@ -167,7 +167,7 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
 		if err := s.Store.Vacuum(ctx); err != nil {
 			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to vacuum database").SetInternal(err)
 		}
-		c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSONCharsetUTF8)
+		c.Response().WriteHeader(http.StatusOK)
 		return nil
 	})
 }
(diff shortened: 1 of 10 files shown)


server/system.go
```diff
@@ -148,4 +148,37 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
...
+	g.POST("/system/vacuum", func(c echo.Context) error {
+		ctx := c.Request().Context()
+		systemStatus := api.SystemStatus{
+			DBSize: 0,
+		}
+		userID, ok := c.Get(getUserIDContextKey()).(int)
+		if ok {
+			user, err := s.Store.FindUser(ctx, &api.UserFind{
+				ID: &userID,
+			})
```
boojack, line 161: Use the following error check format:

```go
if !ok {
  return echo.NewHTTPError(http.StatusUnauthorized, "Missing user in session")
}
```

server/system.go
```diff
@@ -148,4 +148,37 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
...
+			if err != nil {
+				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find user").SetInternal(err)
+			}
+			if user != nil && user.Role == api.Host {
```
boojack, line 165:
```
if user == nil || user.Role != api.Host {
  // throw error
}
```

web/src/components/Settings/SystemSection.tsx
```diff
@@ -60,6 +60,22 @@ const SystemSection = () => {
     });
   };
 
+  const handleVacuumBtnClick = async () => {
+    try {
+      api.vacuumDatabase().then(({ data }) => {
```
boojack, line 65: We should call the system status API again instead of returning filesize in VACUUM API.

server/system.go
```diff
@@ -148,4 +148,26 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
...
+			return echo.NewHTTPError(http.StatusUnauthorized, "Missing user in session")
+		}
+		user, err := s.Store.FindUser(ctx, &api.UserFind{
+			ID: &userID,
+		})
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find user").SetInternal(err)
+		}
+		if user == nil || user.Role != api.Host {
+			// throw error
```
boojack, line 165:
```suggestion
			return echo.NewHTTPError(http.StatusUnauthorized, "Unauthorized")
```

web/src/helpers/api.ts
```diff
@@ -14,6 +14,10 @@ export function upsertSystemSetting(systemSetting: SystemSetting) {
   return axios.post<ResponseObject<SystemSetting>>("/api/system/setting", systemSetting);
 }
 
+export function vacuumDatabase() {
+  return axios.post<ResponseObject<SystemStatus>>("/api/system/vacuum");
```
boojack, line 18:
```suggestion
  return axios.post("/api/system/vacuum");
```

server/system.go
```diff
@@ -148,4 +148,26 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
...
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find user").SetInternal(err)
+		}
+		if user == nil || user.Role != api.Host {
+			return echo.NewHTTPError(http.StatusUnauthorized, "Unauthorized")
+		}
+		if err := s.Store.Vacuum(ctx); err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to vacuum database").SetInternal(err)
+		}
+		c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSONCharsetUTF8)
```
boojack, line 170:
```suggestion
		c.Response().WriteHeader(http.StatusOK)
```

boojack (approved):
LGTM

Issue #685, opened 2022-12-05
UsageHeatMap doesn't account for dailight savings


apixandru:
### Describe the bug

todayTimeStamp = 1670216400000 (Mon Dec 05 2022 05:00:00 UTC / Mon Dec 05 2022 00:00:00 ET)
todayDay = 7
usedDaysAmount = 79
beginDayTimestemp = 1663390800000 (Sat Sep 17 2022 05:00:00 UTC / Sat Sep 17 2022 01:00:00 ET)

One of my notes had the timestamp 1667086444000

This means that the calculation 
utils.getDateStampByDate(m.createdTs) - beginDayTimestemp = 3625200000

3625200000/(1000 * 3600 * 24)
41.958333333333336
but there is no 41.95 index, it should be either 41 or 42

the problem appears to be that beginDayTimestamp didn't account for DST on local date, so the adjusted start of date was 1am instead of 12am

if i manually override beginDayTimestamp to 1663387200000 then the calculations are correct and don't crash

### Steps to reproduce

I'm not sure why the timestamps can be off sometimes 

here is the list of timestamps currently available in the usage heat map
```
[...]

PR #696, opened 2022-12-07
fix: update UsageHeatMap.tsx to account for daylight savings

apixandru:
one possible suggestion for https://github.com/usememos/memos/issues/685


boojack (approved):
LGTM

PR #706, opened 2022-12-08
chore: update preview action

boojack:


.github/workflows/uffizzi-preview.yml
```diff
@@ -1,13 +1,14 @@
 name: Deploy Uffizzi Preview
 
 on:
+  pull_request:
```
ShrutiC-git, line 4: for this workflow, we depend upon the previous workflow - Build PR Image. Once the image is built from source (containing changes in the code) using the uffizzi-build action, this action should be triggered. Triggering this workflow will error out because it does not have the needed artifacts the preview is built upon.

boojack: @ShrutiC-git Thanks for your information!

Issue #742, opened 2022-12-13
Missing space in after/before formatted text rendered list / checklist


fp1l0c5l:
### Describe the bug

![firefox_0Zjbzlj2xB](https://user-images.githubusercontent.com/19887247/207226268-d89dcc20-4505-4f54-b34b-daa85c90f805.png)
![firefox_1gA3fPDihQ](https://user-images.githubusercontent.com/19887247/207226271-e90260c0-f52b-438d-bcbc-faa461a63c3e.png)

### Steps to reproduce

Create some formatted text (hyperlink, italic, bold) within list / checklist.

*Edited typo on screenshots

mauricebauer:
I can confirm this issue in version 0.8.2:
![image](https://user-images.githubusercontent.com/34477772/207626528-3eba5891-f64c-477b-ad79-3c70693ec09a.png)

PR #747, opened 2022-12-14
fix: add a span as wrapper to fix whitespace

m-gschwandtner:
Hi, small fix for #742 

is just a span wrapper around the "li-container" content.


m-gschwandtner:
ops missed the tests ...
1sec

boojack (approved):
LGTM

PR #749, opened 2022-12-16
restrict the html file

lujiefsi:


Zeng1998:
hello, why the html file is not allowed?

lujiefsi:
> hello, why the html file is not allowed?

HI:

we send the security issuse to stevenlgtm@gmail.com

server/resource.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+        "strings"
```
jxpsert, line 10: This line is prefixed by a few spaces, instead of a tab. Replace it and the check should succeed!

lujiefsi:
Is there any way to see what errors happen?

boojack (approved):
LGTM for now. But I think we could return raw text for html resource instead of disabling it.

lujiefsi:
could you verify the report on https://huntr.dev/  ?

jxpsert:
Where? Do you have a direct link?

PR #763, opened 2022-12-17
feat: editor enhancement

Zeng1998:
https://user-images.githubusercontent.com/31177490/208245769-aa0522ae-f249-4f02-a39a-188c65334c0e.mp4

1. support ordered list auto-continuation. (like in the github)
2. replace the `useRefresh` hook to a function `updateEditorHeight`  (more clearer and some weird bugs are avoided.)


boojack (approved):
LGTM

web/src/components/MemoEditor.tsx
```diff
@@ -124,17 +125,52 @@ const MemoEditor = () => {
...
+                    updated = true;
+                  } else {
+                    break;
+                  }
+                }
+                if (updated) {
+                  editorRef.current.setContent(content.join("\n"));
+                  editorRef.current.setCursorPosition(cursorPosition + 4);
+                }
+              }
```
boojack, line 171: I just remove these on my local machine, it still works right. Is there any reason about update cursor position right here?

Zeng1998: Please see the 15s of the video.

This code is to correct the subsequent number when we enter in the middle of an ordered list.

And setting the cursor position with a hard-coded number is because after I call `setContent` the cursor will be at the end.

Zeng1998: Maybe the implementation is a bit cumbersome, but I found that there are similar features in github.

PR #783, opened 2022-12-19
implement plurals for stats

jxpsert:
Implemented plurals for stats, including translations for all included languages.
If amount of memos = 0, 'memos', as in "there are 0 memos"
If amount of memos = 1, 'memo', as in "there is 1 memo"


boojack:
How about following this guide: https://www.i18next.com/translation-function/plurals

jxpsert:
> How about following this guide: https://www.i18next.com/translation-function/plurals

Will do!

jxpsert:
Changed it! @boojack

boojack (approved):
LGTM!

PR #789, opened 2022-12-20
feat: Automatic matching of input punctuation

ChasLui:
#788 For example, type
`hello "|world` will be `hello "|"world`


jxpsert:
It's a bit unclear to me what this does. Does this add a second " after you open a string with ", like VSCode (and other editors) do?

ChasLui:
> It's a bit unclear to me what this does. Does this add a second " after you open a string with ", like VSCode (and other editors) do?

yes, for example, if I type ', the content will change to '', and the cursor will appear between the two ' signs

Zeng1998:
I made another implementation about this. I think it's more clear and simple.  What do you think?

#791

ChasLui:
> I made another implementation about this. I think it's more clear and simple. What do you think?
> 
> #791
After looking at it, it can achieve the same function, but I think the punctuation can be more rich.
Then this pr can be closed.

PR #791, opened 2022-12-20
feat: matching punctuation

Zeng1998:


web/src/components/MemoEditor.tsx
```diff
@@ -15,6 +15,7 @@ import "../less/memo-editor.less";
 
 const listItemSymbolList = ["- [ ] ", "- [x] ", "- [X] ", "* ", "- "];
 const emptyOlReg = /^(\d+)\. $/;
+const pairSymbols = ["[]", "()", '""', "''"];
```
jxpsert, line 18:
```suggestion
const pairSymbols = ["[]", "()", '""', "''", "``", "{}"];
```
Bit more function.

jxpsert:
Suggestion: if the "", () or whatever is empty, and you press backspace on the first (, delete both?
Don't know how doable that is.

Zeng1998:
Yes, I know what you mean.

I am looking for a solution that is not too complicated. If you have some good suggestions, please let me know.

jxpsert:
Perhaps;
- Find index of last deleted character
- If character is in the array:
- Find index of the first corresponding character (deleted character was ``(``: look for ``)``) _after_ the deleted character
- Delete character at found index

web/src/components/MemoEditor.tsx
```diff
@@ -15,7 +15,7 @@ import "../less/memo-editor.less";
 
 const listItemSymbolList = ["- [ ] ", "- [x] ", "- [X] ", "* ", "- "];
 const emptyOlReg = /^(\d+)\. $/;
-const pairSymbols = ["[]", "()", '""', "''"];
+const pairSymbols = ["[]", "()", '""', "''", "{}", "<>", "``", "”“", "‘‘", "【】", "（）", "《》"];
```
jxpsert, line 18: I don't think you'll want to have ``<>`` here. Purely because of maths.
Try typing the equation ``x < 1``, for example. That'll make it go ``x <>``
This would be an even bigger issue with the suggested deletion feature. I recomment dropping it.

ChasLui: > I don't think you'll want to have `<>` here. Purely because of maths. Try typing the equation `x < 1`, for example. That'll make it go `x <>` This would be an even bigger issue with the suggested deletion feature. I recommend dropping it.

I think we can learn from vscode's

in `microsoft/vscode/extensions/xml/xml.language-configuration.json`
```xml
{
	"comments": {
		"blockComment": [ "<!--", "-->" ]
	},
	"brackets": [
		["<!--", "-->"],
		["<", ">"],
		["{", "}"],
		["(", ")"]
	],
	"autoClosingPairs": [
		{ "open": "{", "close": "}"},
		{ "open": "[", "close": "]"},
		{ "open": "(", "close": ")" },
		{ "open": "\"", "close": "\"", "notIn": ["string"] },
		{ "open": "'", "close": "'", "notIn": ["string"] },
		{ "open": "<!--", "close": "-->", "notIn": [ "comment", "string" ]},
		{ "open": "<![CDATA[", "close": "]]>", "notIn": [ "comment", "string" ]}
	],
	"surroundingPairs": [
		{ "open": "'", "close": "'" },
		{ "open": "\"", "close": "\"" },
		{ "open": "{", "close": "}"},
		{ "open": "[", "close": "]"},
		{ "open": "(", "close": ")" },
		{ "open": "<", "close": ">" }
	],
	"colorizedBracketPairs": [
	],
	"folding": {
		"markers": {
			"start": "^\\s*<!--\\s*#region\\b.*-->",
			"end": "^\\s*<!--\\s*#endregion\\b.*-->"
		}
	},
	"wordPattern": {
		"pattern": "(-?\\d*\\.\\d\\w*)|([^\\`\\~\\!\\@\\#\\$\\%\\^\\&\\*\\(\\)\\=\\+\\[\\{\\]\\}\\\\\\|\\;\\:\\'\\\"\\,\\.\\<\\>\\/\\?\\s]+)"
	}
}

```

jxpsert: Uhh.. elaborate?

ChasLui: Use the definition of surfingPairs for these five Pairs

jxpsert:
Looks good to me now. Now it's waiting for codeowners.

jxpsert (approved):

web/src/components/MemoEditor.tsx
```diff
@@ -169,6 +170,29 @@ const MemoEditor = () => {
...
+        event.preventDefault();
+        editorRef.current.insertText("", symbol[0], symbol[1]);
+        return;
+      }
+    }
+
+    if (event.key === "Backspace") {
+      if (!editorRef.current) {
+        return;
+      }
```
boojack, line 185:
```suggestion
```

jxpsert: What does this do/change?

jxpsert: What does this do/change?

boojack: There already has a checker in the beginning. https://github.com/usememos/memos/pull/791/files#diff-e9a053aeb57009e1c1d9fe8f830f4fdd23956ec4fc62446159d07fdd06649272R101

boojack (approved):
LGTM

PR #797, opened 2022-12-20
fix: incorrect cursor when text is selected in range

ChasLui:
Before: When text is selected in the range, the cursor is positioned at the current position + the length of the selected text
After: The cursor will be positioned at the start of the selected text content + TAB_SPACE_WIDTH


boojack (approved):
LGTM

PR #806, opened 2022-12-21
feat: show active panel when searchBar is in focus

ChasLui:
To solve the problem of displaying the active panel when the searchBar is focused after pressing `cmd+f`


boojack (approved):
LGTM

PR #813, opened 2022-12-21
feat: support tag auto-completion

Zeng1998:
Please have a try and make some suggestions.

A relative simple implementation, maybe not working in some corner cases.

### DIFF
--- a/web/src/components/Editor/Editor.tsx
+++ b/web/src/components/Editor/Editor.tsx
@@ -8,8 +8,8 @@ export interface EditorRefActions {
   setContent: (text: string) => void;
   getContent: () => string;
   getSelectedContent: () => string;
-  getCursorPosition: () => number;
-  setCursorPosition: (startPos: number, endPos?: number) => void;
+  getCaretPosition: () => number;
+  setCaretPosition: (startPos: number, endPos?: number) => void;
 }
 
 interface Props {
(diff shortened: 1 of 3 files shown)


ChasLui:
image
Would it be better to remove transparency from the Tag panel background color? Now I can see the `Only visible to you' in the lower level and it feels strange.

ChasLui:
Press esc to close the tag panel

ChasLui:
To prevent false triggers, should I use space+# as a match? For example, in github issue and flomo.

Zeng1998:
> To prevent false triggers, should I use space+# as a match? For example, in github issue and flomo.

Hello, what do you mean about "space+#" ?

In github, when I type a "#"，the issue/pr selection panel are shown.

ChasLui:
> > To prevent false triggers, should I use space+# as a match? For example, in github issue and flomo.
> 
> Hello, what do you mean about "space+#" ?
> 
> In github, when I type a "#"，the issue/pr selection panel are shown.

For example, I have a text `hello world`, when I type a `hello world#` at the end it doesn't show the tag panel, if I type `hello world #` then the tag panel appears.

jxpsert:
> To prevent false triggers, should I use space+# as a match? For example, in github issue and flomo.

I think so, yes. You might want to use URLs that have # in them (think ``https://www.your.site/#about-me``).

Zeng1998:
But when I just want to type #tag in one line.

Do you mean I should type one space before?

ChasLui:
> 

The new line may require some special handling, in github the new line header can be touched by typing # in the tab panel

Zeng1998:
Sorry I can't get the point. As the video show, there seems like no other trigger in github editor? (Just type one "#" is ok.)


https://user-images.githubusercontent.com/31177490/209596226-9ddbce18-5ecd-490e-840a-1c44b7ee163a.mp4

ChasLui:
hello world www.baidu.com 

> Sorry I can't get the point. As the video show, there seems like no other trigger in github editor? (Just type one "#" is ok.)
> 
>  20221227_091612.mp4


https://user-images.githubusercontent.com/10083758/209618532-638148e9-07e3-4213-86dd-758cb6ed0798.mov

PR #847, opened 2022-12-24
fix: prioritize user css by moving it to the body end

m-gschwandtner:
#812

think this is the easiest option


boojack:
it's a little weird to put <style/> in <body/>. But I'm ok.
https://stackoverflow.com/questions/2830296/using-style-tags-in-the-body-with-other-html

boojack (approved):

m-gschwandtner:
I know style in body is not the best, but since the site-specific styles like "/assets/Home-3e078dbe.css" get added on load of a component, appending it to the head wouldn't work as far as I know, not sure tough.

Thanks

PR #877, opened 2022-12-30
fix: logic of email validation

Zeng1998:


boojack:
Another place: https://github.com/usememos/memos/blob/main/api/user.go#L79

boojack (approved):

PR #964, opened 2023-01-17
feat: add qrcode in share dialog

Zeng1998:


yefoenix:
感谢 PR!

提个小建议，二维码的颜色可以改成
 
bgColor={"#F3F4F6"} 
 fgColor={"#4B5563"}

分别是 watermark-container 的背景色 和 .name-text 的字体颜色，这样整体颜色更和谐

Zeng1998:
> And I think we should show QRCode only for those public memos.

I have a different opinion. The QR code is actually the link to the memo, which does exist objectively. It is equivalent to the Link button, no matter which visibility.

boojack (approved):
LGTM

PR #970, opened 2023-01-18
feat: support swipe to switch img on touchscreen

Wujiao233:
it's hard to touch the container div when image ratio be similar to device screen ratio


boojack (approved):
LGTM

PR #979, opened 2023-01-19
fix: round corner issue in resource blocks

Wujiao233:
when upload image scale not 1:1, it look like :
4633b4ff-672a-4589-b5fd-d3a349f3f303
And it can scroll inside container.
After fix:
ec14f909-59f7-44cb-8c34-e388a3ce7315


boojack (approved):
LGTM

PR #980, opened 2023-01-20
fix: video and audio can't play on safari

Wujiao233:
Problem is video (and audio) can't play on safari and on iOS webapp mode .
image
image

video resource also can't play when direct request the url like 
```
https://memos_host/o/r/<id>/filename.mp4
```
a error in console is `Failed to Load Resource, Plugin Handled Load`.

And after some searching ,it looks like current `c.Stream()` function cannot response request with Header `Range: bytes=0-1` correctly. According to [rfc7233](https://www.rfc-editor.org/rfc/rfc7233#section-4.1) it should respone with code 206.And `c.Stream()` function also can't return part of content, it can be test use the following code:
```
curl --range 0-99  https://memos_host/o/r/<id>/filename.mp4 -o output.mp4
# if output.mp4 file size is precisely 100bytes then the api can return part of content
```

So for the video&audio content I use `http.ServeContent` Instead of `Stream()` to return the data because it support Range Header and work well on safari and iOS ,And I haven't found similar function in Echo4.
After saving resource file out of dbfile ,it can be change to use `c.File()`.

After Fix:
image

sry for my bad description, and happy new year.

### DIFF
--- a/server/resource.go
+++ b/server/resource.go
@@ -274,6 +274,10 @@ func (s *Server) registerResourcePublicRoutes(g *echo.Group) {
 		}
 		c.Response().Writer.Header().Set(echo.HeaderCacheControl, "max-age=31536000, immutable")
 		c.Response().Writer.Header().Set(echo.HeaderContentSecurityPolicy, "default-src 'self'")
+		if strings.HasPrefix(resourceType, "video") {
+			http.ServeContent(c.Response(), c.Request(), resource.Filename, time.Unix(resource.UpdatedTs, 0), bytes.NewReader(resource.Blob))
+			return nil
+		}
 		return c.Stream(http.StatusOK, resourceType, bytes.NewReader(resource.Blob))
 	})
 }
--- a/server/resource.go
+++ b/server/resource.go
@@ -274,7 +274,7 @@ func (s *Server) registerResourcePublicRoutes(g *echo.Group) {
 		}
 		c.Response().Writer.Header().Set(echo.HeaderCacheControl, "max-age=31536000, immutable")
 		c.Response().Writer.Header().Set(echo.HeaderContentSecurityPolicy, "default-src 'self'")
-		if strings.HasPrefix(resourceType, "video") {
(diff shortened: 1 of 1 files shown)


server/resource.go
```diff
@@ -274,6 +274,10 @@ func (s *Server) registerResourcePublicRoutes(g *echo.Group) {
 		}
 		c.Response().Writer.Header().Set(echo.HeaderCacheControl, "max-age=31536000, immutable")
 		c.Response().Writer.Header().Set(echo.HeaderContentSecurityPolicy, "default-src 'self'")
+		if strings.HasPrefix(resourceType, "video") || strings.HasPrefix(resourceType, "audio") {
+			http.ServeContent(c.Response(), c.Request(), resource.Filename, time.Unix(resource.UpdatedTs, 0), bytes.NewReader(resource.Blob))
+			return nil
+		}
 		return c.Stream(http.StatusOK, resourceType, bytes.NewReader(resource.Blob))
```
boojack, line 281: Good catch! I think we can use `c.Blob` instead of `c.Stream`. Could you take a test for it in your machine again
https://echo.labstack.com/guide/response/#send-blob

Wujiao233: Actually I tested it before and `c.Blob` doesn't work .. 
image
And it also got `Failed to Load Resource, Plugin Handled Load` on safari 

Test code:
```golang
f, _ := os.Open("./video.mp4")
fd, _ := io.ReadAll(f)

// Routes
e.GET("/test-httpserve.mp4", func(c echo.Context) error {
	ff := bytes.NewReader(fd)
	http.ServeContent(c.Response(), c.Request(), "video.mp4", time.Now(), ff)
	return nil
})
e.GET("/test-blob.mp4", func(c echo.Context) error {
	return c.Blob(http.StatusOK, "video/mp4", fd) // also try http.StatusPartialContent and not work..
})
```

boojack (approved):
LGTM, thanks for your contribution

Issue #919, opened 2023-01-07
Ability to fully disable public memos


sfan5:
### Is your feature request related to a problem?

I use Memos exclusively as a private note taking application for myself.
However the option to make memos public still exists and it's possible that I (or someone else, if the instance is shared) someday mistakenly make a memo public.

### Describe the solution you'd like

For that reason it would be useful to have global admin configuration that can *disable* public memos entirely.

itschip:
I'll be making a PR for this shortly :)

skyue:
效率也太快了，还想说一步到位：来个 XXX 内天可见，然后支持0天可见。

sparrowyang:
我不希望未注册用户看到网站的内容，所以我开启了**禁用公共memos**现在有些问题：
1. 我仍希望已注册用户能发布protected(登录用户可见)的memos，但发布时无法选择。
2. 即使开启这个开关，在点击分享按钮时，仍然能改变memos的可见性为public(所有人可见)。我不希望用户能让memos公布。

sparrowyang:
> > Issue is not in English. It has been translated automatically.
> 
> I don't want unregistered users to see the content of the site, so I turned on **disable public memos** and now I have some problems:
> 
> 1. I still want registered users to be able to publish protected (visible to logged in users) memos, but they cannot be selected when publishing.
> 2. Even if this switch is turned on, when the share button is clicked, the visibility of the memos can still be changed to public (visible to all). I don't want the user to make the memos public.

@itschip

PR #1003, opened 2023-01-31
feat(system): support for disabling public memos

itschip:
This PR opens the possibility to disable public memos as requested in https://github.com/usememos/memos/issues/919

**Changes:**
1. Tooltip and disabled styling on the MemoEditor visibility selector.
2. System setting for disabling public memos

Screenshot 2023-01-31 at 18 00 10

Screenshot 2023-01-31 at 18 00 28

### DIFF
--- a/web/src/components/Settings/SystemSection.tsx
+++ b/web/src/components/Settings/SystemSection.tsx
@@ -6,10 +6,13 @@ import * as api from "../../helpers/api";
 import toastHelper from "../Toast";
 import showUpdateCustomizedProfileDialog from "../UpdateCustomizedProfileDialog";
 import "@/less/settings/system-section.less";
+import { useAppDispatch } from "../../store";
+import { setGlobalState } from "../../store/reducer/global";
 
 interface State {
   dbSize: number;
   allowSignUp: boolean;
+  disablePublicMemos: boolean;
   additionalStyle: string;
   additionalScript: string;
 }
@@ -32,7 +35,10 @@ const SystemSection = () => {
     allowSignUp: systemStatus.allowSignUp,
     additionalStyle: systemStatus.additionalStyle,
     additionalScript: systemStatus.additionalScript,
+    disablePublicMemos: systemStatus.disablePublicMemos,
   });
+
+  const dispatch = useAppDispatch();
 
   useEffect(() => {
     globalStore.fetchSystemStatus();
@@ -44,6 +50,7 @@ const SystemSection = () => {
       allowSignUp: systemStatus.allowSignUp,
       additionalStyle: systemStatus.additionalStyle,
       additionalScript: systemStatus.additionalScript,
+      disablePublicMemos: systemStatus.disablePublicMemos,
     });
   }, [systemStatus]);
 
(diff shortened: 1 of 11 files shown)


hyoban:
After turning on this option, if `Everyone can see` was set before, the displayed information is a bit confusing

image

hyoban:
And, should we take this setting into account at the backend api. When it is enabled, prohibits the creation of public memos

itschip:
> And, should we take this setting into account at the backend api. When it is enabled, prohibits the creation of public memos

You're right 👍🏻 I'll fix that

> After turning on this option, if Everyone can see was set before, the displayed information is a bit confusing

Yeah, looks like I went a bit too fast and forgot to set the value back to "only visible to you" :)

server/memo.go
```diff
@@ -30,6 +30,16 @@ func (s *Server) registerMemoRoutes(g *echo.Group) {
 			return echo.NewHTTPError(http.StatusBadRequest, "Malformatted post memo request").SetInternal(err)
 		}
 
+		// Find system settings
+		systemSettingKey := api.SystemSettingDisablePublicMemosName
+		systemSetting, err := s.Store.FindSystemSetting(ctx, &api.SystemSettingFind{
```
jxpsert, line 35: This can probably be solved with another ``if err != nil`` check?

itschip: You're right. Totally forgot to do an early return here :D

itschip:
@hyoban do you want me to rebase to one commit?

hyoban:
It's fine to have multiple commits, we'll handle that before merging

server/memo.go
```diff
@@ -30,6 +30,19 @@ func (s *Server) registerMemoRoutes(g *echo.Group) {
...
+			Name: &systemSettingKey,
+		})
+		if err != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find system setting").SetInternal(err)
+		}
+
+		if systemSetting.Value == "true" {
```
boojack, line 42: Unmarshal it before using it.

e.g. https://github.com/usememos/memos/blob/main/server/memo.go#L45

web/src/components/common/Selector.tsx
```diff
@@ -3,6 +3,7 @@ import { useTranslation } from "react-i18next";
 import useToggle from "../../hooks/useToggle";
 import Icon from "../Icon";
 import "../../less/common/selector.less";
```
hyoban, line 5: Side effects imports need to be at the end

web/src/components/Settings/SystemSection.tsx
```diff
@@ -6,10 +6,13 @@ import * as api from "../../helpers/api";
 import toastHelper from "../Toast";
 import showUpdateCustomizedProfileDialog from "../UpdateCustomizedProfileDialog";
 import "@/less/settings/system-section.less";
```
hyoban, line 8: Side effects imports need to be at the end

web/src/components/common/Selector.tsx
hyoban, line 5: Side effects imports need to be at the end

server/memo.go
```diff
@@ -53,6 +53,25 @@ func (s *Server) registerMemoRoutes(g *echo.Group) {
 			}
 		}
 
+		// Find system settings
```
boojack, line 56: @itschip Please check out what I have changed in the commit.

https://github.com/usememos/memos/pull/1003/commits/0de531ecf2481f25ec8bca46f1c4def8eb31ac97

boojack (approved):

server/memo.go
itschip: I was a bit confused what to do here! Thanks :)

PR #1014, opened 2023-02-03
feat: support enter to signin

### LINKED ISSUE #1012: UX issue: Sign in needs mouse click
### Describe the bug

This is a usability issue so I personally consider such an issue as a bug though you might classify this as an enhancement request.

### Steps to reproduce

On Login page, fill in username and password, hit enter.

Expected behavior: log in is performed
Actual behavior: nothing happens

Please note that this is typical behavior of every (!) web application around and other applications (for example password managers with auto type sequence `{user}<TAB>{password}<ENTER>`) are designed around this expectation so you shouldn't break it.

Zeng1998:
Closes #1012

### DIFF
--- a/web/src/pages/Auth.tsx
+++ b/web/src/pages/Auth.tsx
@@ -117,6 +117,12 @@ const Auth = () => {
     actionBtnLoadingState.setFinish();
   };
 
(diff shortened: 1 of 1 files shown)


boojack (approved):
LGTM

PR #1020, opened 2023-02-04
fix: tag detection start with /

Zeng1998:


boojack:
Fixed in https://github.com/usememos/memos/pull/1031

PR #1023, opened 2023-02-04
chore: update zh-Hant.json

den13501:
Correct some translation of zh-tw, It is more in line with the language habits of traditional Chinese users


boojack (approved):
LGTM

PR #1028, opened 2023-02-06
more translation correction to traditional Chinese

den13501:
To be in line with the language habits of traditional Chinese users


boojack (approved):
LGTM

PR #1032, opened 2023-02-06
feat: tag filter in explore

### LINKED ISSUE #536: Tags can not be browsed by public viewers

When logged out, tags cannot be clicked and applied as a filter for public viewers. Currently when logged out and you click on a tag in a public post, nothing happens. It appears tags can only be browsed when logged in.

When logged out on the explore page, tags should be able to be clicked to see any public posts with the same tag as a filter. This is good for those using this as a status (blog) for personal use. Tags can then be used to help categorize things for viewers.

### Additional context

_No response_

Zeng1998:
Closes #536

### DIFF
--- a/web/src/pages/Explore.tsx
+++ b/web/src/pages/Explore.tsx
@@ -67,10 +67,6 @@ const Explore = () => {
             shouldShow = false;
           }
         }
-
-        if (textQuery && !memo.content.toLowerCase().includes(textQuery.toLowerCase())) {
-          shouldShow = false;
-        }
         return shouldShow;
       })
     : state.memos;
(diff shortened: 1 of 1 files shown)


web/src/pages/Explore.tsx
```diff
@@ -41,6 +44,42 @@ const Explore = () => {
...
+      })
+    : state.memos;
+
+  const pinnedMemos = shownMemos.filter((m) => m.pinned);
+  const unpinnedMemos = shownMemos.filter((m) => !m.pinned);
+  const memoSort = (mi: Memo, mj: Memo) => {
+    return mj.displayTs - mi.displayTs;
+  };
+  pinnedMemos.sort(memoSort);
+  unpinnedMemos.sort(memoSort);
```
boojack, line 80: Pin/unpin memo is a personal behaviour and should not be display in explore page.

boojack (approved):
LGTM

PR #1036, opened 2023-02-07
feat: add setting for double-click of memos

Zeng1998:


Zeng1998:
#922 1

web/src/types/modules/setting.d.ts
```diff
@@ -10,6 +10,7 @@ interface Setting {
 interface LocalSetting {
   enableFoldMemo: boolean;
   enablePowerfulEditor: boolean;
+  enableDoubleClick: boolean;
```
boojack, line 13: How about `enableDoubleClickEditing`?

PR #1052, opened 2023-02-09
feat: more rss info

hyoban:
1. use `# title` as rss item title
2. use system custom profile as rss title and description
3. rss link in explore page and user page


boojack (approved):
LGTM

PR #1131, opened 2023-02-20
feat: add feat args

Yoshino-s:


server/profile/profile.go
```diff
@@ -22,6 +24,22 @@ type Profile struct {
 	DSN string `json:"-"`
 	// Version is the current version of server
 	Version string `json:"version"`
+	// Feat is the feature set of server, split by comma
+	Feat string `json:"feature"`
+}
+
+const (
+	FeatSSO       Feature = "SSO"
```
boojack, line 32: * Using full name will be more easy to understand;
* Using `FeatureType` instead of `Feature`

And let's move them into a new file `server/profile/feature.go`

```suggestion
	FeatureSSO       FeatureType = "SSO"
```

boojack (changes requested):

server/profile/profile.go
```diff
@@ -22,6 +24,22 @@ type Profile struct {
...
 	// Version is the current version of server
 	Version string `json:"version"`
+	// Feat is the feature set of server, split by comma
+	Feat string `json:"feature"`
+}
+
+const (
+	FeatureSSO       Feature = "SSO"
+	FeatureStorageS3 Feature = "STORAGE_S3"
+)
```
Zeng1998, line 34: I think you are not using these two enum values.

Yoshino-s: these two is using in frontend, holding here for future using in backend

server/profile/profile.go
Zeng1998: image

Yoshino-s: i'd prefer `server/profile/feature.go` rather than `server/feature.go`

server/profile/feature.go
```diff
@@ -0,0 +1,19 @@
...
+import "strings"
+
+type FeatureType string
+
+const (
+	FeatureSSO       FeatureType = "SSO"
+	FeatureStorageS3 FeatureType = "STORAGE_S3"
+)
+
+func (p *Profile) IsFeatEnabled(feat FeatureType) bool {
```
boojack, line 12:
```suggestion
func (p *Profile) IsFeatureEnabled(feat FeatureType) bool {
```

cmd/server.go
```diff
@@ -0,0 +1,125 @@
...
+	return rootCmd.Execute()
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.PersistentFlags().StringVarP(&mode, "mode", "m", "demo", `mode of server, can be "prod" or "dev" or "demo"`)
+	rootCmd.PersistentFlags().IntVarP(&port, "port", "p", 8081, "port of server")
+	rootCmd.PersistentFlags().StringVarP(&data, "data", "d", "", "data directory")
+	rootCmd.PersistentFlags().StringVarP(&feat, "feat", "f", "", "feature flags, split by comma")
```
boojack, line 85: Why do we need a flag to enable features? 

IMO, we should use it together with the `mode`. e.g. default enable these features in dev/demo mode but disable in prod mode.

web/src/store/module/global.ts
```diff
@@ -67,6 +67,9 @@ export const useGlobalStore = () => {
     isDev: () => {
       return state.systemStatus.profile.mode !== "prod";
     },
+    isFeatEnabled: (feat: Feature) => {
```
boojack, line 70:
```suggestion
    isFeatureEnabled: (feature: Feature) => {
```

cmd/server.go
Yoshino-s: in the future, we can use feature args to enable/disable some advanced actions, which may hava break change, like store resources in local file system instead of db.

Yoshino-s: for those people who want to still use resource in db, they can just run as before, but for others, they can use this feature manually

boojack: > store resources in local file system instead of db.

As we already have system settings, maybe it will has a better UX to set the resource storage method in UI + system settings? 

telegram-cloud-photo-size-5-6307292232071362379-y

Zeng1998: So actually... I am confused what exactly this PR brings.

Yoshino-s: > > store resources in local file system instead of db.
> 
> As we already have system settings, maybe it will has a better UX to set the resource storage method in UI + system settings? 
> 
> telegram-cloud-photo-size-5-6307292232071362379-y
> 

looks good

boojack: PTAL https://github.com/usememos/memos/pull/1140

PR #1140, opened 2023-02-22
feat: update storage setting section

boojack:
image


Yoshino-s:
In fact, there are very few scenarios where multiple s3 storages are required to switch to each other in one server.
I think a better ui maybe described as
1. One single choice from `database,s3,local`
2. if choose database, no options need
3. if choose s3, a form shown below and need to fill
4. if choose local, another form need to fill

boojack:
> In fact, there are very few scenarios where multiple s3 storages are required to switch to each other in one server. I think a better ui maybe described as
> 
> 1. One single choice from `database,s3,local`
> 2. if choose database, no options need
> 3. if choose s3, a form shown below and need to fill
> 4. if choose local, another form need to fill

Can we think of local file storage as a type of `storage` too? And save its config information into `storage` table.

Yoshino-s:
maybe a better struct should be like
```sql
CREATE TABLE configuration (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  name TEXT NOT NULL DEFAULT '' UNIQUE,
  value TEXT NOT NULL DEFAULT ''
);
```
then we can do
```sql
INSERT INTO
  configuration (`name`, `value`)
VALUES
  ("local.path", "/var/opt/resources");

SELECT value FROM configuration WHERE name = "s3.access_key";
```
Also, if we need more configuration in the future, we can also integrate them into this table

boojack:
> maybe we should integrate them all into system_setting tables?

`system_setting` saves those simple fields like locale, theme and server profiles. It might not be suitable for structure fields.

Oaklight:
In the case of switching from database to local or s3, should there be an option ( a button for instance) to transfer all previously base64-stored "resources" to the new location?

PR #1146, opened 2023-02-24
[WIP]chore: security improve

Yoshino-s:
doing some staff

1. prevent SSRF in http_meta plugin
2. improve acl design, some api may have break change in auth, check if we need documented them
3. add new session named sessionID, which made a session can be force invalidated


Yoshino-s:
BTW, echo's session is like a sh*t, depending on an archived project, no force expire...

PR #1156, opened 2023-02-25
chore: remove validators on the frontend

Zeng1998:


boojack:
Let's remove all of these validators in frontend.

api/user.go
```diff
@@ -67,6 +67,12 @@ func (create UserCreate) Validate() error {
 	if len(create.Username) > 32 {
 		return fmt.Errorf("username is too long, maximum length is 32")
 	}
+	if len(create.Password) < 6 {
```
boojack, line 70: Let's set the minimum to `3`

api/user.go
```diff
@@ -67,6 +67,12 @@ func (create UserCreate) Validate() error {
...
+		return fmt.Errorf("password is too short, minimum length is 6")
+	}
+	if len(create.Password) > 64 {
```
boojack, line 73: BTW, the password is generated random hash, so we don't have to check its maximum.

Zeng1998: I don't know much about security, but it seems like it could be a point of attack if we don't restrict the max length. E.g., `long password denial of service`.
I think we can set a larger value, such as 512, to ensure that is enough for almost all users. WDYT?

boojack: `512` LGTM

boojack (approved):
LGTM

PR #1179, opened 2023-02-27
fix: model fields are unconsistent with the data queried from the database

tonghs:


Zeng1998:
Nice catch!

boojack (approved):

PR #1202, opened 2023-02-28
I added Turkish Translation file

mehmetaltugakgul:


TheRealJd3:
We should have that added here as well: 
https://github.com/usememos/memos/blob/1f05b52c4e88b9d870e1fe30701d4ca9da76deb2/api/user_setting.go#L39

web/src/components/LocaleSelect.tsx
```diff
@@ -29,12 +29,13 @@ const LocaleSelect: FC<Props> = (props: Props) => {
       <Option value="nl">Nederlands</Option>
       <Option value="sv">Svenska</Option>
       <Option value="de">German</Option>
-      <Option value="es">Español</Option>
+      <Option value="es">Español</Option>
```
boojack, line 32:
```suggestion
      <Option value="es">Español</Option>
```

web/package.json
```diff
@@ -3,7 +3,7 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc && vite build",
-    "lint": "eslint --ext .js,.ts,.tsx, src"
+    "lint": "eslint --ext .js,.ts,.tsx, src --fix"
```
boojack, line 6: We use this script to check if eslint has some warnings or errors instead of fixing them. How about add a new scripts like `lint-fix`?

boojack (changes requested):
And you have to change the server codes too. Please take a look into the changed files in another PR https://github.com/usememos/memos/pull/879/files

web/package.json
boojack: I think we should not change it in this PR. This PR is mainly introducing a new language. WDYT

mehmetaltugakgul:
i think I've also made other updates, there's nothing missing

web/package.json
mehmetaltugakgul: Right, I think the same way, that's why I've reinstated it again.

api/user_setting.go
```diff
@@ -36,7 +36,7 @@ func (key UserSettingKey) String() string {
 }
 
 var (
-	UserSettingLocaleValue             = []string{"en", "zh", "vi", "fr", "nl", "sv", "de", "es", "uk", "ru", "it", "hant", "ko"}
+	UserSettingLocaleValue             = []string{"en", "zh", "vi", "fr", "nl", "sv", "de", "es", "uk", "ru", "it", "hant", "tr", "ko"}
```
TheRealJd3, line 39: Can we have this order match the order in `web/src/components/LocaleSelect.tsx`

TheRealJd3: In the other files `tr` is after `ko` lets keep it the same way here as well

mehmetaltugakgul: Updated. thx.

mehmetaltugakgul:
I have made the necessary changes that you suggested.

web/src/i18n.ts
```diff
@@ -55,6 +56,9 @@ i18n.use(initReactI18next).init({
     ko: {
       translation: koLocale,
     },
+    tr: {
```
TheRealJd3, line 59: order on this file as well, thanks!

mehmetaltugakgul: Updated. Thx.

mehmetaltugakgul:
I have made the necessary changes that you suggested finally.

boojack (approved):

PR #1205, opened 2023-02-28
fix: Incorrect sharing image gerneration (#1157)

CorrectRoadH:
The original sharing image generation method  is generated in startup. When the images is load incompletely , the shared image is incorrect.

So I change the timing of generation sharing image after user click image button.

preview:
CS_2023-02-28_at_22 32 06


boojack (approved):
LGTM!

boojack:
Thanks for your contribution

Issue #1176, opened 2023-02-27
S3 Storage supports to set UsePathStyle to true


HappyZ:
### Is your feature request related to a problem?

Not familiar with how web interface works so not sure how to add an option to the form.

But the idea is to add to https://github.com/usememos/memos/blob/main/plugin/storage/s3/s3.go#L46 like the following:

```
	client := awss3.NewFromConfig(cfg, func(options *awss3.Options) {
		options.UsePathStyle = awss3.Bool(true)
	})
```

I need this because I locally set up storage with MinIO using nginx reverse proxy, and it does not support DNS style hosts like `somebucket.mys3.com`. Instead I'd like to use `mys3.com/somebucket`.

### Describe the solution you'd like

Adds 
```
	client := awss3.NewFromConfig(cfg, func(options *awss3.Options) {
		options.UsePathStyle = awss3.Bool(true)
	})
```

Or if folks like it to be an optional thing:

```
	client := awss3.NewFromConfig(cfg, func(options *awss3.Options) {
		options.UsePathStyle = awss3.Bool(config.UsePathStyle)
	})
```

Then add UsePathStyle to the config and UI. Thanks.

Zeng1998:
The path-style URLs will be deprecessed in the future. So I don't recommend implementing it. 

As for MinIO, there should be some corresponding solutions can be find.

https://aws.amazon.com/cn/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

HappyZ:
gotcha, will see if I find anything for MinIO. Likely an nginx hack

HappyZ:
lol I didn't do anything but a hack to the input. Using above example, so I created a bucket named `somebucket`.

The config is:

```
Name: minio
EndPoint(S3-compatible server URL): https://mys3.com/somebucket
Region(Region name): xxx
AccessKey(Access Key / Access ID): xxx
SecretKey(Secret Key / Secret Access Key): xxx
Bucket(Bucket name): somebucket
URLPrefix(Custom URL prefix; Optional): <empty>
```

Is is expected when you rename/delete a resource in db, it does not delete in s3 storage? @Zeng1998

Zeng1998:
Yes, it's expected for now.  A resource in memos is just a reference to the external storage if you use S3.

IIllIIIlll:
[...]

PR #1230, opened 2023-03-02
feat: add `HostnameImmutable` to aws endpoint config

zhaochunqi:
When uploading files without the old config.  Server url like "https://minio.test.com" would be converted to "https://bucketname.minio.test.com". For the self-host minio instance, the upload would get errors. With this config enabled, everything would just work perfectly! 

some resource related: https://stackoverflow.com/questions/67575681/is-aws-go-sdk-v2-integrated-with-local-minio-server

### DIFF
--- a/plugin/storage/s3/s3.go
+++ b/plugin/storage/s3/s3.go
@@ -32,8 +32,8 @@ func NewClient(ctx context.Context, config *Config) (*Client, error) {
 		return aws.Endpoint{
 			URL:           config.EndPoint,
 			SigningRegion: config.Region,
-			//For some s3-compatible object stores, converting the hostname is not required,
-			//and not setting this option will result in not being able to access the corresponding object store address.
+			// For some s3-compatible object stores, converting the hostname is not required,
+			// and not setting this option will result in not being able to access the corresponding object store address.
 			HostnameImmutable: true,
 		}, nil
 	})
--- a/plugin/storage/s3/s3.go
+++ b/plugin/storage/s3/s3.go
@@ -30,8 +30,10 @@ type Client struct {
 func NewClient(ctx context.Context, config *Config) (*Client, error) {
 	resolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
 		return aws.Endpoint{
-			URL:               config.EndPoint,
-			SigningRegion:     config.Region,
+			URL:           config.EndPoint,
+			SigningRegion: config.Region,
+			//For some s3-compatible object stores, converting the hostname is not required,
+			//and not setting this option will result in not being able to access the corresponding object store address.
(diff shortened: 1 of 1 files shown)


HappyZ:
I think this relates to my question about using UsePathStyle flag.  See https://stackoverflow.com/questions/65953564/issue-with-s3-bucket-naming-policy-and-attempting-to-send-sdk-generated-requests (see updated comment) for counter argument to this change.

zhaochunqi:
> I think this relates to my question about using UsePathStyle flag. See https://stackoverflow.com/questions/65953564/issue-with-s3-bucket-naming-policy-and-attempting-to-send-sdk-generated-requests (see updated comment) for counter argument to this change.

Thanks for the information. That explains why I saw some people here using minios without a problem. Seems like using camelCase won't encounter this problem due to the SDK fallback.

Zeng1998 (commented):
As I said in issue #1176 , in fact, personally, I don't recommend making some changes to the URL style of S3.

zhaochunqi:
> As I said in issue #1176 , in fact, personally, I don't recommend making some changes to the URL style of S3.

Could we make an option for this?  For self-hosted minio, it's not necessary to implement urls like `bucketname.minio.test.com`, and if we do want to use URLs like that for an s3 bucket, we can directly set url like `https://bucketname.minio.test.com` to the endpoint.

plugin/storage/s3/s3.go
```diff
@@ -30,8 +30,9 @@ type Client struct {
 func NewClient(ctx context.Context, config *Config) (*Client, error) {
 	resolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
 		return aws.Endpoint{
-			URL:           config.EndPoint,
-			SigningRegion: config.Region,
+			URL:               config.EndPoint,
```
soheil, line 33: Great job adding support for S3-compatible urls for minio!

plugin/storage/s3/s3.go
```diff
@@ -30,8 +30,9 @@
...
+			SigningRegion:     config.Region,
+			HostnameImmutable: true,
 		}, nil
 	})
```
soheil, line 38: Consider adding comments to explain your added code changes.

soheil (commented):
Hey, I've reviewed your code changes, and have some feedback for you:

The added changes looks great, it looks like you've added support for S3-compatible urls for minio!

One suggestion I have is to add comments to explain the code changes. This should help other developers understand what you've done.

Apart from that, I have no further feedback to provide. Keep up the great work!

zhaochunqi:
In fact, it might be a better choice to add this as an option to the UI. That way, existing configurations won't have problems with it.

boojack (approved):
LGTM

PR #1232, opened 2023-03-02
use $PWD/memos-data as default docker volume path

H3arn:
`~/.memos` might not be a path friendly for maintenance. It's likely that the db will be simply forgotten in the home folder. 

Using `./memos-data` could be a better idea.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ build
 .idea
 
 bin/air
+
+# docker volume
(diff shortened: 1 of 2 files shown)


zhaochunqi:
Actually, I think using volume could be even better!

H3arn:
> Actually, I think using volume could be even better!

Looks good

PR #1233, opened 2023-03-02
add support for s3 path

zhaochunqi:
This pr added support for the s3 path.

If you want to upload your images to like `image` folder, you can set path variable to `path/`. (default value will be empty)

### DIFF
--- a/plugin/storage/s3/s3.go
+++ b/plugin/storage/s3/s3.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"path"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	s3config "github.com/aws/aws-sdk-go-v2/config"
@@ -56,7 +57,7 @@ func (client *Client) UploadFile(ctx context.Context, filename string, fileType 
 	uploader := manager.NewUploader(client.Client)
 	uploadOutput, err := uploader.Upload(ctx, &awss3.PutObjectInput{
 		Bucket:      aws.String(client.Config.Bucket),
-		Key:         aws.String(client.Config.Path + filename),
+		Key:         aws.String(path.Join(client.Config.Path, filename)),
 		Body:        src,
 		ContentType: aws.String(fileType),
 		ACL:         types.ObjectCannedACL(*aws.String("public-read")),
--- a/plugin/storage/s3/s3.go
+++ b/plugin/storage/s3/s3.go
@@ -18,6 +18,7 @@ type Config struct {
 	SecretKey string
 	Bucket    string
 	EndPoint  string
+	Path      string
 	Region    string
 	URLPrefix string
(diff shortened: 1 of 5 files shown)


plugin/storage/s3/s3.go
```diff
@@ -55,7 +56,7 @@ func (client *Client) UploadFile(ctx context.Context, filename string, fileType
 	uploader := manager.NewUploader(client.Client)
 	uploadOutput, err := uploader.Upload(ctx, &awss3.PutObjectInput{
 		Bucket:      aws.String(client.Config.Bucket),
-		Key:         aws.String(filename),
+		Key:         aws.String(client.Config.Path + filename),
```
Zeng1998, line 59: What about using `path.Join` instead of the string concatenation? (e.g., If I set the path variable to `path` but not `path/`)

web/src/components/CreateStorageServiceDialog.tsx
```diff
@@ -170,6 +171,17 @@ const CreateStorageServiceDialog: React.FC<Props> = (props: Props) => {
           onChange={(e) => setPartialS3Config({ secretKey: e.target.value })}
           fullWidth
         />
+        <Typography className="!mb-1" level="body2">
+          Bucket
```
Zeng1998, line 175: `Bucket` -> `Path`

web/src/components/CreateStorageServiceDialog.tsx
```diff
@@ -170,6 +171,17 @@ const CreateStorageServiceDialog: React.FC<Props> = (props: Props) => {
...
+          <span className="text-sm text-gray-400 ml-1">(Storage Path)</span>
+        </Typography>
+        <Input
+          className="mb-2"
+          placeholder="Path"
+          value={s3Config.path}
+          onChange={(e) => setPartialS3Config({ path: e.target.value })}
+          fullWidth
+        />
```
Zeng1998, line 184: I think the setting of the `path` should be after the `bucket name`,  WDYT?

Zeng1998 (changes requested):

web/src/components/CreateStorageServiceDialog.tsx
zhaochunqi: Make sense, let me change that.

web/src/components/CreateStorageServiceDialog.tsx
zhaochunqi: Nice catch! Didn't notice that!

plugin/storage/s3/s3.go
zhaochunqi: that's cool, but what if you want to set your key with `path/subpath`?

Zeng1998 (approved):
LGTM!

boojack (approved):

PR #1213, opened 2023-02-28
Fix: Markdown hyperlinks with parenthesis take first closing parenthesis as final

TheRealJd3:
Fixes #1162 by updating the link regex to pick up edge case with parentheses

PR #1251, opened 2023-03-04
revert: Fix: Markdown hyperlinks with parenthesis take first closing parenthesis as final

boojack:
Reverts usememos/memos#1213


TheRealJd3:
What is the bug caused?

boojack:
> What is the bug caused?

Please check the #1250 

I'd like to revert the change in this release and find a better way to fix it.

PR #1278, opened 2023-03-05
feat: automatically change language on first launch

CorrectRoadH:
There is a problem with the old locale initialization of memos. There are many logics in the middle, but it is always set to `en` in the end.

Let's review these code

``` Typescript
  // any page get locale value by useGlobalState
  xxx = useGlobalState();
  
  // the GlobalState Initial in App.tsx
  useEffect(....)
  
  // the initial Procssing
  export const initialGlobalState = async () => {
    const defaultGlobalState = {
      locale: "en" as Locale,
      appearance: "system" as Appearance,
      systemStatus: {
          ....
        customizedProfile: {
            .....
          locale: "en", // fisrt define
          ...
        },
      } as SystemStatus,
    };
  
    const { locale: storageLocale, appearance: storageAppearance } = storage.get(["locale", "appearance"]);
    if (storageLocale) {
      defaultGlobalState.locale = storageLocale;
    }
    .....  
    try {
      const { data } = (await api.getSystemStatus()).data;
      if (data) {
             ......
            locale: customizedProfile.locale || "en",  // second define
            appearance: customizedProfile.appearance || "system",
            externalUrl: "",
          },
        };
        defaultGlobalState.locale = customizedProfile.locale;  // the latest define
        defaultGlobalState.appearance = customizedProfile.appearance;
      }
    } catch (error) {
      // do nth
    }  
```
The setting sequence is: default value(en) storageLocale(browser cache) getSystemStatus(en, but always returns en)

```Typescript
const { data } = (await api.getSystemStatus()).data;
```

```Typescript
export function getSystemStatus() {
  return axios.get<ResponseObject<SystemStatus>>("/api/status");
}
```

let's look at the backend code
```go
// memos/server/system.go
	      systemStatus := api.SystemStatus{
	 Host:               hostUser,
	 Profile:            *s.Profile,
	 DBSize:             0,
	 AllowSignUp:        false,
	 DisablePublicMemos: false,
	 AdditionalStyle:    "",
	 AdditionalScript:   "",
	 CustomizedProfile: api.CustomizedProfile{
	  Name:        "memos",
	  LogoURL:     "",
	  Description: "",
	  Locale:      "en", // <- look here
	  Appearance:  "system",
	  ExternalURL: "",
	 },
	 StorageServiceID: 0,
	}
```
So language always ends up being defined as en.

I want to change the logic to the following:
- First Open: Detect Browser Language
- Later: use browser cache language

```Typescript
defaultGlobalState.locale = storageLocale || languageCodeCovert(i18n.language);
```
In First, The `storageLocale` is undefine, So the locale is the language that browser detect.
If the `storageLocale` is present or be changed by user. the locale will use it.

What is `languageCodeCovert`

Language codes detected by the browser such as `zh-TW`, `zh-HK` and `en-US`. But traditional Chinese is defined as 'hant', which is a very exception. So I have to handle it additionally. Although I don't think this method is elegant, but I have to do it, if you have an elegant method, please let me know.

```Typescript
// languageCodeCovert.ts
 export const languageCodeCovert = (codename: string): Locale => {
   if (codename === "zh-TW" || codename === "zh-HK") {
     return "hant";
   }
   const shortCode = codename.substring(0, 2);
   return shortCode as Locale;
 };
 ```
If the browser's language is not translated, the final will be "en". The processing code is as follows
```Typescript
  // ....
       ko: {
         translation: koLocale,
       },
     },
     fallbackLng: "en",
   });
 
```


CorrectRoadH:
I see the fail of Eslint-checks. I will change it immediately.

boojack:
Great work! A very detailed PR description.

CorrectRoadH:
> I see the fail of Eslint-checks. I will change it immediately.

I can't fix it. Because it didn't error in my computer.😢

web/src/store/module/global.ts
```diff
@@ -79,4 +81,4 @@ export const useGlobalStore = () => {
       store.dispatch(setAppearance(appearance));
     },
   };
-};
+};
```
boojack, line 84:
```suggestion
};

```

boojack:
> I can't fix it. Because it didn't error in my computer.😢

Fixed.

CorrectRoadH:
> > I can't fix it. Because it didn't error in my computer.😢
> 
> Fixed.

Thank You! 😀

web/src/utils/languageCodeCovert.ts
```diff
@@ -0,0 +1,7 @@
+export const languageCodeCovert = (codename: string): Locale => {
```
boojack, line 1: How about rename `languageCodeCovert` to `convertLanguageCodeToLocale`

boojack (approved):
LGTM

Chaseem44 (commented):
> OsJfaKDe3FpHCrEnSazjXZaC3edOS5LngUuK6pHsDOK-pnjmpddz7-

PR #1292, opened 2023-03-06
feat: support set openai api host

Wujiao233:
support #1280


api/system.go
```diff
@@ -19,4 +19,6 @@ type SystemStatus struct {
 	// Customized server profile, including server name and external url.
 	CustomizedProfile CustomizedProfile `json:"customizedProfile"`
 	StorageServiceID  int               `json:"storageServiceId"`
+	// OpenAI Host
+	OpenAIApiHost string `json:"openAIApiHost"`
```
boojack, line 23:
```suggestion
	// OpenAI API Host
	OpenAIAPIHost string `json:"openAIApiHost"`
```

server/system.go
```diff
@@ -52,6 +52,7 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
 				ExternalURL: "",
 			},
 			StorageServiceID: 0,
+			OpenAIApiHost:    "",
```
boojack, line 55: In backend, we should name it with `API`. `Api` is mainly using in frontend.

PR #1313, opened 2023-03-08
feat: Allow users to customize the refresh time for Daily Reviews

titanventura:


web/src/types/modules/setting.d.ts
```diff
@@ -10,6 +10,8 @@ interface Setting {
 interface LocalSetting {
   enableFoldMemo: boolean;
   enableDoubleClickEditing: boolean;
+  dailyReviewTimeOffsetHourValue: number;
+  dailyReviewTimeOffsetMinuteValue: number;
```
boojack, line 14: Since the time zone is divided according to the hour, so I think saving the hours should be enough. So we don't have to save minute.

titanventura: Changed time offset to only include hour

boojack (approved):

Issue #1162, opened 2023-02-25
Markdown hyperlinks with parenthesis in take first closing parenthesis as final


nickcowans:
### Describe the bug

For example `[Severance](https://en.wikipedia.org/wiki/Severance_(TV_series))` will create a link to "https://en.wikipedia.org/wiki/Severance_(TV_series", rather than "https://en.wikipedia.org/wiki/Severance_(TV_series)"

### Steps to reproduce

1. Paste `[Severance](https://en.wikipedia.org/wiki/Severance_(TV_series))` into a memo
2. Click Save
3. Click on the created link, it will be to "https://en.wikipedia.org/wiki/Severance_(TV_series"

HappyZ:
+1

also `[ [paper](https://arxiv.org/abs/2112.10752) ]` renders as [\[paper](https://arxiv.org/abs/2112.10752) ] which looks to be a similar issue

HappyZ:
maybe `/\[([^[\]\n]+)\]\(([^)\n]+)\)/g` for link extraction instead of `/\[(.*?)\]\((.+?)\)+/` here: https://github.com/usememos/memos/blob/1847756ade10debffee6e4b00271700d41a805d3/web/src/labs/marked/parser/Link.tsx#L9

took a quick look at https://github.com/usememos/memos/blob/1847756ade10debffee6e4b00271700d41a805d3/web/src/labs/marked/parser/PlainLink.tsx#L3 and it doesn't look right either. I'm thinking: `/(https?:\/\/\S+)(?:\s|$)/g`

TheRealJd3:
Changing the regex [here](https://github.com/usememos/memos/blob/1847756ade10debffee6e4b00271700d41a805d3/web/src/labs/marked/parser/Link.tsx#L9) to `/\[(.*?)\]\((.+?)\)$/` also fixes things

HappyZ:
for my own curiosity how does the parser work for single line string containing multiple links?

TheRealJd3:
It doesn't handle them - good point !

HappyZ:
Found https://stackoverflow.com/questions/67940820/how-to-extract-markdown-links-with-a-regex and modified to exclude /n as well: `/\[([^][\n]+)\](\(((?:[^()\n]+|(?2))+)\))/g`

this might work for both scenarios + multiple links in one string

boojack:
Current frontend parser is very simple and mainly using to get tags from a raw content.

[...]

PR #1315, opened 2023-03-08
fix: Update markdown hyperlink regex

BapRx:
Hey, I had the same issue as in #1162 and I noticed that the fix was revert. Here's my PR to fix the issue.

Before:

After:
image

### DIFF
--- a/web/src/labs/marked/parser/Link.tsx
+++ b/web/src/labs/marked/parser/Link.tsx
@@ -6,7 +6,7 @@ import BoldEmphasis from "./BoldEmphasis";
 import PlainText from "./PlainText";
 import { matcher } from "../matcher";
 
-export const LINK_REG = /\[(.*?)\]\((.+?)\)+/;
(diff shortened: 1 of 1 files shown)


web/src/labs/marked/parser/Link.tsx
```diff
@@ -6,7 +6,7 @@ import BoldEmphasis from "./BoldEmphasis";
 import PlainText from "./PlainText";
 import { matcher } from "../matcher";
 
-export const LINK_REG = /\[(.*?)\]\((.+?)\)+/;
+export const LINK_REG = /\[([^\]]+)\]\(([^)]+)\)/;
```
boojack, line 9:
```suggestion
export const LINK_REG = /\[(.*?)\]\(([^\s]+)\)/;
```

BapRx: I don't understand your update, your suggestion doesn't work:
image

boojack: I see

boojack (approved):

PR #1354, opened 2023-03-14
feat: update find resource with linked memo amount

boojack:


boojack:
https://github.com/usememos/memos/pull/1351

PR #1366, opened 2023-03-17
fix: the incorrectly height of grid row in safari

CorrectRoadH:
There is bug in the resource dashboard. 
CS_2023-03-15_at_23 19 412x
CS_2023-03-17_at_13 27 432x

The bug cause the incorrectly calc for grid row height in safari. to fix it using `h-auto` intead of  `h-full` in container div.

fixed
CS_2023-03-17_at_14 17 39@2x


boojack (approved):
LGTM

PR #1367, opened 2023-03-17
refactor: use function findMatchingParser to reduce duplicate code

unliar:


web/src/labs/marked/index.tsx
```diff
@@ -1,67 +1,75 @@
...
-        return (
-          <>
-            {parser.renderer(matchedStr)}
-            {marked(retainContent.slice(1), blockParsers, inlineParsers)}
-          </>
-        );
-      }
-    }
-  }
+type Parser = typeof blockElementParserList[0];
```
boojack, line 4: We can declare its structure instead of using `typeof blockElementParserList[0]`

boojack (approved):
LGTM

web/src/labs/marked/index.tsx
unliar, line 1: declare type Parser here

PR #1392, opened 2023-03-20
fix: daily review page & setting dialog style

xqa:
- Fix daily review sharing pop-up style 
- Fix language selection box style


boojack (approved):
LGTM

PR #1400, opened 2023-03-22
fix: the expand button's z-index is the same as Header

Steven-nagisa-Y:
Changed Header's z-index which is the same as expand button from z-10 to z-20.


boojack (approved):

PR #1418, opened 2023-03-27
feat: add setup cmd

dshemin:
Hi!

Thanks a lot for such good self-hosted application. But I have an issue: I tried to write an Ansible script for deploying Memos to my home server and want to create host user from my playbook. In order to be able to do it I made this command.


boojack:
Could you write a docs about usage scenario and steps to use it?

PR #1425, opened 2023-03-29
feat: The loaded resource for pagination

CorrectRoadH:
## Functions:
- The loaded resource for pagination
- Loading all resources before searching
- Loading all resources before clearing

## Screen
CS_2023-03-30_at_09 00 44@2x
CS_2023-03-30_at_09 00 49@2x


web/src/pages/ResourcesDashboard.tsx
```diff
@@ -150,9 +157,46 @@ const ResourcesDashboard = () => {
...
+      resolve();
+    }
+  };
+
   const handleSearchResourceInputChange = (query: string) => {
-    setQueryText(query);
-    setSelectedList([]);
+    // to prevent first tiger when page is loaded
+    if (query === queryText) return;
+    loadAllResources(() => {
```
boojack, line 196:
```
await fetchAllResources()
setQueryText()
// ...
```

boojack (approved):
LGTM, otherwise

web/src/pages/ResourcesDashboard.tsx
```diff
@@ -150,9 +157,46 @@ const ResourcesDashboard = () => {
...
+      } else {
+        setIsComplete(false);
+      }
+    } catch (error: any) {
+      console.error(error);
+      toast.error(error.response.data.message);
+    }
+  };
+
+  const loadAllResources = async (resolve: () => void) => {
```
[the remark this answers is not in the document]
CorrectRoadH: Use the callback format because both clear and search need to fetch all resources. Writing the same code twice is inelegant. And in the future, there may be other functions that need to fetch all resources. So I think this abstraction is appropriate.

web/src/pages/ResourcesDashboard.tsx
```diff
@@ -150,9 +157,45 @@ const ResourcesDashboard = () => {
...
+
+  const loadAllResources = async (resolve: (allResources: Resource[]) => void) => {
+    if (!isComplete) {
+      loadingState.setLoading();
+      try {
+        await resourceStore.fetchResourceList().then((allResources) => {
+          loadingState.setFinish();
+          setIsComplete(true);
+          resolve(allResources);
+        });
```
boojack, line 182:
```suggestion
        const allResources = await resourceStore.fetchResourceList();
        loadingState.setFinish();
        setIsComplete(true);
        resolve(allResources);
```

boojack (approved):
LGTM

PR #1472, opened 2023-04-04
feat: add front end unit test

CorrectRoadH:
## What does the PR do?
add some hook, redux, function and component and snap test and testing in CI.

## Test cases
Hook:
useLoading
- all path coverage

useToggle
- all path coverage

Reducer:
filter
- add a tag
- add a content
- add visibility limit
- add a new tag to replace old tag

Function:
- i18n

component:
- memos
- resourceCover
- resourceItemDropdown

### Why upgrade these deps: @types/node @types/react @types/react-dom
Because there is a bug that throws error when `pnpm build`.
[reactjs - Duplicate identifier 'LibraryManagedAttributes' - Stack Overflow](https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes)

### Why add 6000+ code?
It seem caused by pnpm-lock.yaml. I try remove the pnpm-lock.yaml and download from latest source. But it didn't have any effect that reduce the number of new code🤔 

### Why upgrade vite version  from 4.0.0 to 4.0.4(**Obsolete**)
Because vitest need vite 4.0.4 to fix the bug of display type does not exist
[typescript - Vitest defineConfig, 'test' does not exist in type 'UserConfigExport' - Stack Overflow](https://stackoverflow.com/questions/72146352/vitest-defineconfig-test-does-not-exist-in-type-userconfigexport).

### use result.current[0] instead of value to expect
normal hook test like
```typescript
  it("setFailed should set the state to failed", () => {
    const { result } = renderHook(() => useLoading());
    act(() => {
      result.current.setFinish();
    });
    expect(result.current.isLoading).toBe(false);
    expect(result.current.isFailed).toBe(false);
    expect(result.current.isSucceed).toBe(true);
  });
```
But useToggle return a list. So i can't call the function by name. So like 
```typescript
  it("useToggle should toggle the value to false", () => {
    const { result } = renderHook(() => useToggle(true));
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
    const [_, toggle] = result.current;
    act(() => {
      toggle();
    });
    expect(result.current[0]).toBe(false);
  });
```
if i use `const [value, toggle] = result.current;`and use `expect(value).toBe(false);`, the value wouldn't update the value.

### The bug when testing i18n
The implementation of `findNearestLanguageMatch` in this i18n seems to rely on the fallback  function of i18n when some special language such as zh. So some tests don't pass, but they work fine in the browser.


itschip:
@CorrectRoadH have you been thinking of picking this up again? If not, I can give it a go :)

CorrectRoadH:
> @CorrectRoadH have you been thinking of picking this up again? If not, I can give it a go :)

🤣 No. You can try it.

Issue #1408, opened 2023-03-24
可否在 memo 的 ... 处添加个“复制链接”？


jetsung:
### Is your feature request related to a problem?

是

### Describe the solution you'd like

希望在 `memo` 的 ... 处添加个“复制链接”
![image](https://user-images.githubusercontent.com/3362147/227546530-3b32805f-0f81-46e9-8c20-e705087e35f8.png)

因为当内容很长，比如代码分享时，需要滚动鼠标滚轮到下方才能找到“链接”，很不方便
![image](https://user-images.githubusercontent.com/3362147/227548340-76725071-c3ea-4b1e-8c47-fa9c9a3929d4.png)

HappyZ:
Might be even simpler to move the bottom row to the top :P 

I like the idea to add copy link though to simplify user journey

HappyZ:
I'll take it.

HappyZ:
Updated 1.12.1 and noticed #1498 this change removed copy link in each memo, which basically rolls back the original CL. Can I learn a bit more why it's removed again?

HappyZ:
@boojack

boojack:
@HappyZ The main reason is to reduce the number of buttons, while the link is part of sharing, so it is put in the share dialog. Just like Twitter does.

<img width="590" alt="image" src="https://user-images.githubusercontent.com/24653555/230803351-7a145ca4-f2a0-4c85-8049-dc079c0e6cc8.png">

PR #1474, opened 2023-04-05
chore: add "copy link" for each memo

HappyZ:
addressing issue #1408


web/src/components/ShareMemoDialog.tsx
```diff
@@ -179,6 +154,31 @@ const ShareMemoDialog: React.FC<Props> = (props: Props) => {
...
+              bgColor={"#F3F4F6"}
+              fgColor={"#4B5563"}
+              includeMargin={false}
+            />
+          </div>
+          <span className="time-text">{memo.createdAtStr}</span>
+          <div className="memo-content-wrapper">
+            <MemoContent content={memo.content} showFull={true} />
+            <MemoResources resourceList={memo.resourceList} />
+          </div>
```
boojack, line 180: Is this expected? The watermark is on top of content.
image

HappyZ: Yea I moved it so for really long content, if you simply want to do image or scan qr code it can be on the top without scrolling down. Let me know what you think.

boojack: The content is more important than the QR code. Maybe we should keep the origin order in this PR and think of a better layout style.

HappyZ: To confirm, you mean no change to ShareMemoDialog at all, or still keep the first row "visibility, image, link"?

boojack: Yeah

boojack (approved):
LGTM

Issue #1410, opened 2023-03-24
Option to Disable OpenAI


Tally:
Please, allow disabling of OpenAI integration in settings. It is not necessary and superfluous for many use-cases.

Zeng1998:
https://github.com/usememos/memos/pull/1539#issuecomment-1510027883

PR #1515, opened 2023-04-11
feat: add option to hide ask ai

joaonunomota:
Add ability to hide Ask AI from users via system settings, as suggested in #1410.


web/src/components/Header.tsx
```diff
@@ -107,13 +109,15 @@ const Header = () => {
...
           {!isVisitorMode && (
             <>
-              <button
-                id="header-ask-ai"
-                className="px-4 pr-5 py-2 rounded-lg flex flex-row items-center text-lg text-gray-800 dark:text-gray-300 hover:bg-white hover:shadow dark:hover:bg-zinc-700"
-                onClick={() => showAskAIDialog()}
-              >
-                <Icon.Bot className="mr-3 w-6 h-auto opacity-70" /> {t("ask-ai.title")}
-              </button>
+              {showAskAI && (
```
boojack, line 112: How about hide it when openai config is undefined or openai api key config is undefined? So we don't need to introduce a new setting item.

web/src/components/Settings/SystemSection.tsx
```diff
@@ -104,18 +94,22 @@ const SystemSection = () => {
...
   };
 
   const handleSaveOpenAIConfig = async () => {
     try {
       await api.upsertSystemSetting({
         name: "openai-config",
-        value: JSON.stringify(openAIConfig),
+        value: JSON.stringify(state.openAIConfig),
       });
+      globalStore.setSystemStatus({ openAIConfig: state.openAIConfig });
```
joaonunomota, line 112: Added after request to backend to avoid mismatch with frontend in case of an error. It's done the other way round for `disablePublicMemos`, does anyone know is there a reason for that?

web/src/components/Header.tsx
joaonunomota: I put the openai config in with the rest of the global state and I'm checking it's not empty in showAskAI

web/src/components/Settings/MyAccountSection.tsx
```diff
@@ -46,16 +48,18 @@ const MyAccountSection = () => {
...
       </div>
-      <div className="section-container openapi-section-container">
-        <p className="title-text">Open API</p>
-        <p className="value-text">{openAPIRoute}</p>
-        <span className="btn-danger mt-2" onClick={handleResetOpenIdBtnClick}>
-          {t("common.reset")} API
-        </span>
-        <div className="usage-guide-container">
-          <pre>{`POST ${openAPIRoute}\nContent-type: application/json\n{\n  "content": "Hello #memos from ${window.location.origin}"\n}`}</pre>
+      {showAskAI && (
```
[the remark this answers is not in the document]
joaonunomota: Remove the code changes to Open ID API

web/src/components/Settings/SystemSection.tsx
boojack: I think both works. The frontend will "refresh" state before or after backend api request finished.

boojack (approved):

Issue #563, opened 2022-11-25
Unexpected empty lines when copying-pasting


dmitriisurmin:
### Describe the bug

Hello!

When I copy text in memo using selection, in pasting blank lines appear between. Expected that this to work like the built-in copy function works now, without gaps.

https://user-images.githubusercontent.com/118022040/203881939-a18b5c73-d028-4599-9570-5509b1799c2a.mov

### Steps to reproduce

1. Select text in a memo
2. Copy it with `ctrl+c` or using context menu
3. Paste in some input field/text editor

### Screenshots or additional context

I didn't dive deep into what goes to clipboard, but in some text editors it can be pasted with no empty lines, some not. In my glance, this is because of memo markup. Every line goes into the own paragraph.
```html
<p>First line</p>
<p>Second line</p>
```
Which, without clipboard control, leads to this issue.

Zeng1998:
I think the extra empty lines are added when selected and copied from the rich-text (including in other websites).
So maybe it's better to handle it when pasting ?

dmitriisurmin:
I mentioned that some editors **not** insert extra lines. But probably they are doing it **wrong** :) most of the apps add these lines.

I think there is another option if extra lines don't come from the rich-text. I would suggest don't split each line into a separate `<p>` element, but put them together in one paragraph, maybe, or split with `<br>`. It will not copy extra lines. As I know, many web markdown viewers do like that.
```html
<p>
First line
Second line
</p>
```

PR #1524, opened 2023-04-13
Unexpected empty lines when copying-pasting #563

ameerturk96:
I looked into the issue and I found out when you copy from the memos using keyboard it added an extra line between lines, so what I did was manipulating the innerText when copying using crtl+c and removing the extra line added. looking forward to your feedback.

### DIFF
--- a/web/src/components/Editor/Editor.tsx
+++ b/web/src/components/Editor/Editor.tsx
@@ -24,7 +24,7 @@ interface Props {
 }
 
 const Editor = forwardRef(function Editor(props: Props, ref: React.ForwardedRef<EditorRefActions>) {
-  const { className, initialContent, placeholder, fullscreen, onContentChange: handleContentChangeCallback } = props;
+  const { className, initialContent, placeholder, fullscreen, onPaste, onContentChange: handleContentChangeCallback } = props;
   const editorRef = useRef<HTMLTextAreaElement>(null);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


CorrectRoadH:
I think to listen Copy Event instead of the Ctrl+C Key-Down Event is a better way to fix the bug.
Such as https://developer.mozilla.org/en-US/docs/Web/API/Element/copy_event

ameerturk96:
> I think to listen Copy Event instead of the Ctrl+C Key-Down Event is a better way to fix the bug. Such as https://developer.mozilla.org/en-US/docs/Web/API/Element/copy_event WDYT?
> 
> I saw your commit, first used the paste event, and then changed it to the ctrl+c event. Why is this? Is there a reason I haven't thought of?

I tried to use the event listener, but because the text is being manipulated before rendering or other reasons that I might don’t know of, the event object only returns the first line "copied", I thought this might be something you know of

CorrectRoadH:
> > I think to listen Copy Event instead of the Ctrl+C Key-Down Event is a better way to fix the bug. Such as https://developer.mozilla.org/en-US/docs/Web/API/Element/copy_event WDYT?
> > I saw your commit, first used the paste event, and then changed it to the ctrl+c event. Why is this? Is there a reason I haven't thought of?
> 
> I tried to use the event listener, but because the text is being manipulated before rendering or other reasons that I might don’t know of, the event object only returns the first line "copied", I thought this might be something you know of

I tried to achieve the same effect based on your idea. 
``` typescript
// MemosList.tsx
...
  useEffect(() => {
    addEventListener("copy", handleCopy);
    return () => {
      removeEventListener("copy", handleCopy);
    };
  }, []);

  const handleCopy = (event: ClipboardEvent) => {
    event.preventDefault();
    const rawStr = document.getSelection()?.toString();
    if (rawStr !== undefined) {
      const splitString = rawStr.split("\n\n");
      const text = splitString.join("\n");
      copy(text);
    }
  };
...
```
The code works fine! This solution is based on yours, but covers more cases (like copy and ctrl+c). WDYT👀

CorrectRoadH (changes requested):

PR #1526, opened 2023-04-13
feat: improve i18n support as a whole

lincolnthalles:
- Remove dayjs in favor of /helpers/datetime.ts, which uses Intl.DateTimeFormat and Date. Dayjs is not exactly i18n friendly and has several locale related opened issues.

- Move/refactor date/time code from /helpers/utils.ts to /helpers/datetime.ts.

- Fix Daily Review weekday not changing according to selected date.

- Localize Daily review weekday and month.

- Load i18n listed strings from /locales/{locale}.json in a dynamic way. This makes much easier to add new locales, by just adding a properly named json file and listing it only in /web/src/i18n.ts and /api/user_setting.go.

- Fallback languages are now set in /web/src/i18n.ts.

- Full language codes are now preffered, but they fallback to 2-letter codes when not available.

- The locale dropdown is now populated dynamically from the available locales. Locale names are populated by the browser via Intl.DisplayNames(locale).

- /web/src/i18n.ts now exports a type TLocale from availableLocales array. This is used only by findNearestLanguageMatch(). As I was unable to use this type in ".d.ts" files, I switched the Locale type from /web/src/types/i18n.d.ts to string.

- Move pretty much all hardcoded text strings to i18n strings.

- Add pt-BR translation.

- Remove site.ts and move its content to a i18n string.

- Rename zh.json to zh-Hans.json to get the correct language name on selector dropdown.

- Remove pt_BR.json and replace with pt-BR.json.

- Some minor layout spacing fixes to accommodate larger texts.

- Improve some error messages.

There are lots of small changes, but I also took some time testing and it should work fine.


CorrectRoadH (approved):
LGTM

web/package-lock.json
boojack, line 1: Any reason about creating this file and  updating `yarn.lock`.

lincolnthalles: Not really, yarn did this by itself

web/yarn.lock
boojack, line 1: I will regenerate it in the next PRs.

boojack (approved):

boojack:
BTW, what's your yarn version?

```bash
yarn -v
```

lincolnthalles:
> BTW, what's your `yarn` version?
> 
> ```shell
> yarn -v
> ```
It's the latest, 3.5.0, and node = 18.15.0

PR #1531, opened 2023-04-15
fix: toast overload max size error

Fog3211:
## Fix the issue of no error prompt when the file size exceeds the limit.

image
image


boojack (approved):

PR #1550, opened 2023-04-16
Add auto collapse feature for all memos issue #1463

polyeung:
I added a auto collapse switch in preference setting,
image

By switching off, all memos will not collapse. If switching on, all memos that exceed certain words limit will collapse as usual.


CorrectRoadH (changes requested):
LGTM overall

polyeung:
Thanks for your reply! I have sorted out the style check :)

PR #1554, opened 2023-04-17
fix: the broken install docs link

m3nu:
I re-arranged the install docs a bit to separate them into Docker and hosted options:

https://usememos.com/docs

This also breaks the said link in the README. Most appropriate alternative would be this:

https://usememos.com/docs#installation


boojack (approved):
LGTM

PR #1562, opened 2023-04-18
feat: Allow linking memos from other memos

palmenhq:
Hello,
Thanks for a fun and good app! Here's a suggestion on how one could reference memos from other memos. This is my first PR here and I didn't find much docs about contributing, so  I figured I'd just give it a go and see what you think 👍 no harsh feelings if you don't like it or want a different approach - it's completely unsolicited haha.

Changes worth noting
- [x] The killer feature: Created an inline markdown parser that links "memo references" (`m/123`) to actual memos
- [x] Refactored search input field to be reusable
- [x] Created an "non-interactive" version of the markdown parser to allow the memo content to be used as buttons
- [x] Changed the editor icon for "resources" to match the sidebar icon and used the "file" one for selecting memos instead 

What's left before this PR is complete
- [ ] Updating locales (I didn't quite understand how to do this and didn't find any docs, cc @boojack i guess?)

Future suggestions
- [ ] Creating an index of referenced memos, so that I can see on `m/2` that `m/1` mentions it. Seems like a larger thing though so leaving it out of scope for now.

memo

### DIFF
--- a/web/src/components/Dialog/BaseDialog.tsx
+++ b/web/src/components/Dialog/BaseDialog.tsx
@@ -7,6 +7,7 @@ import store from "@/store";
 import { useDialogStore } from "@/store/module";
 import theme from "@/theme";
 import "@/less/base-dialog.less";
+import { BrowserRouter } from "react-router-dom";
 
 interface DialogConfig {
   dialogName: string;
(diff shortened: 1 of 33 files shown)


boojack:
Thanks so much for the great PR! 

Actually, it has been planned for long time. https://github.com/orgs/usememos/projects/5?pane=issue&itemId=16186793 

Your solution is to put the relationship in the content, just like the `tag`. This is simple and easy to understand. But it's not conducive to search. 

So we need to put the information about this relationship into a separate table in database. We'd like to build a relationship between `memo` <-> `memo` as same as [`memo` <-> `resource`](https://github.com/usememos/memos/blob/main/store/db/migration/prod/LATEST__SCHEMA.sql#L87).

palmenhq:
@boojack Oh cool - thanks for the insight and thanks for looking at my PR! I left the relationship out to keep the size of the PR down, but if you want I'm happy to amend with a relationship table and such 👍

boojack:
@palmenhq Please keep going if you like. The UI/UX is very nice. And as you said that you'd like to keep the size of the PR down, so you can make a separate PR for a schema change. We can discuss its design deeply.

PR #1575, opened 2023-04-20
feat: memos relation backend implment

CorrectRoadH:
- [x] Create memos relation table
- [x] Create a find relation memos id in store
- [x] An additional API to get relation memos id
- [x] add unit test

## What does the PR do:
- four new API
  - POST `/memo/relation` to create a relation between two memos
  - GET `/memo/relation/:memoId` to get a memos all relation
  - DELETE `/memo/relation/:memoId/:relatedMemoId/:type` to delete a relation 
  - DELETE `/memo/relation` to delete all null relation(such as a memos of a relation already be deleted)
- test case
  - add a relation
  - assert relation num
  - delete a relation 
  - assert relation num
  - add a relation
  - delete a memos
  - clean null relation
  - assert relation num
  - assert an error for adding a relation with the invalid relation type

## Why relation test in memo test
🤔 I think it is not bad. relation test must contain multiple memos operator. So to test relation when testing memo, I think can reduce very unnecessary code.


boojack:
@palmenhq FYI

palmenhq:
hey @CorrectRoadH, awesome stuff! I was curious if you have any ideas how to create the relationships from a user and/or consumer perspective?

Perhaps we should join forces as i recently made #1562? An idea could be to try parsing out any mention of `m/NNN` (representing a memo link, similar to how `#tags` are used) and attach any found ids to the `POST /api/memo` request payload? (similar to how `resourceIdList` works)

@CorrectRoadH @boojack any thoughts on that approach?

boojack:
@palmenhq If we put the relation in memo content, it will be hard to process in the future. So I suggest we should use the same way as resource&memo. 👇

image

CorrectRoadH:
Implement in https://github.com/usememos/memos/pull/1618.🥳

PR #1578, opened 2023-04-20
fix: only delete last file when select multiple files #1576

CorrectRoadH:
image
Because the `handleCheck` function be `useMemo`👀


web/src/hooks/useEvent.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  useIsoMorphicEffect(() => {
+    cache.current = value;
+  }, [value]);
+
+  return cache;
+}
+
+// TODO: Add React.useEvent ?? once the useEvent hook is available
+const useEvent = function useEvent<F extends (...args: any[]) => any, P extends any[] = Parameters<F>, R = ReturnType<F>>(
```
boojack, line 18:
```suggestion
function useEvent<F extends (...args: any[]) => any, P extends any[] = Parameters<F>, R = ReturnType<F>>(
```

boojack (approved):
LGTM

PR #1585, opened 2023-04-22
chore: add memo_relation

boojack:
Add a new table `memo_relation`:

* `memo_id` is the id of source memo;
* `related_memo_id` is the id of related memo;
* `type` is the type of relation: "REFERENCE" | "ADDITIONAL"


boojack:
@palmenhq @CorrectRoadH FYI

PR #1602, opened 2023-04-26
reappearing of dialog should add body scrolling class.

deEshu2002:
When dialog is reappeared after being in a hidden state. Then reappeaning should block further body scrolling for consistent UX.


boojack (approved):

PR #1605, opened 2023-04-26
fix: memo content availability for visitor mode

deEshu2002:
MemoContent being part of "Memo" Component and Memo part of "Explore" page. MemoContent throws error and doesn't render without User states. Since user is "undefined" for visitor mode, destructuring user states to get localSetting is not possible. Hence adding further guard to check for visitorMode and making autoCollapse state "true" by default is the right choice.

### DIFF
--- a/web/src/components/MemoContent.tsx
+++ b/web/src/components/MemoContent.tsx
@@ -32,7 +32,8 @@ const MemoContent: React.FC<Props> = (props: Props) => {
 
   //variable for auto-collapse
   const userStore = useUserStore();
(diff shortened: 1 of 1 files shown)


deEshu2002:
Hey boojack, Uffizzi deployment (https://app.uffizzi.com/github.com/usememos/memos/pull/1602) for my previous pull request #1602 resulted in the error caused by the MemoContent component and failed to render. In this pull request i have provided the possible patch for that bug. After reviewing this can you check #1602 too. Thank you.

boojack (approved):

PR #1620, opened 2023-05-02
chore: check allow sign up setting in sso

boojack:


Shuenhoy:
Hi @boojack, I think sometimes it's useful to disable signup, but still allow users from SSO. For example, with a custom OAuth, all users from there should already be verified.

Issue #1608, opened 2023-04-27
置顶状态的Memo编辑后接口返回值丢失置顶状态


Rabithua:
## 实际请求API
<img width="360" alt="截屏2023-04-27 13 12 51" src="https://user-images.githubusercontent.com/34543831/234765244-b295cc4d-3808-41bf-86b1-7b95da03c986.png">

## 返回结果
<img width="589" alt="截屏2023-04-27 13 10 38" src="https://user-images.githubusercontent.com/34543831/234765097-a2714d9b-6ee2-4a20-adfe-676fd3509980.png">

**然而实际上这条Memo状态一直是置顶的**

### Steps to reproduce

如上图

CorrectRoadH:
It has been fixed in #1629

PR #1629, opened 2023-05-03
fix: return pinned status after edit

CorrectRoadH:
resolving #1608
## What does the PR do?
add an additional query after PATCH. to query the pinned status and add it to the response value.


boojack (changes requested):

CorrectRoadH:
@boojack PTAL

boojack (approved):
LGTM

PR #1636, opened 2023-05-05
feat: add preliminar Windows support

lincolnthalles:
Add preliminar Windows support for both development and production environments.

Default profile.Data will be set to `C:\ProgramData\memos` on Windows. Folder will be created if it does not exist, as this behavior is expected for Windows applications.

System service installation can be achieved with third-party tools, explained in docs/windows-service.md.

Not sure if it's worth using https://github.com/kardianos/service to make service support built-in.

This could be a nice addition alongside #1583 (add Windows artifacts)


boojack (approved):

PR #1641, opened 2023-05-08
feat: add support for image thumbnail

athurg:
While publishing a large number of HD images, the memoe will take a lot of time to load.

So I just generate a smaller thumbnail while uploading image resources, and only load the thumbnail in the memo. If click on the image, it will load the original HD image lazy.

### DIFF
--- a/server/resource.go
+++ b/server/resource.go
@@ -115,6 +115,8 @@ func (s *Server) registerResourceRoutes(g *echo.Group) {
 				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to unmarshal storage service id").SetInternal(err)
 			}
 		}
+
+		publicID := common.GenUUID()
 		if storageServiceID == api.DatabaseStorage {
 			fileBytes, err := io.ReadAll(sourceFile)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


Zeng1998:
I think dynamic compression when loading the image is even more expensive than loading the full image directly.

Zeng1998 (commented):
Do we need to delete the thumbnail in the delete api?

https://github.com/usememos/memos/blob/main/server/resource.go#L273

athurg:
> Do we need to delete the thumbnail in the delete api?
> 
> https://github.com/usememos/memos/blob/main/server/resource.go#L273

You're right, I have fixed it.

server/resource.go
```diff
@@ -127,24 +129,40 @@ func (s *Server) registerResourceRoutes(g *echo.Group) {
...
+
+			fileBytes, err := io.ReadAll(sourceFile)
 			if err != nil {
-				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to create file").SetInternal(err)
+				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to read resource file").SetInternal(err)
 			}
-			defer dst.Close()
-			_, err = io.Copy(dst, sourceFile)
+
+			err = os.WriteFile(filePath, fileBytes, 0666)
```
Zeng1998, line 145: We don't need to modify the original filename.  It should be handled by the path template.

Zeng1998 (changes requested):

Zeng1998 (commented):
I still recommend that it is best not to modify the original filename, this will break the path template settings.

athurg:
> I still recommend that it is best not to modify the original filename, this will break the path template settings.

I have changed the thumbnail's naming. Now the original filename was untouched, and the thumbnail file store in a separate cache dir with resource's PublicID as it's filename.

Zeng1998 (approved):
LGTM

Jonnyan404:
> > 我还是建议最好不要修改原来的文件名，这样会破坏路径模板的设置。
> 
> 我已经更改了缩略图的命名。现在原始文件名没有改变，缩略图文件存储在一个单独的缓存目录中，文件名是资源的 PublicID。

为啥昨晚的版本更新,我的不显示缩略图?
还是说必须新上传的才能显示?

athurg:
> > > 我还是建议最好不要修改原来的文件名，这样会破坏路径模板的设置。
> > 
> > 
> > 我已经更改了缩略图的命名。现在原始文件名没有改变，缩略图文件存储在一个单独的缓存目录中，文件名是资源的 PublicID。
> 
> 为啥昨晚的版本更新,我的不显示缩略图? 还是说必须新上传的才能显示?

It's not a bug. The Resource Dashboard doesn't support the show the thumbnail before, it only shows an icon.

I have add thumbnail support for the Resource Dashboard, please move to #1666 .

PR #1646, opened 2023-05-09
feat: add max upload size setting to UI & UI improvements

### LINKED ISSUE #1577: overload max size: 32MB
### Describe the bug

overload max size: 32MB

### Steps to reproduce

1. Go to "Resources"
2. Click “Add” icon
3. Upload File

image

lincolnthalles:
- feat: add max upload size setting to UI

- feat: max upload size setting is checked on UI during upload, but also enforced by back-end.

- feat: add HelpButton component with some links to official docs were appropriate

- remove LearnMore component in favor of the much more versatile HelpButton (can show tooltips, links or a dialog)

- fix: overflowing mobile layout for Create SSO, Create Storage and other Settings dialogs

- refactor (back-end): change some if/else to switch statements

- refactor (back-end): inline some err == nil checks

! Existing databases without the new setting 'max-upload-size-mib' will show an upload error, but this can be user-fixed by simply setting the value on system settings UI.

Unfortunately this conflicts with #1640, but I think it's a better approach to edit the upload limit on UI rather than command line.
image

Closes #1577


server/system.go
```diff
@@ -44,6 +44,7 @@ func (s *Server) registerSystemRoutes(g *echo.Group) {
 			AllowSignUp:        false,
 			IgnoreUpgrade:      false,
 			DisablePublicMemos: false,
+			MaxUploadSizeMiB:   32,
```
boojack, line 47: The default value is 32. So we don't need to add 

* store/db/migration/prod/0.12/06__default_upload_limit.sql
* store/db/seed/10005__system_setting.sql

store/system_setting.go
```diff
@@ -94,6 +94,15 @@ func (s *Store) FindSystemSetting(ctx context.Context, find *api.SystemSettingFi
 	return systemSettingRaw.toSystemSetting(), nil
 }
 
+func (s *Store) GetSystemSettingValueOrDefault(ctx *context.Context, find api.SystemSettingName, defaultValue string) string {
+	if ss, err := s.FindSystemSetting(*ctx, &api.SystemSettingFind{
```
boojack, line 98: ss -> setting

boojack (approved):
LGTM, otherwise

CorrectRoadH (approved):
LGTM

boojack:
And please update the icon into a normal size. It's a bit too big.

image

server/system.go
lincolnthalles: done

store/system_setting.go
lincolnthalles: done

boojack (approved):
LGTM, thanks for your contribution.

server/system.go
Zeng1998: > The default value is 32. So we don't need to add
> 
> * store/db/migration/prod/0.12/06__default_upload_limit.sql
> * store/db/seed/10005__system_setting.sql

There is a bug in https://github.com/usememos/memos/blob/main/server/resource.go#L73 , because the default value of MaxUploadSizeMiB isn't set in the database.

lincolnthalles: > 

Indeed.

In this case, my initial approach to seed and migrate the database was a little better.

In databases without the setting "max-upload-size-mib", the backend will return "File size exceeds allowed limit of 0 MiB".

Changing the default value "32" to something else in memos "System" section solves this, as it will create the missing setting key.

Simply changing that "0" parameter in https://github.com/usememos/memos/blob/main/server/resource.go#L73 to "32" works, but this breaks server enforcing the setting if uploads are explicitly disabled by setting max size to "0".

As this can be perceived as security issue (filling all disk space would break server functionality), I will look into this, hoping to find a better solution.

PR #1647, opened 2023-05-10
fix: wrong position of UsageStatItem's popup

athurg:
After scroll the page, the popup div of the UsageStatItem was wrong, like the screenshot below:

bug


boojack (approved):
LGTM

PR #1651, opened 2023-05-11
Use marked library and mermaid in memo display.

Jbbouille:
In order to do this:
image


Jbbouille:
I didn't saw it would conflict with https://github.com/usememos/memos/pull/1562. I don't know what I should do. Should i wait for the other PR to be merged before doing this ?

Jbbouille:
whoops I didn't saw that I breaked the Todo buttons. Sorry !

Jbbouille:
Okay it's fixed now.

I hope I have forgot nothin more.

Zeng1998:
marked? a big change.

boojack:
@Jbbouille Any idea to display empty lines with marked?

image

Jbbouille:
@boojack based on [this](https://stackoverflow.com/questions/33191744/how-to-add-new-line-in-markdown-presentation) you have to make two or more space at the end of a line.

CorrectRoadH:
🤔 I think the PR breaks the compatibility, which is hard to accept.

Jbbouille:
@CorrectRoadH can you tell me what it brokes, and I will correct it.

CorrectRoadH:
> @CorrectRoadH can you tell me what it brokes, I will fix it.

the render effect:
the same text
```markdown
How to learn CSS?
- test 1
- test 2
Conclusion  
```
in old 
CleanShot 2023-05-13 at 07 48 54@2x

in the PR
CleanShot 2023-05-13 at 07 49 12@2x

CorrectRoadH:
> @CorrectRoadH for the style it is fixed.
> 
> But for the "Conclusion" item which is not in the list, I don't know what is desirable.
> 
> If we want to display [markdown](https://spec-md.com/) or even [common markdown](https://commonmark.org/), I think this is the right render, but if we want to display some custom language like markdown I should fix it.
> 
> In the memos website it is advertise that it's [Plain text with Markdown](https://usememos.com/feature/plain-text).
> 
> What should I do ?

The PR is OK for me. I think `marked` can reduce the cost of maintenance. But there are different opinions in the team, so we are still discussing. 👀

clipcarl:
Hi, I'm using this branch of memos and I've noticed several issues. I've attached a patch for them.

First, memos is overly optimistic / aggressive about what types of text it will convert to hash tags. This is actually also a problem in the main branch but it breaks more stuff in this branch. For example, in this branch apostrophes (') are not displayed correctly in any memo because the underlying code of this branch replaces them with a numeric HTML entity (`&#39;`) and numeric HTML entities are incorrectly translated into hash tags which breaks their rendering. (The problem with the rendering of numeric HTML entities is also in the main branch but the main branch does replace apostrophes with them.)

Here are a few examples of some text that in my opinion should **not** be converted into hash tags but **are** anyway breaking rendering or the likely intent of the user:
- `Rule #1: You don't talk about ...`
- `#2`
- `https://example.com/#header3`
- `First rate A#1 quality`
- `&#33;`

I've solved this problem by using a much stricter regex to find tags:
`(?<=^|\p{White_Space})#(([\p{L}_][\p{L}\p{N}_]*)|(\p{N}+\p{L}[\p{L}\p{N}_]*))`

This means that in order to be matched tags must follow one of these two formats:
- `<1 letter or underscore><0 or more letters, numerals or underscores>`
- or `<1 or more numerals><1 letter><0 or more letters, numerals or underscores>`

In both cases the tag must be preceded by the beginning of the line or a whitespace in order to work around some of the rendering issues above.

Nevertheless the regex still matches every tag I had in some markdown files I've collected.

I'm using unicode-aware metacharacters in the regex so that it should work properly in every language even those for which the regular regex metacharacters don't work.

Another issue with this branch is that it does not always properly render tags or set the proper tag filter when they're clicked. The following lines demonstrate the problem:
```
#tag
#tag2
#tag_3
```
The second and third tags do not render properly and when clicked do not set the proper filter. I've rewritten and simplified the `paragraph()` function in `web/src/labs/render.ts` to correct the logic. It should also be much faster.

With this patch this branch seems pretty usable to me now (though I haven't tested very long).

[memos-use-marked-library-20230522-cet.patch.gz](https://github.com/usememos/memos/files/11541624/memos-use-marked-library-20230522-cet.patch.gz)

Jbbouille:
@clipcarl Thanks for the comment, I have hadded your patch on this merge request.

clipcarl:
Hi, thanks for looking at my patch so quickly and getting it in! :-)

I notice that one change I made didn't seem to make it in. Right now LINE_BREAK_REG is:
`export const LINE_BREAK_REG = /\n|\r|\r\n/g;`
but I think it should be:
`export const LINE_BREAK_REG = /\r\n|\n|\r/g;`

This is because if I'm not while regexes are in general greedy, regex alternation is opportunistic; the first left to right match wins. So with the first form `\r\n` will never be matched.

Jbbouille:
Thank you @clipcarl I didn't knew that.

clipcarl:
Hi @boojack . Will the the new gomark plugin support tables? That's a feature that I absolutely need!

Is there any architecture documentation or notes on how gomark will be implemented in memos to make it easier for people to jump in and contribute? In general I feel that memos could use better source documentation.

Thanks!

clipcarl:
Pinging @Jbbouille about possibly resurrecting this again as it seems table support is not forthcoming.

PR #1664, opened 2023-05-15
fix: File size exceeds allowed limit of 0 MiB

lincolnthalles:
fix: File size exceeds allowed limit of 0 MiB

This could happen in databases without "max-upload-size-mib" setting.

Now, both the front-end and the back-end will start with a default limit of 32 MiB, even if the key is absent.

It is still possible to disable uploads by setting the value to 0.


boojack (approved):
LGTM

PR #1687, opened 2023-05-20
Fix fail to open file while generate thumbnail

athurg:
Thumbnail cache path haven't been create before write file, which will cause the new guys fail to get thumbnail always.


boojack (approved):
LGTM

PR #1699, opened 2023-05-21
fix: md without search icon

CorrectRoadH:
Old:
CleanShot 2023-05-21 at 14 38 33@2x
When the width is md. the user interface has no search button and no home side.

## What does PR do:
- show the search button when the width is md
- fix error z-index
otherwise:
CleanShot 2023-05-21 at 14 44 34@2x

New:
CleanShot 2023-05-21 at 14 45 18@2x

### DIFF
--- a/web/src/components/Header.tsx
+++ b/web/src/components/Header.tsx
@@ -32,7 +32,7 @@ const Header = () => {
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM

PR #1707, opened 2023-05-22
Fix system memo visibility was replace by user's setting

athurg:
The memos visibility in `System Setting` must priority than the user'.

For example, while the user
- Set `Disable public memos` in `Admin-System-Settings` to true
- Set `Default memo visibility` in `Basic-Preferences-Preferences` to `Public to everyone`.

The new memos should use only `Private` visibility. But it shows `Public to everyone` on the memos create dialog, and not selectable.

Memos create dialog

preferences setting

system setting


boojack (approved):
LGTM

PR #1713, opened 2023-05-22
Replace filename to publicid

athurg:
Currently, we use filename in the storage path. Because the filename comes from the user's input, it may cause some problems, such as:

- Same filename will replace each other.
- Characters in filename is difference between all OS. Valid name in Windows may cause problems in Linux.

So I replace the origin filename into a UUID which exists in the code. The new filename used to storage only, and the origin filename still save in the database for review and shown in UI.

### DIFF
--- a/server/resource.go
+++ b/server/resource.go
@@ -143,14 +143,10 @@ func (s *Server) registerResourceRoutes(g *echo.Group) {
 				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find local storage path setting").SetInternal(err)
 			}
 			localStoragePath := "assets/{publicid}"
-			if systemSettingLocalStoragePath != nil {
-				var s string
-				err = json.Unmarshal([]byte(systemSettingLocalStoragePath.Value), &s)
+			if systemSettingLocalStoragePath != nil && systemSettingLocalStoragePath.Value != "" {
+				err = json.Unmarshal([]byte(systemSettingLocalStoragePath.Value), &localStoragePath)
 				if err != nil {
 					return echo.NewHTTPError(http.StatusInternalServerError, "Failed to unmarshal local storage path setting").SetInternal(err)
-				}
-				if s != "" {
-					localStoragePath = s
 				}
 			}
 			filePath := filepath.FromSlash(localStoragePath)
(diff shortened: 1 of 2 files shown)


server/resource.go
```diff
@@ -142,19 +142,24 @@ func (s *Server) registerResourceRoutes(g *echo.Group) {
 			if err != nil && common.ErrorCode(err) != common.NotFound {
 				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find local storage path setting").SetInternal(err)
 			}
-			localStoragePath := "assets/{timestamp}_{filename}"
+			localStoragePath := "assets/{publicid}"
 			if systemSettingLocalStoragePath != nil {
-				err = json.Unmarshal([]byte(systemSettingLocalStoragePath.Value), &localStoragePath)
+				var s string
+				err = json.Unmarshal([]byte(systemSettingLocalStoragePath.Value), &s)
```
boojack, line 148: Why not just use `localStoragePath` directly?

boojack (changes requested):

server/resource.go
```diff
@@ -142,19 +142,24 @@ func (s *Server) registerResourceRoutes(g *echo.Group) {
 			if err != nil && common.ErrorCode(err) != common.NotFound {
 				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to find local storage path setting").SetInternal(err)
 			}
-			localStoragePath := "assets/{timestamp}_{filename}"
+			localStoragePath := "assets/{publicid}"
 			if systemSettingLocalStoragePath != nil {
```
boojack, line 146:
```suggestion
			if systemSettingLocalStoragePath != nil && systemSettingLocalStoragePath.Value != "" {
```

server/resource.go
athurg: There's a bug if use `localStoragePath` directly.

If I set a valid `localStoragePath` in System Settings, and then reset it to empty. The expect value of `localStoragePath` should be the default value of `assets/{publicid}`. But if you directly unmarshal to `localStoragePath`, it will be a value of empty string, instead of default `assets/{publicid}`

boojack (approved):
LGTM

PR #1730, opened 2023-05-24
fix: infinite loop while daily memos more than DEFAULT_MEMO_LIMIT

athurg:
While the memos count more than DEFAULT_MEMO_LIMIT, there is a bug in DailyReview page, which will caused infinite request to `/api/memo`


boojack (approved):
LGTM

PR #1738, opened 2023-05-25
vscode setting enforcement for go111module="on" and installability of go extension's schema.

deEshu2002:
We have to specifically ensure that GO111MODULE isn't set to "off", which could lead to looking for package at GOPATH/src or GOROOT/src. VSCode need to know that module-mode is enabled for the go command.


boojack (approved):
LGTM

PR #1787, opened 2023-06-01
chore: clean Dockerfile to exclude musl-dev

athurg:
We already disabled `CGO_ENABLED` in golang build, so the `musl-dev` which used to install `musl-libc` doesn't need any more. Just clean it from Dockerfile


boojack (approved):
LGTM

Issue #1698, opened 2023-05-21
Editing shortcut with display time filter resets date when editing


adamshand:
### Describe the bug

When editing a shortcut with a display time filter, the displayed date is set to today, regardless of what the actual date for the filter is.

### Steps to reproduce

1. Create a shortcut with a display time filter 
2. Edit the shortcut
3. In the edit widget, the display time will be shown as the time you edited the shortcut (instead of the time that it is actually set to).

boojack:
Shortcuts have been removed in the recent release.

PR #1793, opened 2023-06-04
filter display time

Duscansi:
Fix for the following issue:
Editing shortcut with display time filter resets date when editing #1698

I added an implementation on which we detect if we access the createShortcutDialog using the create new shortcut button or via editing existing shortcuts.
If we are editing an existing shortcut, the displaytime won't be altered with the current time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,26 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - uses: sonarsource/sonarqube-scan-action@master
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
+      # If you wish to fail your job when the Quality Gate is red, uncomment the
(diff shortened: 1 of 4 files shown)


web/src/components/CreateShortcutDialog.tsx
```diff
@@ -181,11 +187,18 @@ const MemoFilterInputer: React.FC<MemoFilterInputerProps> = (props: MemoFilterIn
 
   useEffect(() => {
     if (type === "DISPLAY_TIME") {
```
Zeng1998, line 189:
```suggestion
    if (type === "DISPLAY_TIME" && filter.value.value === "") {
```

I think we only need to add this condition to check whether it is a `creation` or an `editing` state. WDYT?

Zeng1998 (changes requested):

web/src/components/CreateShortcutDialog.tsx
Duscansi: Hi thanks for your comment. Yes I guess this will work too :) I will adjust it and try to create a new pull request

web/pnpm-lock.yaml
```diff
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
Zeng1998, line 1: And you don't need to modify this file.

Zeng1998 (changes requested):

Duscansi (commented):
I will adjust it

sonar-project.properties
athurg, line 1: I suggest you to skip this file because not all developers use this tool.

.github/workflows/build.yml
athurg, line 1: And also this file, should we really need this?

PR #1795, opened 2023-06-04
fix: user default 'Basic Setting' should follow server's setting

### LINKED ISSUE #1794: User's default 'Basic Setting' should follow server's setting
### Describe the bug

When a new user sign up, default locale & appearance will be set to 'en' & 'light' even server has another setting already.

I think this is a wrong behavior, default setting should follow server's setting.

### Steps to reproduce

1. sign out current user
2. clear all localstorage item
3. sign up a new user & login
4. user's default locale always be 'en' & appearance always be 'light'

_No response_

s0urcelab:
try to fix #1794


boojack (approved):

PR #1804, opened 2023-06-06
chore: modify the error message when registering a Host user

2hangpeng:
The minimum length of the password is 3, but the error message given to the front end is 6


2hangpeng:
Modified, please check.

> Please also modify this file.
> 
> https://github.com/usememos/memos/blob/main/setup/setup_test.go#L104

Modified, please check.

boojack (approved):

PR #1807, opened 2023-06-07
feat: public in days

### LINKED ISSUE #1122: 支持可见时间范围设置：3天、30天之类的。

应该很多人把memos当微博用，很多分享是即兴的，时间久了可能不合时宜，回去一个个删除或调整为仅自己可见，都比较费事。这方面微信确实把人性拿捏的很到位。

支持设置超过N天（对比创建时间）的Memos可见范围设置：

- 仅登录可见
- 仅自己可见

有个细节：针对置顶的不做处理。

### Additional context

_No response_

athurg:
Add support for filter public memos by days.

Any user can config a `public-in-days` setting, which will hide all public memos create before that. Both `/memos/all` in `Explore` and RSS feed will affect by this.

This PR also close #1122

### DIFF
--- a/server/memo.go
+++ b/server/memo.go
@@ -450,8 +450,8 @@ func (s *Server) registerMemoRoutes(g *echo.Group) {
 	g.GET("/memo/all", func(c echo.Context) error {
 		ctx := c.Request().Context()
 		findMemoMessage := &store.FindMemoMessage{}
-		_, ok := c.Get(getUserIDContextKey()).(int)
-		if !ok {
+		_, userLogined := c.Get(getUserIDContextKey()).(int)
+		if !userLogined {
 			findMemoMessage.VisibilityList = []store.Visibility{store.Public}
 		} else {
 			findMemoMessage.VisibilityList = []store.Visibility{store.Public, store.Protected}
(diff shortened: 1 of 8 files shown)


server/memo.go
```diff
@@ -702,3 +710,49 @@ func (s *Server) getMemoDisplayWithUpdatedTsSettingValue(ctx context.Context) (b
 	}
 	return memoDisplayWithUpdatedTs, nil
 }
+
+// FilterPublicMemosByDays filter public MemoMessages by creator's public-in-days setting.
+func FilterPublicMemosByDays(ctx context.Context, s *store.Store, list []*store.MemoMessage) ([]*store.MemoMessage, error) {
+	settings, err := s.FindUserSettingList(ctx, &api.UserSettingFind{Key: api.UserSettingPublicInDays})
```
boojack, line 716: `FindUserSetting` will be enough.

server/memo.go
```diff
@@ -702,3 +710,49 @@ func (s *Server) getMemoDisplayWithUpdatedTsSettingValue(ctx context.Context) (b
 	}
 	return memoDisplayWithUpdatedTs, nil
 }
+
+// FilterPublicMemosByDays filter public MemoMessages by creator's public-in-days setting.
+func FilterPublicMemosByDays(ctx context.Context, s *store.Store, list []*store.MemoMessage) ([]*store.MemoMessage, error) {
```
boojack, line 715: Do we need to call this function in other package? If not let's naming it in lower case `filterPublicMemosByDays`.

server/memo.go
athurg: We need fetch setting for each creator, not only the current user. So FindUserSetting is not satisfied.

athurg:
I have upgrade to filter memos in `store` layer with SQL. This will be better and solve some problems.

skyue:
will this feature continue? haha~

PR #1812, opened 2023-06-08
fix: systemSetting in UI changed unexpectedly

athurg:
In `Admin-System` section of `Setting` page, some settings state always be changed unexpectedly without any toggle on it. These settings include:

## `Allow user signup`
This setting was changed unexpectedly because itself. 

While it was toggled, it only call server API and update `state` store in page, the value in `globalStore` was not affected.

But if the `globalStore` was changed by any other settings, it will reset the `state` store in page to the value cached in `globalStore`.

## `Disable public memos` & `Display with updated time`
These two settings confused the local storage keys, so it changed each other while toggle.

### DIFF
--- a/web/src/components/Settings/SystemSection.tsx
+++ b/web/src/components/Settings/SystemSection.tsx
@@ -74,6 +74,7 @@ const SystemSection = () => {
       ...state,
       allowSignUp: value,
     });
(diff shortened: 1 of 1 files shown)


boojack (approved):
LGTM

PR #1802, opened 2023-06-05
feat: add japanese locale

kakik0u:
Added Japanese language and setting files.

PR #1822, opened 2023-06-12
Create hr.json

diomed:
Croatian Language


Zeng1998 (changes requested):
You also need to modify the other files.

Like this PR, https://github.com/usememos/memos/pull/1802

web/src/i18n.ts
```diff
@@ -7,7 +7,8 @@ export const availableLocales = [
   "de",
   "en",
   "es",
-  "fr",
```
Zeng1998, line 10: Unnecessary spaces.

Zeng1998 (changes requested):

web/src/i18n.ts
```diff
@@ -7,7 +7,8 @@ export const availableLocales = [
...
+  "fr",
```
boojack, line 10:
```suggestion
  "fr",
```

PR #1829, opened 2023-06-14
fix: use `??` in className use

sedationh:
问题表现如下
image
修改后
image


boojack (approved):

PR #1866, opened 2023-06-30
chore: listMemos sort by id for memos post/update at the same time

Vespa314:
For memo with equal sorting time (e.g., sorted by creation time, where memos created via an API may occur within the same second), the one with a higher ID should be placed in the front.


boojack (approved):
LGTM

Issue #1619, opened 2023-05-01
Reported some critical security vulnerabilities


mnqazi:
### Describe the bug

https://huntr.dev/bounties/4747a485-77c3-4bb5-aab0-21253ef303ca/

https://huntr.dev/bounties/e1107d79-1d63-4238-90b7-5cc150512654/

### Steps to reproduce

There is a link in description.

mnqazi:
@boojack Any Update?

mnqazi:
Why there is no response? I reported it a month ago. Just look at the severity of vulnerability. This is critical. @boojack please resolve it.

mnqazi:
@boojack It is possible to reproduce the scenario #1596 with my finding.

benjick:
@mnqazi why don't you send a PR?

mnqazi:
@athurg I actually submitted it on huntr.dev which is bug bounty platform for github repo. I submitted the bug there, so only maintainer can see the private reports.

mnqazi:
if you want, I can post the POC here

CorrectRoadH:
> https://huntr.dev/bounties/4747a485-77c3-4bb5-aab0-21253ef303ca/

👀 Yes, I can see the video now. I will try to reproduce it. Because I don't know how to use huntr.dev before, I didn't waste a little time. I am sorry for it.😭

CorrectRoadH:
I reproduce all vulnerabilities🤯. I am very thankful to you🙏. I need some time to fix it.

mnqazi:
Thanks

PR #1868, opened 2023-06-30
fix: ignore internalpath field when creating resource

CorrectRoadH:
fix: https://github.com/usememos/memos/issues/1619


boojack (approved):
LGTM

PR #1870, opened 2023-07-01
fix: pin memos of other people

CorrectRoadH:
fix https://huntr.dev/bounties/3ff3325a-1dcb-4da7-894d-81a9cf726d81/


CorrectRoadH:
May we can add a new middle to process all relation issues(verify the memos and user)

boojack (approved):

PR #1885, opened 2023-07-03
fix: auto expand height of setting page

athurg:
Since we have split the *Setting* into a page instead of dialog, its height could expand to the whole screen instead of a fixed value.


boojack (approved):
LGTM

PR #1886, opened 2023-07-03
fix: OpenID was disappear sometimes

athurg:
While making a request to */v1/status*, the *User* cache will be modified unexpectedly. After that, if you request */user/me* again, the response *User* doesn't have *OpenID* and *Email* field.


boojack (approved):
LGTM

PR #1888, opened 2023-07-03
fix: header archived id

Jiraiya8:
The ID of the header archived is the same as the ID of the header settings.


boojack (approved):

PR #1897, opened 2023-07-05
feat: Add support for purged resource link

athurg:
Currently, the resource link has two forms like below:
- `/o/r/:resourceId/:publicId`, which is marked as DEPRECATED.
- `/o/r/:resourceId/:publicId/:filename`

Both of them are too redundant. In fact, just a `resourceId` is enough to fetch a resource.

So I made this PR to add support for a new link form like `/o/r/:resourceId`.

For compatible with the old version, the old styles are still be supported, but should migrate the the new one.

And in the next version, I'll change the link in frontend code to this new form.


boojack (approved):
LGTM

PR #1924, opened 2023-07-11
feat: format message from telegram and upload attachments

aatumaykin:
Improvements to the bot's telegrams:
- links are added to the message and markdown format
- if the message is forwarded from a public channel, a link to the original message is added.
- if there is a pdf document in the message, then it is added to the note

before
Снимок экрана 2023-07-13 в 13 26 32

after
Снимок экрана 2023-07-13 в 13 24 03


boojack (approved):

athurg:
There're some suggestions, maybe you can consider about it:

1. You can defined a enum type for `MessageEntity.Type`.
1. Just use `switch...case` with a `default` branch in `ConvertToMarkdown` instead a lot of `if` conditions, and all of the `IsXXX()` method of `MessageEntity` could remove.
1. Do not remove the message type check in `Bot.Start` before you have support all message type of Telegram. Add the type you have full supported only.
1. Maybe the comment for struct's field should added as inline mode, and align the field type with same columns like other codes.
1. I think the Telegram's message format (which works with message entity) maybe not full compatible with Markdown, so I just skip that in the first version. Could you please check the other message entity type (other than Bold/Italic... you have considered) should work fine in Markdown with the original text.
1. Maybe *.gitignore* 's modifications should be remove before everyone think it should be.

Otherwise It's a good idea to extend the Telegram's message, thanks for your work!

.gitignore
athurg, line 1: Maybe you should keep this file untouched.

plugin/telegram/bot.go
athurg, line 1: Do not remove those code before you have add support for **ALL** Telegram message type. If you just add one/more new message type, just add them here to skip check.

plugin/telegram/chat.go
```diff
@@ -1,9 +1,36 @@
 package telegram
 
 type Chat struct {
-	ID        int    `json:"id"`
-	Title     string `json:"title"`
-	Type      string `json:"type"`
```
athurg, line 6: Just add a enum type `ChatType` and some const for exists value, like the code in other packages.

plugin/telegram/chat.go
athurg, line 1: Maybe a `ChatType` define with enum values is better than a lot of Bool functions.

plugin/telegram/document.go
athurg, line 1: Comment for struct's field should be inline, and align all the field's type in same column like the other code. For example:

```golang
type Document struct {
	FileID       string `json:"file_id"`        // FileID ...
	FileUniqueID string `json:"file_unique_id"` // FileUniqueID ...
}
```

plugin/telegram/message.go
athurg, line 1: Just change the code style for struct's field

plugin/telegram/message_entity.go
athurg, line 1: Add a `MessageEntityType` with all enum values instead of string type

server/telegram.go
```diff
@@ -86,6 +92,8 @@ func (t *telegramHandler) MessageHandle(ctx context.Context, bot *telegram.Bot,
 			mime = "image/jpeg"
 		case ".png":
 			mime = "image/png"
+		case ".pdf":
```
athurg, line 95: Could you please add more mime support, such as videos?

server/telegram.go
```diff
@@ -158,3 +166,49 @@ func generateKeyboardForMemoID(id int) [][]telegram.InlineKeyboardButton {
 
 	return [][]telegram.InlineKeyboardButton{buttons}
 }
+
+func ConvertToMarkdown(text string, messageEntities []telegram.MessageEntity) string {
```
athurg, line 170: If all Telegram message entity could convert or compatible with `Markdown`?

If this can be done, just add all entity type, else leave it unsupported is better than support with exceptions.

.gitignore
aatumaykin: Ok

server/telegram.go
aatumaykin: The entity specifies the type of formatting: emoji, spoiler, url, mention, classic bold, italic, etc.

I have added processing of those that your markdown supports. Not in markdown right now. for example, underline.

For specific entities there is no analogue in markdown and not all of them makes sense to try to format somehow. Therefore, they are ignored.

server/telegram.go
aatumaykin: Here you need to look at the situation. So far I have added mp4. In the other cases that I tested, mime is taken from the data in the message

aatumaykin:
The final formatting may turn out to be bad, due to the fact that incorrect coordinates of words are often indicated in the block for formatting in telegram messages.

Added support for the following types of messages:
- Audio
- Animation
- Video
- VideoNote
- Voice
- Document

oarsheo:
[Code Review AI](https://github.com/marketplace/code-review-ai): `plugin/telegram/blob.go`

1. The `GetMimeType` method could be simplified by using a map to store the mime types. This would make the code cleaner and easier to maintain.

2. The `GetMimeType` method could be improved by handling the case where the file extension is not recognized. Currently, it defaults to "application/octet-stream", which may not be appropriate for all file types.

Here's how you could implement these suggestions:

```go
var mimeTypes = map[string]string{
	".jpg": "image/jpeg",
	".png": "image/png",
	".mp4": "video/mp4",
}

func (b Blob) GetMimeType() string {
	if b.MimeType != "" {
		return b.MimeType
	}

	mime, ok := mimeTypes[path.Ext(b.FileName)]
	if !ok {
		// Handle unknown file extension
		// This could be logging an error, returning a default mime type, etc.
		return "application/octet-stream"
	}

	return mime
}
```

-------

[Code Review AI](https://github.com/marketplace/code-review-ai): `plugin/telegram/handle.go`

1. Code Duplication: There is a lot of repeated code for handling different types of messages (Photo, Animation, Audio, Document, Video, VideoNote, Voice). This can be refactored into a separate function to reduce code duplication. 

Refactored code:

```go
func handleMessageType(ctx context.Context, message *Message, messageType string, b *Bot) ([]Blob, error) {
    var fileID, fileName, mimeType string
    switch messageType {
    case "Photo":
        fileID = message.GetMaxPhotoFileID()
    case "Animation":
        fileID = message.Animation.FileID
        fileName = message.Animation.FileName
        mimeType = message.Animation.MimeType
    // Add similar cases for Audio, Document, Video, VideoNote, Voice
    }
    blob, err := b.downloadFileID(ctx, fileID)
    if err != nil {
        return nil, err
    }
    blob.FileName = fileName
    blob.MimeType = mimeType
    return append(blobs, *blob), nil
}
```
Then, in your main function, you can call this function for each message type:

```go
blobs, err := handleMessageType(ctx, &message, "Photo", b)
if err != nil {
    return err
}
```

--------

[Code Review AI](https://github.com/marketplace/code-review-ai): `plugin/telegram/message_entity.go`

1. The `MessageEntity` struct could benefit from adding comments to the fields. This will improve the readability and maintainability of the code.

```go
type MessageEntity struct {
	Type     MessageEntityType `json:"type"`   // Type of the entity.
	Offset   int               `json:"offset"` // Offset in UTF-16 code units to the start of the entity.
	Length   int               `json:"length"` // Length of the entity in UTF-16 code units.
	URL      string            `json:"url"`    // URL for “text_link” only, url that will be opened after user taps on the text.
	User     *User             `json:"user"`   // User for “text_mention” only, the mentioned user.
	Language string            `json:"language"` // Language for “pre” only, the programming language of the entity text.
}
```

2. Consider using `iota` for the `MessageEntityType` constants if they are not being serialized/deserialized to/from a database or external API. This can make your code more idiomatic and easier to maintain.

```go
type MessageEntityType int

const (
	Mention MessageEntityType = iota
	Hashtag
	CashTag
	BotCommand
	URL
	Email
	PhoneNumber
	Bold
	Italic
	Underline
	Strikethrough
	Code
	Pre
	TextLink
	TextMention
)
```

3. If `MessageEntityType` values are used in serialization/deserialization, consider using `string` type for `MessageEntityType` and define constants as untyped strings. This will make your code more idiomatic and easier to maintain.

```go
type MessageEntityType string

const (
	Mention       = "mention"
	Hashtag       = "hashtag"
	CashTag       = "cashtag"
	BotCommand    = "bot_command"
	URL           = "url"
	Email         = "email"
	PhoneNumber   = "phone_number"
	Bold          = "bold"
	Italic        = "italic"
	Underline     = "underline"
	Strikethrough = "strikethrough"
	Code          = "code"
	Pre           = "pre"
	TextLink      = "text_link"
	TextMention   = "text_mention"
)
```

aatumaykin:
I do not agree with the latest comments, they look like a suggestion/requirement to refactor the entire code of the telegram plugin.

The Switch in GetMimeType has been moved from the already existing code. Before PR, it didn't bother anyone that there were no checks and the default value of octet-stream 

Message Entity Type is a string in json that returns telegram. Therefore, int does not fit there in any way

aatumaykin:
refactoring

boojack (approved):
Thanks for your contribution again!

athurg (commented):
LGTM

PR #1932, opened 2023-07-12
chore: split Go binary and src for dev

athurg:
This is just a optimization for developers.

While we have already cached the *GOPATH* of container into the host before.

But the Golang's mod (which contains only source code) could be shared with the common *GOPATH* in the host, instead of a separated directory only this project, this will save a lot of disk spaces.

The binary in *GOPATH* should still cached in our project's path, because the container OS may different with the host.


boojack (approved):
LGTM

PR #1933, opened 2023-07-12
chore: update MemoRelationListView.tsx

sergeivassiljev:
Simplified the arrow function in the map function by removing the curly braces and the return keyword, since it only has one expression.


boojack:
Please fix the error from eslint check.

boojack (approved):

PR #1934, opened 2023-07-12
feat: implement memos chat backend function

CorrectRoadH:
## What does the PR do?
- add three API
/openai/chat-completion(old)
/openai/enabled(old)
/openai/chat-streaming
- revert the OpenAI config setting
- skip  chat-streaming API timeout


boojack (approved):
LGTM

PR #1950, opened 2023-07-14
feat: add support for auto backup db file

athurg:
This PR add a new feature to auto backup db file into a single file for backup.

The backup files will be named as the db file with a suffix, like `-YYYmmdd-hhmmss`.

Because we need more test, there is no setting UI provided in this version, this will be done in next version. User who want to try this can manually run a SQL in the db like below:

```sql
insert into system_setting(name,value) values('auto-backup-interval','3600')
```

The `3600` means the interval of auto backup by seconds.


VlaK0r:
Thank you for the great program!
I didn't understand how to do it, can I learn more?)

PR #1953, opened 2023-07-14
fix: incorrect localization key for sign-up failure message

Ajaykumbhare:
- Updated the localization key used for the sign-up failure message in the `handleSignUpButtonClick` function of `Auth.tsx`.
- Replaced `common.signup-failed` with the correct localization key `message.signup-failed` to ensure proper localization.
- This fix ensures that the correct localization key is used for the sign-up failure message, providing accurate translations and a better user experience.


boojack (approved):

Issue #1952, opened 2023-07-14
Incorrect localization key used for sign-up failure message


Ajaykumbhare:
### Describe the bug

When a sign-up attempt fails in the code, the localization key `common.signup-failed` is being used to display an error message. However, this key does not exist in the i18n file. The correct key to be used is `message.signup-failed`.

This bug should be fixed by updating the code to use the correct localization key for the sign-up failure message.

### Steps to reproduce

1. Open the `Auth.tsx` file in your codebase.
2. Locate the `handleSignUpButtonClick` function within the file.
3. In the `handleSignUpButtonClick` function, there is a line of code that reads: `toast.error(t("common.signup-failed"));`
4. Navigate to the `web/src/locales/en.json` file in your codebase.
5. Inside the `en.json` file, search for the `common` object.
6. Verify that the `common` object does not contain a key named `signup-failed`.

These steps illustrate the problem where the incorrect localization key `common.signup-failed` is being used in the code, and it does not exist in the `en.json` file under the `common` object.

### Screenshots or additional context

![ray-so-export](https://github.com/usememos/memos/assets/17526893/52e54e19-12d0-42ef-9f8f-1fd742e65897)

PR #1954, opened 2023-07-14
feat: add typeScript support to enforce valid translation keys

Ajaykumbhare:
To prevent issues like #1952, I have introduced TypeScript support to enforce the usage of valid translation keys.

- Added TypeScript support to enforce valid translation keys when using the `useTranslate` hook.
- Modified the type definitions of the `useTranslate` hook to restrict the allowed translation keys to those available in the `en.json` file.
- This enhancement ensures that developers are alerted to any invalid translation keys at compile-time, reducing the possibility of runtime errors related to missing or mistyped keys.
- By enforcing valid translation keys, we improve the maintainability and reliability of the localization system.


boojack (approved):
LGTM, great refactor!

PR #1957, opened 2023-07-14
fix: exclude commas in tags

pipe01:
Currently, if you try to write a tag and add a comma right after it, e.g. `I love using #Memos, it's great`, a tag named `#Memos,` will be created instead of `#Memos`, which is what I expected. This is solved by adding commas to the set of excluded characters in the tag regex in both the front end and back end, and this could be expanded by excluding all punctuation characters using `[[:punct:]]`.

### DIFF
--- a/api/v1/tag.go
+++ b/api/v1/tag.go
@@ -176,7 +176,7 @@ func convertTagFromStore(tag *store.Tag) *Tag {
 	}
 }
 
-var tagRegexp = regexp.MustCompile(`#([^\s#]+)`)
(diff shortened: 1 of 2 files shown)


boojack (approved):

pipe01:
On second thought this should definitely include all punctuation marks, because the same issue occurs with dots and exclamation points. Should I make a new PR?

boojack:
@pipe01 SGTM, maybe we can change it to `[A-Za-z/]`.

PR #1958, opened 2023-07-14
feat: update russian translate and new translate message

aatumaykin:
- add new localization keys: common.mark, setting.system-section.display-with-updated-time
- replaced text "MEMOS" in editor dialog to key amount-text.memo_one
- replaced some keys with semantically appropriate ones


boojack (approved):
LGTM

Issue #1916, opened 2023-07-08
Username URLs instead of account number


jrbgit:
### Is your feature request related to a problem?

It would be nice to have the url be /u/username instead of /u/1

### Describe the solution you'd like

Not sure if this is possible. my guess is it would need to be done on the 'server' and then possibly need some conf changes if it's running behind a reverse proxy like nginx

### Additional context

not a big issue - just a thought

aatumaykin:
fixed to https://github.com/usememos/memos/pull/1977

PR #1966, opened 2023-07-15
feat(security): disable access for anonymous users, when disablePublicMemos is true

aatumaykin:
- Added check system.disablePublicMemos for non authorized users.
If user is anonymous and disablePublicMemos is on, then redirect to auth page 
- hide link to explore on sidebar for non authorized users.

### DIFF
--- a/web/src/components/Header.tsx
+++ b/web/src/components/Header.tsx
@@ -1,7 +1,7 @@
 import classNames from "classnames";
 import { useEffect } from "react";
 import { NavLink, useLocation } from "react-router-dom";
-import { useLayoutStore, useUserStore } from "@/store/module";
+import { useLayoutStore, useUserStore, useGlobalStore } from "@/store/module";
 import { useTranslate } from "@/utils/i18n";
 import { resolution } from "@/utils/layout";
 import Icon from "./Icon";
@@ -16,6 +16,9 @@ const Header = () => {
   const layoutStore = useLayoutStore();
   const showHeader = layoutStore.state.showHeader;
   const isVisitorMode = userStore.isVisitorMode() && !userStore.state.user;
+  const {
+    state: { systemStatus },
+  } = useGlobalStore();
 
   useEffect(() => {
     const handleWindowResize = () => {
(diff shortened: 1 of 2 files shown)


boojack:
Two design points:

* The `disablePublicMemos` is mainly used for new public memos created by users. For those existing old public memos, they will still be displayed on the explore page.
* The explore page should always be visible, whether logged in or not, just like someone's blog.

So is there some misunderstanding here, or do you have some other thoughts?

aatumaykin:
> Two design points:
> 
> * The `disablePublicMemos` is mainly used for new public memos created by users. For those existing old public memos, they will still be displayed on the explore page.
> * The explore page should always be visible, whether logged in or not, just like someone's blog.
> 
> So is there some misunderstanding here, or do you have some other thoughts?

I think it should be possible to disable public access to all notes with one button. In my opinion, the disablePublicMemos setting can be used just for this. I make disablePublicMemos = true and all public memos become inaccessible to anonymous users. You need to be logged in to view them. At the same time, the visibility state in the notes themselves does not need to be changed.

There are only edits for the frontend, but in the future it will be necessary to finalize the backend. I also think that using id access to users and notes is a bad idea. You can sort through all the notes through brute force. From a security point of view, it is now considered good practice to use uuid identifiers.

boojack (approved):
LGTM

boojack:
> I think it should be possible to disable public access to all notes with one button. In my opinion, the disablePublicMemos setting can be used just for this. I make disablePublicMemos = true and all public memos become inaccessible to anonymous users. You need to be logged in to view them. At the same time, the visibility state in the notes themselves does not need to be changed.
> 
> There are only edits for the frontend, but in the future it will be necessary to finalize the backend. I also think that using id access to users and notes is a bad idea. You can sort through all the notes through brute force. From a security point of view, it is now considered good practice to use uuid identifiers.

For the uid problem, we can use `username` instead of `id` for users in the routers. And add a `uuid` field to the memo model to replace the `id` in the router. In the backend, we should still keep using the `id`. WDYT?

Related issue: https://github.com/usememos/memos/issues/1916

aatumaykin:
> > I think it should be possible to disable public access to all notes with one button. In my opinion, the disablePublicMemos setting can be used just for this. I make disablePublicMemos = true and all public memos become inaccessible to anonymous users. You need to be logged in to view them. At the same time, the visibility state in the notes themselves does not need to be changed.
> > There are only edits for the frontend, but in the future it will be necessary to finalize the backend. I also think that using id access to users and notes is a bad idea. You can sort through all the notes through brute force. From a security point of view, it is now considered good practice to use uuid identifiers.
> 
> For the uid problem, we can use `username` instead of `id` for users in the routers. And add a `uuid` field to the memo model to replace the `id` in the router. In the backend, we should still keep using the `id`. WDYT?
> 
> Related issue: #1916

Yes, I saw issue about username. I agree that username can be used. I haven't studied the code in detail, so I don't know if there is currently a uniqueness limit for username.

If the username is already unique, then there will be a minimum of problems with substitution in the url.

For memos and resources, you can generate a uuid based on id + salt (username or user id). This will greatly complicate brute force.

But all these changes are best done iteratively, with different commits/pull requests. Not in this PR. Many small PRs are better than one big one.

PR #1973, opened 2023-07-16
chore: disallow destructuring 't' from useTranslation

Ajaykumbhare:
This commit adds a linting rule to disallow the destructuring of the 't' property from the result of the useTranslation function call. The no-restricted-syntax rule in the ESLint configuration has been updated to enforce this restriction. The intention is to promote alternative approaches like using the useTranslate hook for localization.


boojack (approved):
LGTM

PR #1974, opened 2023-07-16
fix: exclude all punctuation chars except underscore in tags

pipe01:
The new regex uses the `u` modifier in the frontend regex which is only supported on es6 and above, but I think it's not a huge deal.

Related to #1957

### DIFF
--- a/api/v1/tag_test.go
+++ b/api/v1/tag_test.go
@@ -2,6 +2,8 @@ package v1
 
 import (
 	"testing"
+
+	"golang.org/x/exp/slices"
 )
 
 func TestFindTagListFromMemoContent(t *testing.T) {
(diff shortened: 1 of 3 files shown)


GadiZimerman:
@CodiumAI-Agent /review

api/v1/tag_test.go
```diff
@@ -35,7 +35,7 @@ func TestFindTagListFromMemoContent(t *testing.T) {
 		},
 		{
 			memoContent: "#tag1 http://123123.com?123123#tag2 \n#tag3  #tag4 http://123123.com?123123#tag2) ",
-			want:        []string{"tag1", "tag2", "tag2)", "tag3", "tag4"},
+			want:        []string{"tag1", "tag2", "tag3", "tag4"},
 		},
```
boojack, line 39: It'll be better to add more test cases.

boojack (approved):
LGTM

api/v1/tag_test.go
pipe01: Just added one more test and fixed the test logic not checking the values of the slices.

waterdrops: @pipe01 I will share more info about multi-level tags with you: I've just pulled the latest code and built an image using the main branch, however multi-level tags are still not rendering properly.
Would you please check it again?  Thank you very much.
#2009 #2024

PR #1977, opened 2023-07-17
feat: use username instead of uid

aatumaykin:
#1916

- replaced user Id with username in links
- I had to do a bit of refactoring on the frontend and backend

### DIFF
--- a/web/src/helpers/consts.ts
+++ b/web/src/helpers/consts.ts
@@ -1,6 +1,6 @@
 // UNKNOWN_ID is the symbol for unknown id
 export const UNKNOWN_ID = -1;
-export const UNKNOWN_USERNAME = "-";
+export const UNKNOWN_USERNAME = "";
 
 // default animation duration
 export const ANIMATION_DURATION = 200;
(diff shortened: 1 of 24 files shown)


web/src/helpers/consts.ts
```diff
@@ -1,5 +1,6 @@
 // UNKNOWN_ID is the symbol for unknown id
 export const UNKNOWN_ID = -1;
+export const UNKNOWN_USERNAME = "-";
```
boojack, line 3: How about using empty string instead of `-`?
```suggestion
export const UNKNOWN_USERNAME = "-";
```

boojack (approved):
LGTM, otherwise

web/src/helpers/consts.ts
aatumaykin: by analogy with UNKNOWN_ID, UNKNOWN_USERNAME is substituted into the query and returns an empty result, since the user "-" was not found. I did not alter the logic of the project much

boojack: Let's use the empty string, because `Boolean("") === false` but `Boolean("-") !== false`.

aatumaykin: ok, fixed to empty string

PR #1979, opened 2023-07-17
fix: copy code button toast message

Ajaykumbhare:
Refactored the code for the "Copy Code" button to utilize i18 strings for displaying the success message. Replaced the hard-coded value with the appropriate i18 string "Code copied successfully."


boojack (approved):
LGTM

Issue #1980, opened 2023-07-17
Disable selection of future dates in daily review section


Ajaykumbhare:
### Describe the bug

In our app's daily review section, there is a problem where users are able to select and view posts from future dates. This behavior is inconsistent with the app's functionality, as users cannot post in the future. To ensure a consistent and intuitive user experience, we need to disable the ability to select future dates in the daily review section.

### Steps to reproduce

1. Open the app and navigate to the daily review section.
2. Observe that the app displays posts from the current day as expected.
3. Click on next button, You will see always "Maybe no data was found, maybe it should be another option.".

### Screenshots or additional context

<img width="1020" alt="image" src="https://github.com/usememos/memos/assets/17526893/b915df84-4981-4746-9665-6a32320680ec">
<img width="775" alt="image" src="https://github.com/usememos/memos/assets/17526893/d2bbe426-ba37-4084-a51a-48b1341aab6a">

PR #1983, opened 2023-07-18
fix: disable selection of future dates in daily review section

Ajaykumbhare:
- Implemented a change to disable the ability to select future dates in the date picker.
- Added validation checks to ensure that the user cannot navigate to posts from future dates using the "next" button in the daily review section.
- This commit addresses GitHub issue #1980.


boojack (approved):
LGTM

PR #1984, opened 2023-07-18
Add rowStatus parameter for memo api document

EINDEX:
Found rowStatis was missing parameters in the API document. 

Just add it.


boojack (approved):
LGTM

PR #1987, opened 2023-07-18
fix: delete multiple resources, close #1986

### LINKED ISSUE #1986: Delete a batch of resource will throw a db lock error
### Describe the bug

image

### Steps to reproduce

1. Go to /resources page
2. Select two or more resources
3. Click trash button

_No response_

0x-jerry:
Fix #1986

### DIFF
--- a/web/src/pages/ResourcesDashboard.tsx
+++ b/web/src/pages/ResourcesDashboard.tsx
@@ -94,7 +94,6 @@ const ResourcesDashboard = () => {
         style: "warning",
         dialogName: "delete-resource-dialog",
         onConfirm: async () => {
-          // fix #1986
           for (const resourceId of selectedList) {
             await resourceStore.deleteResourceById(resourceId);
           }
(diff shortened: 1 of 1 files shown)


0x-jerry:
I looked at the backend code, maybe we can just exec delete sql statement instead of use a transaction.

I'm not sure if I'm right, because I rarely code with sql and golang.

https://github.com/usememos/memos/blob/8bcc2bd715778ef093bb640ab162b9eec4fe1809/store/resource.go#L169-L189

web/src/pages/ResourcesDashboard.tsx
```diff
@@ -94,9 +94,12 @@ const ResourcesDashboard = () => {
         style: "warning",
         dialogName: "delete-resource-dialog",
         onConfirm: async () => {
-          selectedList.map(async (resourceId: ResourceId) => {
+          // fix #1986
```
boojack, line 97:
```suggestion
```

boojack (approved):

boojack:
@0x-jerry For read-only operations, we can query directly through the database without a transaction. But for write operations, it is best to execute them in a transaction.

PR #1989, opened 2023-07-19
feat: add AutoBackupInterval in SystemSetting

athurg:
This PR add support for config `AutoBackupInterval` in system setting page, which was noticed on #1950 .


boojack (approved):
LGTM 👍

PR #1996, opened 2023-07-21
chore: update ja.json

kusanaginoturugi:
It is strange to translate "about" as "Notes について". However, if we define "について" as "about", the sentence becomes "について Memos". It is better not to translate here.


boojack (approved):

PR #1999, opened 2023-07-21
fix: add resource.clear to translate

aatumaykin:
fix src/pages/ResourcesDashboard.tsx(255,24): error TS2345: Argument of type '"resource.clear"' is not assignable to parameter of type 'Translations'.


boojack (approved):
LGTM

Issue #2016, opened 2023-07-23
Hashtag filter not working for certain tags


Ajaykumbhare:
### Describe the bug

The hashTag filter is not functioning correctly on the home page when clicking on specific hashtags in the posts.

### Steps to reproduce

1. Go to the home page.
2. Add a new post with the content: "I love #apple and #banana! #fruit_lover."
3. Click on the save button to post the content.
4. Observe the timeline where the posts are displayed.
5. Click on #apple and #banana hashTags

### Screenshots or additional context

<img width="1158" alt="image" src="https://github.com/usememos/memos/assets/17526893/6c6b5f01-46e8-47b0-87bc-77707b95ccd7">

<img width="1158" alt="image" src="https://github.com/usememos/memos/assets/17526893/c8f676a9-7886-4026-86a5-08de58feaf73">

<img width="1158" alt="image" src="https://github.com/usememos/memos/assets/17526893/eab66bca-6af6-4d38-bfe0-10e9c7caaa73">

<img width="1158" alt="image" src="https://github.com/usememos/memos/assets/17526893/72d3331d-619c-4397-b16d-f8d846277326">

gujin03:
same as https://github.com/usememos/memos/issues/2013

PR #2017, opened 2023-07-23
fix: hashtag filter for Unicode characters

Ajaykumbhare:
This commit addresses an issue #2016  with the hashtag filter, which was not working correctly for hashtags containing Unicode characters. The filter was only capturing single-letter hashtags instead of complete hashtags.


Ajaykumbhare:
@boojack I haven't touched the Go file, but the backend test is failing. Is there any way to debug the issue?

boojack (approved):
LGTM

PR #2018, opened 2023-07-23
fix: empty state display issue with resourceList style Grid

Ajaykumbhare:
This commit addresses a display issue with the empty state when the resourceList style is set to Grid. In the Grid style, the empty state was not being displayed correctly, leading to a layout inconsistency.

image


boojack (approved):
LGTM

Issue #660, opened 2022-12-02
Search on Explore page.


mariushosting:
### Is your feature request related to a problem?

It would be interesting to have the search bar also on the explore page. An option to Activate or Deactivate search on Explore page.

### Describe the solution you'd like

People or visitors should be able to search memos on front end page "Explore". We should have an option to Activate or Deactivate search on Explore page.

zhengcygg:
+1

PR #2025, opened 2023-07-24
feat: add search bar in archived and explore pages

### LINKED ISSUE #607: Allow searching archived items

- Currently once a memo is archived, I am unable to use search to find it.
- When we have more memos, from time to time it makes sense to archive it to keep it clean. However when we need to get back to one of those old ones we can't even if there is a tag attached to it. Only option is to manually scroll to find it. 

- It would be great to have an additional switch to allow searching in Archived memos also
- Also the search could be exclusive. ie. Either search in normal only OR in archived ones only

### Additional context

Requesting to review this feature requirement. Kindly correct me, in case I missed to find the option if a similar one already exists.

tranquanghuy0801:
### Description
Add a search bar in Archived Memos page
Resolves #607 and #660

### Archived Page

image

### Explore Page

image

### DIFF
--- a/web/src/pages/Archived.tsx
+++ b/web/src/pages/Archived.tsx
@@ -6,9 +6,9 @@ import useLoading from "@/hooks/useLoading";
 import ArchivedMemo from "@/components/ArchivedMemo";
 import MobileHeader from "@/components/MobileHeader";
 import Empty from "@/components/Empty";
-import "@/less/archived.less";
 import SearchBar from "@/components/SearchBar";
 import MemoFilter from "@/components/MemoFilter";
+import "@/less/archived.less";
 
 const Archived = () => {
   const t = useTranslate();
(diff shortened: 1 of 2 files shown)


web/src/pages/Archived.tsx
```diff
@@ -1,25 +1,31 @@
...
 import toast from "react-hot-toast";
-import { useMemoStore } from "@/store/module";
+import { useFilterStore, useMemoStore } from "@/store/module";
 import useLoading from "@/hooks/useLoading";
 import ArchivedMemo from "@/components/ArchivedMemo";
 import MobileHeader from "@/components/MobileHeader";
 import Empty from "@/components/Empty";
 import "@/less/archived.less";
+import SearchBar from "@/components/SearchBar";
+import MemoFilter from "@/components/MemoFilter";
```
boojack, line 11:
```suggestion
import Empty from "@/components/Empty";
import SearchBar from "@/components/SearchBar";
import MemoFilter from "@/components/MemoFilter";
import "@/less/archived.less";
```

boojack (approved):
LGTM

PR #2026, opened 2023-07-24
feat: add command to move blob from local to db

athurg:
With this PR, we can move resource blob from local file back to database, the only thing you need is run a sub command of *Memos*.

First of all , **STOP Your Container** and **Backup anything**.
Second of all, **STOP Your Container** and **Backup anything**.
AGAIN and AGAIN and AGAIN, **STOP Your Container** and **Backup anything**.

If you have finish all above, just run a cmd:
```
docker run -it -v ~/.memos/:/var/opt/memos ghcr.io/usememos/memos:latest mvrss
```

PS: Next version, we'll add support to move resource between more storages.

### DIFF
--- a/cmd/memos.go
+++ b/cmd/memos.go
@@ -114,6 +114,77 @@ var (
 			}
 		},
 	}
+
+	mvrssCmd = &cobra.Command{
+		Use:   "mvrss",
+		Short: "Move resource between storage",
+		Run: func(cmd *cobra.Command, _ []string) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+
+			from, err := cmd.Flags().GetString(mvrssCmdFlagFrom)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/memos.go
```diff
@@ -114,6 +114,77 @@ var (
 			}
 		},
 	}
+
+	mvrssCmd = &cobra.Command{
+		Use:   "mvrss",
```
boojack, line 119: Let's add a comment for the name, otherwise others may not understand the meaning and may be confused with RSS.

boojack (approved):
LGTM, otherwise

cmd/memos.go
athurg: I've added a comment here. By the way, in the `cobra.Command`'s `Short` field, there's already a description for `mvrss`.

```golang
	mvrssCmd = &cobra.Command{
		Use:   "mvrss", // `mvrss` is a shortened for 'means move resource'
		Short: "Move resource between storage",
		Run: func(cmd *cobra.Command, _ []string) {
```

PR #2032, opened 2023-07-26
chore: update ja.json

kusanaginoturugi:
I think free means freedom, not freemiam.

Fixed some strange expressions in the heatmap section.

Added spaces before and after English words that are in Japanese sentences.


boojack (approved):

PR #2039, opened 2023-07-28
feat: add system setting to disable password-based login

traumweh:
Currently, password-based login is still activated if one is using SSO login, allowing one to log into the same account using both password and SSO. Also, if one signs up via SSO, then a random (20 character) password will be generated for the new user, which could be used to log into the new account, circumventing SSO, even if it would require brute force because the password is randomized.

This PR aims at solving this possibly unwanted behaviour by adding a server setting to disable password-based login altogether.

In the future it might also make sense to remove the `NOT NULL` constraint on the password field in the user database table as well as adding a way for users to individually decide whether they want to use password-based login, SSO-based login or both. But that is something for another time.

### DIFF
--- a/api/v1/system_setting.go
+++ b/api/v1/system_setting.go
@@ -215,7 +215,6 @@ func (s *APIV1Service) registerSystemSettingRoutes(g *echo.Group) {
 			}
 
 			identityProviderList, err := s.Store.ListIdentityProviders(ctx, &store.FindIdentityProvider{})
-
 			if err != nil {
 				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to upsert system setting").SetInternal(err)
 			}
--- a/api/v1/system_setting.go
+++ b/api/v1/system_setting.go
@@ -208,6 +208,21 @@ func (s *APIV1Service) registerSystemSettingRoutes(g *echo.Group) {
 		if err := systemSettingUpsert.Validate(); err != nil {
 			return echo.NewHTTPError(http.StatusBadRequest, "invalid system setting").SetInternal(err)
 		}
+		if systemSettingUpsert.Name == SystemSettingDisablePasswordLoginName {
+			var disablePasswordLogin bool
+			if err := json.Unmarshal([]byte(systemSettingUpsert.Value), &disablePasswordLogin); err != nil {
+				return echo.NewHTTPError(http.StatusBadRequest, "invalid system setting").SetInternal(err)
+			}
+
+			identityProviderList, err := s.Store.ListIdentityProviders(ctx, &store.FindIdentityProvider{})
+
+			if err != nil {
+				return echo.NewHTTPError(http.StatusInternalServerError, "Failed to upsert system setting").SetInternal(err)
+			}
+			if disablePasswordLogin && len(identityProviderList) == 0 {
+				return echo.NewHTTPError(http.StatusForbidden, "Cannot disable passwords if no SSO identity provider is configured.")
+			}
+		}
 
 		systemSetting, err := s.Store.UpsertSystemSetting(ctx, &store.SystemSetting{
 			Name:        systemSettingUpsert.Name.String(),
--- a/api/v1/system_setting.go
+++ b/api/v1/system_setting.go
@@ -19,6 +19,8 @@ const (
 	SystemSettingSecretSessionName SystemSettingName = "secret-session"
 	// SystemSettingAllowSignUpName is the name of allow signup setting.
 	SystemSettingAllowSignUpName SystemSettingName = "allow-signup"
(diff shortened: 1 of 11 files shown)


boojack:
When SSO fails, or the admin cannot sign in via SSO, how can the admin sign in to the system if account password login is disabled?

boojack:
BTW, there is now the option to disable external user registration.

image

traumweh:
> When SSO fails, or the admin cannot sign in via SSO, how can the admin sign in to the system if account password login is disabled?

This is a typical problem with SSO-only authentication flows. I took reference by [Immich](https://github.com/immich-app/immich)'s implementation, in which case it simply isn't possible to login if the SSO service fails. In that case it would be necessary to manually disable the setting in the database and then the system would automatically revert to using the passwords.

One solution would be to add a script to the container image which automatically executes the sqlite statement to toggle the option or have an environment-variable for the container that could overrule the database option. But either way, true SSO-only login assumes a stable identity provider (and maybe a workaround to disable SSO).

In my opinion this setting is intended to be used only by admins who are aware of the requirement for stable SSO. It would e.g. enable admins to require their users to use two-factor authentication via SSO without still having the second Memos-internal login flow that only requires one-factor passwords.

I think if it is properly documented - i.e. warnings about the consequences and information about ways to disable it if one locked themselves out - then this issue is more just an inconvenience. Maybe a UI-warning upon activation would be useful too, tho I don't think I am familiar enough with react to do it properly.

Nevertheless, I am very interested in your view on this.

boojack:
I think we can first add a double confirm popup at the toggle button, with description to give a warning. Also, for the situation where login via username/password is completely blocked, the simplest solution is to delete that setting item in the database.

boojack:
We should check if there is SSO config when updating the settings in backend. And do not allow enabling this option if no SSO configurated.

https://github.com/usememos/memos/blob/main/api/v1/system_setting.go#L205

traumweh:
> I think we can first add a double confirm popup at the toggle button, with description to give a warning. Also, for the situation where login via username/password is completely blocked, the simplest solution is to delete that setting item in the database.

> We should check if there is SSO config when updating the settings in backend. And do not allow enabling this option if no SSO configurated.
> 
> https://github.com/usememos/memos/blob/main/api/v1/system_setting.go#L205

Those sound like good ideas. I'll try adding a double confirm popup and the check for configured SSO today or tomorrow.

traumweh:
I implemented the two-step confirmation (of which the second step consists of a typing-barrier) and prohibited the API call if no identity provider is configured. On top of that I also added an extra warning when one is about to remove an identity provider while password-based login is disabled.

web/src/components/DisablePasswordLoginDialog.tsx
```diff
@@ -0,0 +1,98 @@
+import { useState } from "react";
+import { toast } from "react-hot-toast";
+import { useTranslate } from "@/utils/i18n";
+import { useGlobalStore } from "@/store/module";
+import * as api from "@/helpers/api";
+import Icon from "./Icon";
+import { generateDialog } from "./Dialog";
+import { Button } from "@mui/joy";
```
boojack, line 8: Imports for third-party dependencies should be placed above the component. We can put it on the first line of the file.

boojack (approved):
LGTM, some nits

api/v1/system_setting.go
```diff
@@ -201,6 +208,21 @@ func (s *APIV1Service) registerSystemSettingRoutes(g *echo.Group) {
...
 			return echo.NewHTTPError(http.StatusBadRequest, "invalid system setting").SetInternal(err)
 		}
+		if systemSettingUpsert.Name == SystemSettingDisablePasswordLoginName {
+			var disablePasswordLogin bool
+			if err := json.Unmarshal([]byte(systemSettingUpsert.Value), &disablePasswordLogin); err != nil {
+				return echo.NewHTTPError(http.StatusBadRequest, "invalid system setting").SetInternal(err)
+			}
+
+			identityProviderList, err := s.Store.ListIdentityProviders(ctx, &store.FindIdentityProvider{})
+
```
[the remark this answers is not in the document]
traumweh: Did you mean like this? Or also the empty line above?

https://github.com/usememos/memos/blob/ae2856c5da2049139fd0652273e130f4ba636f97/api/v1/system_setting.go#L216-L218

web/src/components/DisablePasswordLoginDialog.tsx
traumweh: Wasn't aware about that convention. I hope this is correct now?

https://github.com/usememos/memos/blob/ae2856c5da2049139fd0652273e130f4ba636f97/web/src/components/DisablePasswordLoginDialog.tsx#L1-L8

boojack (approved):

PR #2041, opened 2023-07-28
fix: show full content in detail page

### LINKED ISSUE #1373: Show full content in the detail page

When open the detail page of a long memo, I'll see the full content first and then it's folded into a small card.

1. The flash of content is not a good UX.
2. It doesn't make sense to only show a small card since this is the detail page of the memo.

Just show the full content in the detail page.

### Additional context

image

gera2ld:
fix #1373 again


boojack (approved):

PR #2042, opened 2023-07-28
fix: avoid content flash on auto collapse

gera2ld:
**Background:** Currently if auto-collapse is on, long posts will be fully shown first and then collapsed, leading to significant content flash.

This PR will render all posts with a max height to avoid the content flash, then check `scrollHeight` instead of `clientHeight`.


boojack (approved):
LGTM

Issue #2070, opened 2023-08-03
feat(telegram): Support Telegram Channel


qazxcdswe123:
### Is your feature request related to a problem?

Read message from channel as a way to support multiple telegram user

closes: https://github.com/usememos/memos/issues/1906

### Describe the solution you'd like

I'm interested in working on adding this feature, but I want to make sure it aligns with the project's goals before implementing it.

Once I receive approval from the maintainer, I will begin working on implementing this feature.

Details

- Backend: Check if message is `channel_post`. Should only continue when the channel name match
    - If so, send api request to memos (and I'm not sure if button to change visibility is going to work on channel or not)
    - If not, reply to user as is

- Frontend:
  - Add channel name option

boojack:
Please go ahead. If you have any questions about the implementation, you can ask some helps from @athurg.

qazxcdswe123:
Update:

I managed to support channel post, however, there is not such `userid` in channel, the only identifier will be the channel name, which leads to another kv pair (key: `telegram-channel-name`) in table `user_setting`

boojack:
> the only identifier will be the channel name

It might be better to put it into system setting instead of user setting.

qazxcdswe123:
> > the only identifier will be the channel name
> 
> It might be better to put it into system setting instead of user setting.

Hi

Currently the `telegram-user-id` is stored in `user_setting`, so I think putting them in the same table would be more aligned with the current implementation?

PR #2078, opened 2023-08-04
WIP: feat(telegram): Support Telegram Channel

qazxcdswe123:
> Update:
> 
> I managed to support channel post, however, there is not such userid in channel, the only identifier will be the channel name, which leads to another kv pair (key: telegram-channel-name) in table user_setting
> 
> And TBH, I don't really know how to do that, I guess I'll make a draft PR first 🤣

ref: https://github.com/usememos/memos/issues/2070

I'm not sure if adding a user setting is a good move, let me know if there's a better way to do it

### DIFF
--- a/plugin/telegram/bot.go
+++ b/plugin/telegram/bot.go
@@ -29,13 +29,13 @@ func NewBotWithHandler(h Handler) *Bot {
 const noTokenWait = 30 * time.Second
 const errRetryWait = 10 * time.Second
 
-// Start start an infinity call of getUpdates from Telegram, call r.MessageHandle while get new message updates.
+// Start start a long polling using getUpdates to get Update, call r.MessageHandle while get new message updates.
 func (b *Bot) Start(ctx context.Context) {
 	var offset int
 
 	for {
 		updates, err := b.GetUpdates(ctx, offset)
-		if err == ErrInvalidToken {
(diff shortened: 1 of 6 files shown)


plugin/telegram/bot.go
```diff
@@ -61,7 +61,8 @@ func (b *Bot) Start(ctx context.Context) {
 				continue
 			}
 
-			// handle Message update
+			log.Info("receive update", zap.Any("update", update))
```
athurg, line 64: This log maybe remove or change to `Debug` instead of `Info`

server/telegram.go
```diff
@@ -37,6 +37,14 @@ func (t *telegramHandler) MessageHandle(ctx context.Context, bot *telegram.Bot,
 		return fmt.Errorf("fail to SendReplyMessage: %s", err)
 	}
 
+	// TODO: handle channel
+	// Message from channel
+	//if message.SenderChat != nil {
```
athurg, line 42: Do not leave any commented code, you can just uncomment them or remove those lines.

athurg (changes requested):

qazxcdswe123:
Forgot to mention that this is just a prototype and needs *a lot of* cleanup (my bad). And I got distracted a little bit then completely forgot this thing!

I guess I shall close it for now, until it is finished? 🥲

PR #2096, opened 2023-08-06
fix: minor spelling and wording changes for en and de 

dareiff:
I went through some of english and german localizations to correct or add minor things.

But also added `invalid-tag-name` to `en.json`. Which also means it should be translated elsewhere.


boojack (approved):

PR #2102, opened 2023-08-07
feat: use user avatar in ShareMemoDialog

akring:
Use user avatar in ShareMemoDialog instead of default logo.

CleanShot 2023-08-07 at 14 57 27@2x


boojack (approved):

PR #2107, opened 2023-08-07
updated Dutch translation

jxpsert:
Updated some weird translations I did before, and added new ones for new keys found in the English file.

I ignored some keys that don't seem to be used in the UI.


boojack (approved):
LGTM

PR #2116, opened 2023-08-09
feat(telegram): Support Telegram Channel

qazxcdswe123:
Hi

This PR adds support for handling telegram channel message, I also formatted some code while developing.

To use it, simply add bot to your channel and add `telegram-channel-name` in `user_setting` table and it's done.

However, currently the bot will also reply a message under the channel name (as pic below), which may be annoying, can obviously can not be used in public channel. 
At first I thought it could be fixed by disabling the bot's `post_message` privilege in channel, but this cause an error since the bot first sends a "request pending message" then edit the message, if no message was sent in the first place, it won't be able to edit the message which cause errors.
This can be fixed by deleting the bot's message after the message was sent, or completely refactor the send message logic.

image

Also, the channel name is not a unique identifier (you can have multiple channel with the same name), but the channel id can not be directly obtained from telegram app, it must go through the api. This is not intuitive so I tend to use the channel name as identifier :D

### DIFF
--- a/api/v1/system_setting.go
+++ b/api/v1/system_setting.go
@@ -36,7 +36,7 @@ const (
 	SystemSettingStorageServiceIDName SystemSettingName = "storage-service-id"
 	// SystemSettingLocalStoragePathName is the name of local storage path.
 	SystemSettingLocalStoragePathName SystemSettingName = "local-storage-path"
-	// SystemSettingTelegramBotToken is the name of Telegram Bot Token.
+	// SystemSettingTelegramBotTokenName is the name of Telegram Bot Token.
 	SystemSettingTelegramBotTokenName SystemSettingName = "telegram-bot-token"
 	// SystemSettingMemoDisplayWithUpdatedTsName is the name of memo display with updated ts.
 	SystemSettingMemoDisplayWithUpdatedTsName SystemSettingName = "memo-display-with-updated-ts"
--- a/api/v1/user_setting.go
+++ b/api/v1/user_setting.go
@@ -20,8 +20,10 @@ const (
 	UserSettingAppearanceKey UserSettingKey = "appearance"
 	// UserSettingMemoVisibilityKey is the key type for user preference memo default visibility.
 	UserSettingMemoVisibilityKey UserSettingKey = "memo-visibility"
-	// UserSettingTelegramUserID is the key type for telegram UserID of memos user.
+	// UserSettingTelegramUserIDKey is the key type for telegram UserID of memos user.
 	UserSettingTelegramUserIDKey UserSettingKey = "telegram-user-id"
+	// UserSettingTelegramChannelNameKey is the key type for telegram channel name user sent memo from.
+	UserSettingTelegramChannelNameKey UserSettingKey = "telegram-channel-name"
 )
 
 // String returns the string format of UserSettingKey type.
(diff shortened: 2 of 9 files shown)


qazxcdswe123:
cc @athurg

qazxcdswe123:
fiexed CI

athurg:
`Channel` is used for publishing message in Telegram, and only the admins (almost just one person or a bot) send messages, other subscriber just receive without sending anything. So in Telegram's API, there's no `sender` in channel update, which could used for identified the sender.

So I think it's may not satisfied for sending Memos in Channel, maybe `Group` is better than that.

qazxcdswe123:
Hi

I 'm afraid that I don't get your idea, what't the point of getting `sender`?

Also, IMHO supporting `Group` is weird since I don't think memos is a place for logging group chat message? 

Initially I thought this could serve as a "sync" between channel and memos, even though it's a bit off my original intention while developing (mostly the bot reply under channel's name part), I think it can still help some channel operator who happens to be a memos user? Support for multi telegram user is more like a side effect.

athurg:
Telegram bot is designed as a convenient way to send/save personal memos.

If you‘d like to sync the history of a group or channel to Memos, maybe there's a better way to do this, the exists bot maybe not the one.

For example, the bot require assign the Telegram user and Memos user first. But for a group/channel message sync, maybe it'll be better to auto create a Memo user from Telegram which is not exists.

And as you say, the bot should not reply a confirm message while receive a message from group/channel, and should set the message's visibility to `Protected` as default instead of `Private`.

qazxcdswe123:
close: https://github.com/usememos/memos/issues/1906 and 
related: https://github.com/qazxcdswe123/telegramMemoBot/issues/16

> Telegram bot is designed as a convenient way to send/save personal memos.

Yes I totally agree, but some channels are also run by one man so that count as "personal memos that I would also like to share with others"?

> If you‘d like to sync the history of a group or channel to Memos, maybe there's a better way to do this, the exists bot maybe not the one.

Probably not history but new messages from channels, beside that I agree with this part.

>  For example, the bot require assign the Telegram user and Memos user first. But for a group/channel message sync, maybe it'll be better to auto create a Memo user from Telegram which is not exists.

They can probably share a same memos account? Given that they are using the same memos instance, I don't think a new user for everyone in group/channel is really that necessary. Also, maybe one have multiple telegram account?

> And as you say, the bot should not reply a confirm message while receive a message from group/channel, and should set the message's visibility to Protected as default instead of Private.

Yeah I agree with you, this is a bit off my original thought...

athurg:
If you still think it's better to support channel&group message, and share Memos user with more than one Telegram user, I think there is a walk around here:

- If message is from a private chat ( means send to the bot directly), save it to Memos user  bind with the sender's id, which you can got from `result.message.from.id` or `result.message.chat.id` from Telegram's JSON request.
- If message is from a group chat (mentions and/or not?), save it to Memos user bind to the group's id, which you can got from `result.message.sender_chat.id` or `result.message.chat.id` from Telegram's JSON request.
- If message is from a channel, save it to Memos user bind to the channel's id, which you can got from `result.channel_post.sender_chat.id` or `result.channel_post.chat.id` from Telegram's JSON request.

For better compatible, maybe `chat.id` and `chat.type` should always be used together.

By the way, the Telegram user id looks like always be positive, with the channel/group id be negative, but I didn't found any document to describe it.

qazxcdswe123:
Thank you for your suggestion. You are right, this is a less-than-ideal solution :D. The complete solution is too complicated for me so I decided to close this PR :)

PR #2120, opened 2023-08-10
Update - add header for get resource

NicoleRobin:
add header for resource response to specify resource name to fix this issue: #2098


boojack (approved):
LGTM

PR #2124, opened 2023-08-11
chore: minor cleanup

qazxcdswe123:
Some minor cleanup found when implementing an unsuccessful feature, may be unnecessary but good to have


boojack (approved):
LGTM, good catches!

PR #2154, opened 2023-08-19
Update Dockerfile

Josecep:


boojack:
FYI https://github.com/usememos/memos/pull/1718. Any reason to remove this line? @Josecep

PR #2165, opened 2023-08-24
feat: add configuration option to bind server to specific address

iSLC:


boojack (approved):

Could you add some description for why and when to use this flag?

iSLC:
By default the application would listen on `0.0.0.0:port`. Which on the VPS that I'm using would make the application publicly available via the assigned server `ip:port`. I wanted to make the application listen to `127.0.0.1` (or `localhost`) and only be accessible via a reverse-proxy with nginx (_preferably with SSL_) and unable to access it directly through the server `ip:port`.

I also wanted to name the flag `host` first, but the shorthand `h` would probably probably conflict with a `help` flag, if any. In case anyone wonders why it's named `addr` instead (_tried to avoid inconsistent naming_).

PR #2214, opened 2023-09-13
fix: multiple inline latex parsing

kadaliao:
Fix parsing and rendering of multiple inline LaTeX code within a single line.

pre fix:
CleanShot 2023-09-13 at 17 07 17@2x

fixed:
CleanShot 2023-09-13 at 17 02 26@2x


boojack (approved):
LGTM

PR #2215, opened 2023-09-13
feat: notify by telegram while new memo create by HTTP

athurg:
This PR add a new feature, which'll send a notification by Telegram bot to users who bind a Telegram UserID. By this feature, users could fetch the updates in the Memos server without manually refresh the web page.


boojack (approved):
LGTM

PR #2218, opened 2023-09-14
chore: optimize performance of /memo/stats

athurg:
This PR will improvement the performance of `/api/v1/memo/stats`. In my case about 500+ memos, this'll decrease the latency of this API from 300ms to 3ms.


boojack (approved):
LGMT

PR #2220, opened 2023-09-15
chore: better date picker

athurg:
This PR contains two small optimization in DatePicker:
- Add two buttons to increase/decrease year.
- Add padding 0 on the month to keep the buttons position.


boojack (approved):
LGTM

PR #2222, opened 2023-09-15
fix: vite proxy setting

athurg:
This PR fix some settings for vite proxy, so we can keep the origin HTTP Host header and add some XFF headers to the backend API server.


boojack (approved):
LGTM

PR #2241, opened 2023-09-18
fix: content sanitization in getimage endpoint

victorsch:
Used the bluemonday package to sanitize content in images download from remote servers to prevent XSS


boojack (approved):

PR #2282, opened 2023-09-26
chore: move sql code of UserSetting into Driver

athurg:


athurg:
Maybe it'll conflict with #2281 because both of them add code in the same line of `Driver`, I'll fixed it if it happened.

PR #2339, opened 2023-10-06
chore: compress image and reduce 500kb in network

bestony:
feature:

- Compress web/public/logo.png from more than 800kb to 255kb.

Why do this?

It takes me over 30 seconds to load this file. So I think I need to compress it.

image


boojack (approved):
LGTM, good catch!

sulisu:
Why not use webp format as 0.14.x did? Browser that can not support webp will not well handle the moderm script technology used in Memos too, I think.

Sorry I am not familiar with github develop process, so I can not provide some PR, Just a suggestion.

Issue #2384, opened 2023-10-15
Storage setting changed don't take effect


alex-guoba:
### Describe the bug

In "Setting" -> "Storage" page,  I'm trying to change the current object storage from "Local file system" to "Database" , it seems not take effect on the website.   But if I refresh the pages, then setting took effect.

### Steps to reproduce

1. Go to  "Setting" -> "Storage"
2. click on "Database"
3. the focal point didn't changed.

### Screenshots or additional context

**The API return failed, this might cauth the page exception.**

/memos.api.v2.SystemService/GetSystemInfo INTERNAL: failed to get file info: stat root:helloworld@tcp(127.0.0.1)/memos: no such file or directory
    at makeCall (index-5a48164e.js:300:6112)
    at async at (index-5a48164e.js:300:11332)
    at async Object.getSystemInfo (index-5a48164e.js:300:11843)
    at async Object.fetchSystemStatus (index-5a48164e.js:320:20777)

PR #2385, opened 2023-10-15
fix: Storage setting changed don't take effect

alex-guoba:
bug fix for:
https://github.com/usememos/memos/issues/2384

### DIFF
--- a/store/db/mysql/mysql.go
+++ b/store/db/mysql/mysql.go
@@ -67,6 +67,11 @@ func (*DB) BackupTo(context.Context, string) error {
 	return errors.New("Please use mysqldump to backup")
 }
 
+func (d *DB) CurrentSize(ctx context.Context) (int64, error) {
+	// return 0, errors.New("Not implemented")
+	return 0, nil
+}
+
 func (d *DB) Close() error {
 	return d.db.Close()
 }
--- a/store/driver.go
+++ b/store/driver.go
@@ -17,6 +17,9 @@ type Driver interface {
 	Vacuum(ctx context.Context) error
 	BackupTo(ctx context.Context, filename string) error
 
+	// current file is driver
+	CurrentSize(ctx context.Context) (int64, error)
+
 	// Activity model related methods.
 	CreateActivity(ctx context.Context, create *Activity) (*Activity, error)
 	ListActivity(ctx context.Context, find *FindActivity) ([]*Activity, error)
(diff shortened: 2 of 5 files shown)


store/driver.go
```diff
@@ -17,6 +17,9 @@ type Driver interface {
 	Vacuum(ctx context.Context) error
 	BackupTo(ctx context.Context, filename string) error
 
+	// current file is driver
+	CurrentSize(ctx context.Context) (int64, error)
```
boojack, line 21: How about `GetCurrentDBSize`?

boojack (commented):
LGTM. But this PR maybe is not related with fixing storage setting changed issue?

store/driver.go
alex-guoba: Good idea.

store/db/mysql/mysql.go
```diff
@@ -67,6 +67,11 @@ func (*DB) BackupTo(context.Context, string) error {
 	return errors.New("Please use mysqldump to backup")
 }
 
+func (*DB) GetCurrentDBSize(context.Context) (int64, error) {
+	// return 0, errors.New("Not implemented")
```
boojack, line 71:
```suggestion
	// TODO: implement this.
```

alex-guoba:
> LGTM. But this PR maybe is not related with fixing storage setting changed issue?

actually it is.

1. /memos.api.v2.SystemService/GetSystemInfo will return http error code to client. Because `os.stat()` can't be used in database DSN 
2. page will retrive system status again and update local storage after API call finished. 
```
  const handleActiveStorageServiceChanged = async (storageId: StorageId) => {
    await api.upsertSystemSetting({
      name: "storage-service-id",
      value: JSON.stringify(storageId),
    });
    await globalStore.fetchSystemStatus();
    setStorageServiceId(storageId);
  };
```

if the API call failed. the local storage wouldn't update unless  refreshing the whole page.

store/db/mysql/mysql.go
alex-guoba: Done.
I have a question about this feature. If I modified my setting to Local file. but the memos message was still in Database. In this scenario, what's the "Database file size" means?  database size or object storage size?

boojack: It's the size of database file. Not the image resources in local file system.

boojack (approved):

PR #2405, opened 2023-10-19
chore: update Shortcuts link

BaksiLi:
Hello! I have done a few minor changes to the shortcut:
- The description in the README. `A shortcut for iOS` -> `Shortcuts (iOS, iPadOS or macOS)`, more consistent with others.
- Shortcut
  - A more specific **error message** (like `Error: No API Endpoint`)
  - If failed to save, prompt for copying the memo **text to the clipboard**, so that the user can save it manually (rather than losing it).
  - Prompt for confirmation after tags being appended. User can either confirm or **modify the text**. This helps when quick saving quotes from an article using share sheet and want to make comments.


BaksiLi:
I made some modifications (iCloud Shortcuts sharing is painful).

Thanks!

quanru:
Great job,could you please provide a version that no need confirm when click quick memo(not share scene in other app).

PR #2410, opened 2023-10-20
fix: fail to update user's update_ts

athurg:
This PR fix a bug, which will cause user's update on `update_ts` failed when using *MySQL* as storage driver.

### DIFF
--- a/web/src/pages/Setting.tsx
+++ b/web/src/pages/Setting.tsx
@@ -43,7 +43,7 @@ const Setting = () => {
   };
 
   return (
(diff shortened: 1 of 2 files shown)


web/src/pages/Setting.tsx
```diff
@@ -43,7 +43,7 @@ const Setting = () => {
   };
 
   return (
-    <section className="@container w-full max-w-3xl min-h-full flex flex-col justify-start items-start px-4 sm:px-2 sm:pt-4 pb-8 bg-zinc-100 dark:bg-zinc-800">
+    <section className="@container w-full max-w-5xl min-h-full flex flex-col justify-start items-start px-4 sm:px-2 sm:pt-4 pb-8 bg-zinc-100 dark:bg-zinc-800">
```
boojack, line 46: Not related with this PR

boojack (changes requested):

Issue #2417, opened 2023-10-21
logo file is so big, even the biggest logo :LOGO图片实在太大，几乎高达1M


csthenry:
### Is your feature request related to a problem?

logo file is so big, even the biggest logo, it should be  zipped within 100KB.
This goes against the keyword of lightweight!

### Describe the solution you'd like

it should be  zipped within 100KB.

moth-boi:
Yea, I'm surprised that it's not even SVG.

PR #2418, opened 2023-10-21
fix: smaller logo file

moth-boi:
A small fix for #2417 
In my opinion it would be better if the logo was SVG since it's just logo and there is no need for detail.
But I shrunk it to 56KB from 250KB


lincolnthalles:
SVG lose a lot of details, but on the other hand it's only 8 KB.

logo_trans
logo_black

Issue #2397, opened 2023-10-17
Fuzzy matching for tag autocomplete


gitkeniwo:
### Is your feature request related to a problem?

Rn, there is a tag autocomplete feature implemented, but it's matching from the beginning of a multi-level tag. 

Could we make it support fuzzy matching like **vscode**, or matching from the middle of a tag?

Say I have a tag `#language/nederlands` for Dutch memos, at this current version of memos, every time I need to input that tag, i will have to first type in `#lan` and get `#language` selected. And because there are more than 5 tags alphabetically before `nederlands`, i need to select `#language` first, typing in`/n`, and then the desired tag starts to appear in the prompt, which could be quite inefficient on mobile.
<img width="283" alt="image" src="https://github.com/usememos/memos/assets/34304497/2ef06390-78b5-4cab-89b0-29d0e10f84da">

For a vscode style matching, all I need to type in is something like `#lggnds` and it should find `#language/nederlands` correctly. For matching from the middle, you could do something like `#neder`, so don't need to first select `#language` either.

### Describe the solution you'd like

 Fuzzy matching

PR #2421, opened 2023-10-21
feat: Better tag suggestion

moth-boi:
Better tag suggestion as mentioned in issue #2397
At first I tried to implement Levenshtein Distance Fuzzy Sort but I found this method more reliable.


boojack (approved):

PR #2432, opened 2023-10-23
feat: support filter creator in /api/v2/memos

athurg:
This PR add a new feature, which can filter creator (by ID) in memos API v2.


proto/api/v2/memo_service.proto
```diff
@@ -78,6 +78,8 @@ message ListMemosRequest {
 
   // Filter is used to filter memos returned in the list.
   string filter = 3;
+
+  int32 creator_id = 4;
```
boojack, line 82:
```suggestion
  optional int32 creator_id = 4;
```

Issue #2433, opened 2023-10-24
Document or Allow opt-out of metrics collections


KevinSJ:
### Is your feature request related to a problem?

First and foremost, I have to say thank you for creating this product. As a user, I really appreicate the time and effort that have gone into its development. While my feature request may come across as a bit candid, I sincerely hope that my suggestions can be taken into consideration.

I noticed there's been a significant change of behavior for this application since upgrading to v0.16.1: it began collecting and sending metrics data to posthog, This came to my attention primarily because my ad-blocker started actively intercepting and blocking outbound traffic to the Posthog domain.

Here's an image illustrating the issue:
![image](https://github.com/usememos/memos/assets/7546654/ef88345c-36a0-4654-b154-ad285e14063f)

And further search in the code base proved that this was added as part of v0.16.1

https://github.com/usememos/memos/blob/79bbe4b82aeb0fe0f41e8410fc7fa1eff6b1aae9/server/service/metric/metric.go

### Describe the solution you'd like

As a dev myself, I understand the importance of collecting metrics in a product. As a user of this **privacy-first** application, I'd like to request:

- **Documentation in README**: It would be highly beneficial to clearly outlines what metrics are being collected and how they are utilized. This will provide a decent level of transparency and help the users understand how their data are being collected.
- **Opt-Out Mechanism or Explict Notification**: Provide a way to opt-out of the metrics collection and/or explictly stating metrics are being collected.

### Additional context

I'd like to emphasize that I'm offering this suggestion from a user perspective, and with no desire to incite a debate on the whether the product should collect any metrics. 

Ultimately, the team behind the product should have the final say on their product's policies and practices.

That said, I'm more than happy to implment either of the suggestions I made.

boojack:
See more on our [privacy policy](https://www.usememos.com/legal/privacy-policy).

boojack:
> Opt-Out Mechanism or Explict Notification: Provide a way to opt-out of the metrics collection and/or explictly stating metrics are being collected.

SGTM, please feel free to make a PR for this. @KevinSJ

athurg:
I fully agree with you, @KevinSJ . I think we'd better do not collection anything from user, because it is a Self-Host product, who's user maybe very care about it's data.

As a compromise, I think we should add a param to disable this at least.

KevinSJ:
@athurg @boojack Thank you both for your response and the work that's been done to resolve this issue. Really appreciate it.

dwong33:
[...]

PR #2445, opened 2023-10-26
chore: make metric collect configurable

athurg:
This PR add a switch to disable the metric collector start from `v0.16.1`, which is mentioned on #2433 .


athurg:
This PR will replace by #2447

PR #2457, opened 2023-10-27
feat: week from monday in heatmap for zh-Hans and ko

athurg:
This PR add a new feature, that makes Monday as the first day of week, while `locale` is `zh-Hans` or `ko`.


diomed:
shouldnt monday be first day of the week universally?

athurg:
> shouldnt monday be first day of the week universally?

Of course it could be, what do you think about @boojack ? Should we add a new user/system setting for this?

PR #2480, opened 2023-11-03
fix: private memos being public

### LINKED ISSUE #2462: Disable public memos - not working
### Describe the bug

The flag "Disable public memos" does not block public notes from being created or viewed publicly.

### Steps to reproduce

1. enable "Disable public memos"
2. create a public note
3. browse to the public note from unauth session

_No response_

itschip:
- fix(web/memo): filter out public option
- feat(api/memo): sanity check for disabled public memos

Resolves #2462

Screenshot 2023-11-03 at 13 56 16
Screenshot 2023-11-03 at 13 58 00

### DIFF
--- a/api/v1/memo.go
+++ b/api/v1/memo.go
@@ -313,6 +313,11 @@ func (s *APIV1Service) CreateMemo(c echo.Context) error {
 				createMemoRequest.Visibility = Private
 			}
 		}
+
+		// sanity check for visibility
+		if createMemoRequest.Visibility == Public && disablePublicMemos {
+			return echo.NewHTTPError(http.StatusBadRequest, "Public memos are disabled").SetInternal(errors.New("public memos are disabled"))
+		}
 	}
 
 	createMemoRequest.CreatorID = userID
--- a/api/v1/memo.go
+++ b/api/v1/memo.go
@@ -313,11 +313,6 @@ func (s *APIV1Service) CreateMemo(c echo.Context) error {
 				createMemoRequest.Visibility = Private
 			}
 		}
-
-		// sanity check for visibility
-		if createMemoRequest.Visibility == Public && disablePublicMemos {
-			return echo.NewHTTPError(http.StatusBadRequest, "Public memos are disabled").SetInternal(errors.New("public memos are disabled"))
-		}
 	}
 
 	createMemoRequest.CreatorID = userID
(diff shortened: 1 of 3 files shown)


itschip:
We could also disable the select option. What do you like the best @boojack

api/v1/memo.go
```diff
@@ -313,6 +313,11 @@ func (s *APIV1Service) CreateMemo(c echo.Context) error {
 				createMemoRequest.Visibility = Private
 			}
 		}
+
+		// sanity check for visibility
+		if createMemoRequest.Visibility == Public && disablePublicMemos {
```
boojack, line 318: Right now we allow admin to create public contents. And I think it's works good for me. 

So this check might be unnecessary.

https://github.com/usememos/memos/pull/2480/files#diff-e1e4065afaaae81d389a92c11ef9ac0fdbec4681bc20647843ce88dcf0193420L311

itschip: Ah, missed that. Sounds good!

itschip:
@boojack We have access to the user role on the frontend, correct? If so, I'll just disable the public option if you're not an admin. Does that sound cool?

web/src/components/MemoEditor/index.tsx
```diff
@@ -465,12 +472,15 @@ const MemoEditor = (props: Props) => {
...
             {VISIBILITY_SELECTOR_ITEMS.map((item) => (
-              <Option key={item} value={item} className="whitespace-nowrap">
+              <Option key={item} value={item} className="whitespace-nowrap" disabled={disableOption(item)}>
                 {t(`memo.visibility.${item.toLowerCase() as Lowercase<typeof item>}`)}
               </Option>
             ))}
           </Select>
         </div>
+        <div className="flex-grow">
+          {systemStatus.disablePublicMemos && <Typography color="warning">{t("memo.visibility.disabled")}</Typography>}
```
boojack, line 482: This warning text might be too long in mobile view. Let's remove it.

boojack:
> @boojack We have access to the user role on the frontend, correct? (HOST?) If so, I'll just disable the public option if you're not an admin. Does that sound cool?

SGTM

web/src/pages/MemoDetail.tsx
```diff
@@ -156,12 +165,15 @@ const MemoDetail = () => {
...
                     >
                       {VISIBILITY_SELECTOR_ITEMS.map((item) => (
-                        <Option key={item} value={item} className="whitespace-nowrap">
+                        <Option key={item} value={item} className="whitespace-nowrap" disabled={disableOption(item)}>
                           {t(`memo.visibility.${item.toLowerCase() as Lowercase<typeof item>}`)}
                         </Option>
                       ))}
                     </Select>
-                  </>
+                    <div className="flex-grow">
```
boojack, line 173: And this warning text should be removing too.

boojack (approved):

web/src/pages/MemoDetail.tsx
itschip: Done @boojack

PR #2512, opened 2023-11-14
docs: add obsidian plugin

quanru:
The plugin usage effect is as follows, taking a shortcut as an example:
https://github.com/quanru/obsidian-periodic-para/assets/11739753/3d967527-b021-4a15-acd4-f5e6b0701a99


boojack (approved):
LGTM!

Issue #625, opened 2022-11-28
heic格式的图片无法显示


xueyangcs:
### Describe the bug

如题，上传heic图片之后无法在网页中显示

### Steps to reproduce

1、在memos编辑框中点击文件按钮并上传heic格式图片
2、查看该memo，上传的图片显示为broken picture的图标

Zeng1998:
I think `heic` is not a common web image format now.

Some other apps convert this format to others for displaying in the browser.

PR #2527, opened 2023-11-17
feat: add support for convert HEIC on Apple devices

### LINKED ISSUE #2164: HEIC格式图片不支持
### Describe the bug

HEIC格式图片不支持，HEIC应该是比较有发展前途的图片格式

### Steps to reproduce

支持一下HEIC图片格式

_No response_

athurg:
HEIC is a common image format on iOS devices, but most browsers cannot preview it directly. Instead, all Apple OS has a built-in function that converts HEIC to a more common image format when uploading on the WEB.

However, there require the `accept` attribute of `<input>` must be set, and the value should be a `image/*` type, `*` is not worked. And, due to the limitations of the HTML standard, `accept` does not support excluding some types through the blacklist mechanism.

In the end, I duplicated the upload resource button, and now there're two buttons for upload :
1. First one still use the old `image` icon, and can only upload images with auto convert HEIC.
2. The new one with a paper clip icon for uploading any type of file, but without HEIC convert.

This PR will resolved #2164, #625, #70 .

### DIFF
--- a/web/src/components/CreateResourceDialog.tsx
+++ b/web/src/components/CreateResourceDialog.tsx
@@ -11,6 +11,7 @@ const fileTypeAutocompleteOptions = ["image/*", "text/*", "audio/*", "video/*", 
 
 interface Props extends DialogProps {
   onCancel?: () => void;
+  imageOnly?: boolean;
   onConfirm?: (resourceList: Resource[]) => void;
 }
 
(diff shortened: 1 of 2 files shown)


athurg:
User can manually choose whether enable the convert in the OS's dialog.

PR #2536, opened 2023-11-21
fix: correct some Chinese translation inaccuracy

Kellermaan:


Kellermaan:
Hi, I've found and fixed some inaccuracy in the Chinese translation of this project which may cause some confusing.

boojack (approved):

PR #2539, opened 2023-11-22
chore: remove invalid access token from db

athurg:
This PR just remove all invalid tokens while it removed from the cookie by reasons other than logout, such as expired or any others.


boojack (approved):
LGTM

Issue #2517, opened 2023-11-15
Postgres DB Driver Support


debugger24:
### Is your feature request related to a problem?

Postgres is widely used database, adding support for Postgres will be very handy for others too.

### Describe the solution you'd like

Either using a ORM or adding one more configuration for postgres db support.

Irvingouj:
Can I take this one @boojack ?

boojack:
@Irvingouj Of course you can, maybe adding one more configuration for Postgres is a better solution than using ORM. WDYT?

Irvingouj:
@boojack you mean follow the existing pattern and avoid using ORM?
Correct me if I'm wrong, currently for Mysql and SQLite, all implementation of the Driver interface is string concatenation of sqls. If you think in future, there will be no more database to support, we may just not add a new ORM for extra layer of abstraction and follow what is already there. 

However, using ORM will solve the problems where
 **If there is any database changes, we need to manually update all sqlite, mysql and postgres implementation, which is error prone and repetitive**

Though, Golang ORMs had very bad reputations. Alternatively, we may not necessarily need an ORM. The solutions we have is 
1. [sqlc](https://github.com/sqlc-dev/sqlc), where all we do is writting SQL quries and go code is generated for us, and it supports all sqlite, mysql and posgres.
2. sqlbuilders, though which sqlbuilder to chose is debatable, but most of them supports all three database we discussed here.


Let me know what do you think, and I will proceed accordingly.

Irvingouj:
> @Irvingouj We also think that Go's ORM is kind of not good, so we'd not to use it. As for the two options you mention, the `sqlc` looks a little weird to use. So I prefer the second way, but I feel it's pretty much the same as the existing way, except that now it's just splicing the string directly into the SQL string.

I will give an implementation from a sqlbuilder and support postgres, then we may explore the posibility of replacing it with current database changes

Irvingouj:
WIP
https://github.com/Irvingouj/memos/tree/TrySquirrel

PR #2569, opened 2023-12-03
feat: support Postgres

Irvingouj:
Added PostgreSQL support using [Squirrel](github.com/Masterminds/squirrel)) This comes after our talk in [Issue #2517](https://github.com/usememos/memos/issues/2517) .

- All automated tests are green. 
- Did a bunch of manual checks too, and everything looks good.

I'd love to get your thoughts and any feedback you have. 

@boojack 

To answer the questions we raised in the original issue, I think we can say, it is not worth it to provide a general abstraction to support all three different SQL databases. The SQL builders can do only so little (tried different ones), to help us build sql queries. Golang does not have a decent ORM at this point as well. Hence we will move on by changing the implementation of each driver independently when we have a database schema changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/usememos/memos
 go 1.21
 
 require (
+	github.com/Masterminds/squirrel v1.5.4
 	github.com/aws/aws-sdk-go-v2 v1.22.1
 	github.com/aws/aws-sdk-go-v2/config v1.22.1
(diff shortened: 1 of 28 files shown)


boojack (approved):
LGTM overall, please fixes the error from golang linter.

boojack (approved):
Thanks for your contribution! Let's merge first. I may do some refactoring later.

Issue #1843, opened 2023-06-17
Markdown tables support


dejanstojanovic:
### Is your feature request related to a problem?

I was trying to add a table to my note in common format and notice that it does not work

| column1 | column 2 |
|-----------|----------- |
| value 1   | valuee 2   |

It would be nice if we could add a table to a note content

### Describe the solution you'd like

Render out markdown table syntax

turtaf:
+1

rvnkcode:
Tables aren't rendered in `0.18.2`.

PR #2573, opened 2023-12-03
feat: tables support

shagr4th:
Column alignment is supported, as well as pipe escaping
Should fix Issue #1843


shagr4th:
Please do not merge for now, the regex does not match table rows starting without a pipe (as most parsers do)

jxpsert:
This is a much-needed feature. Checks seem to pass, except for the backend ones that haven't run and aren't relevant.
Merging this will greatly improve memos.

shagr4th:
Actually this is not quite merge-ready, because using a regex is quite limiting to determine the validity of a markdown table expression
Right now, the regex I wrote mandates the use of a starting "|" at the beginning of the header (which is optional in many parsers), and does not check if the headers count and hyphens count is the same (as it should) : this one could trigger false positives on text content not intended to be in a table

Also, inserting a wrongly formatted table alongside other tables, breaks some of them (even if they have a correct syntax)

Anyway, it's a good start I think

shagr4th:
Sorry for the multiple commits, it was more difficult I thought. Right now, this PR pass most of the formatting tests found in the [markdown-it test suite](https://raw.githubusercontent.com/markdown-it/markdown-it/master/test/fixtures/markdown-it/tables.txt)

The ones actually failing are related to backlash escaping (not supported right now in the parser), but this is not related to this table extension, so maybe for a future PR

boojack (approved):
LGTM. The syntax of `table` is really hard to implement with regular. Thanks for your contribution.

PR #2629, opened 2023-12-16
Update frontend.go to disallow crawling by default

jhademcconnell:
Setting robots.txt to disallow web crawlers by default would help Memos continue to be a privacy-first solution


boojack:
Actually this is an unneeded change. If the user doesn't set the instance url, it will simply return the error. https://github.com/usememos/memos/blob/main/server/frontend/frontend.go#L78

PR #2658, opened 2023-12-22
feat: use vite plugin pwa for generate right sw.js

quanru:
Existing sw.js is actually an empty implementation, not a true PWA


boojack (approved):
LGTM

Issue #2668, opened 2023-12-26
Date/Time Format Missing from Shareable Memo Images


jhademcconnell:
### Describe the bug

Creating Memo shareable images had the date/time on the top left of the Memo: `MM/DD/YYYY, HH:MM:SS AM/PM`

In the recent build the date is missing and the time is formatted in:
![IMG_4249](https://github.com/usememos/memos/assets/13276370/c859d809-db36-4e96-bb34-b9d087808316)

### Steps to reproduce

1. Build the latest Memos codebase
2. Create a Memo
3. Export the Memo as a shareable image
4. Observe the top left date/time format

### The version of Memos you're using

v0.18.1

jhademcconnell:
Tested and fixed with the latest build:

![image](https://github.com/usememos/memos/assets/13276370/b5ca525d-a3e4-4151-8b90-b89705bc27a6)

PR #2672, opened 2023-12-26
fix: date format in share memo dialog(#2668)

subks:
issue #2668


boojack (approved):

PR #2684, opened 2023-12-28
chore: fix misuse of package path instead of filepath.path

lincolnthalles:
As stated by https://pkg.go.dev/path, "path" is mainly for URLs, "path.filepath" for file systems


boojack (approved):
LGTM

PR #2693, opened 2024-01-01
Delete .vscode directory

ambi88dex:


Juneezee:
Could you explain why?

ambi88dex:
> Could you explain why?

I was suggesting deleting .vscode directory because if we somebody wants to have custom configs in .vscode, they would be overidding the folder. Secondly, the .vscode property is something which shouldn't be part of source code as it is not source code.

boojack:
`./web/.vscode` can also be deleted.

ambi88dex:
Sure. I can add that folder as well in the PR

ambi88dex:
Hi @boojack & @Juneezee , I've added the mentioned directories to be deleted.
Kindly give it a look!

boojack (approved):

PR #2695, opened 2024-01-02
fix: docker-compose.dev.yaml

nalderto:
While attempting to configure my development environment, I noticed that the Docker compose file for development was no longer working as expecting, as described by the [documentation](https://www.usememos.com/docs/contribution/development).  [This commit](https://github.com/usememos/memos/commit/638f17a02c8a02d9e94f47365d2fafd551b6aea0) moved the `docker-compose.dev.yaml` file into the `scripts` directory.  However, as there are numerous relative paths in the compose file, we need to update those.

Additionally, I noticed that the DB container wasn't configured properly and hopefully resolved that.

I can make a follow up PR to the [documentation repo](https://github.com/usememos/dotcom/blob/main/content/docs/contribution/development.md) once this is merged in.


.gitignore
```diff
@@ -16,6 +16,9 @@ build
...
 .idea
 
+# Docker Compose Environment File
+.env
+
 bin/air
 
-dev-dist
\ No newline at end of file
+dev-dist
```
nalderto, line 24: Added a newline to the end of the file

scripts/docker-compose.dev.yaml
```diff
@@ -1,8 +1,13 @@
+version: "3.0"
+name: memos-dev
 services:
   db:
     image: mysql
     volumes:
-      - ./.air/mysql:/var/lib/mysql
+      - ./../.air/mysql:/var/lib/mysql
+    environment:
+      MYSQL_ALLOW_EMPTY_PASSWORD: yes
```
nalderto, line 9: Some env var related to setting the password is [required](https://github.com/docker/kitematic/issues/2175).

boojack (approved):

Issue #1191, opened 2023-02-27
Unable to Access Storage Files in a Private Bucket


MiracleXYZ:
## Issue description

When setting up R2 object storage using the instructions provided in [this document](https://usememos.com/docs/storage) without specifying an `r2.dev` custom prefix, I am unable to access the storage file. Attempting to open the file in a new tab results in the following error message:

```
<Error>
  <Code>InvalidArgument</Code>
  <Message>Invalid Argument: Authorization</Message>
</Error>
```

However, after setting an `r2.dev` prefix, my photos preview correctly. It is worth noting that `r2.dev` is a "managed public access" feature, which is not recommended for production environments (as noted in [this document](https://developers.cloudflare.com/r2/data-access/public-buckets/#enable-managed-public-access-for-your-bucket)).

To ensure safe access to storage files in production environments, we recommend using [pre-signed URLs](https://developers.cloudflare.com/r2/data-access/s3-api/presigned-urls/). Are there any plans to support pre-signed URL access in the future?

## Steps to reproduce

1. Follow the instructions provided in [this document](https://usememos.com/docs/storage) to set up R2 storage.
2. Upload a file to the storage.
3. Attempt to preview the file.

## Screenshots or additional context

_None at the moment._

Zeng1998:
We store the URLs of S3 resources in the database as external links, `r2.dev` or a custom domain name is enough.

MiracleXYZ:
Something new: After the latest commit (#1190), newly uploaded files have the original URL as the external link, instead of using the custom (e.g. `r2.dev`) URL prefix. This issue is causing problems with accessing the files on the custom domain.

MiracleXYZ:
> We store the URLs of S3 resources in the database as external links, `r2.dev` or a custom domain name is enough.

Using pre-signed URLs would provide an expiration period for access to stored data, adding an additional layer of security and control. While `r2.dev` links may be convenient for temporary public access, they do not offer the same level of security and control as pre-signed URLs.

MiracleXYZ:
I will try to submit a pull request to implement pre-signed URLs when I have the time...

crowley-mark:
I followed the instructions for R2 twice on the day 0.11.0 was released. For both, the result was that I was not able to access the resource once successfully uploaded to R2.

+1 from me.

fierceX:
+1 It is important to read files in private storage buckets through programs instead of changing the bucket to public. This can avoid important files from being leaked.

[...]

PR #2696, opened 2024-01-02
feat: add {uuid} in path template when using local storage or S3

ertuil:
Add an addition tag `{uuid}` to the `replacePathTemplate`.

It is a workaround to leak the public links of a resource when using S3-based object storage. Currently, all resource blobs stored in S3 (R2, OSS) are set to be public. It is insecure as the resources for the private memos are also accessible on the Internet. It happens in many situations. For example, the adversary might gauss the private resources' URLs based on the public memos' URLs even using a timestamp. Also, the resource's accessibility will not be modified when the memo's accessibility is changed.

Using an additional {uuid} might reduce this risk.  Actually, it is better to use a signed URL, but it might bring a large range refaction. As a result, using an additional UUID might help with little modification to the current structure. (related issues #1191 )

Meanwhile, it is also possible to avoid filename conflict using local storage.

### DIFF
--- a/api/v1/resource.go
+++ b/api/v1/resource.go
@@ -354,6 +354,8 @@ func replacePathTemplate(path, filename string) string {
 			return fmt.Sprintf("%02d", t.Minute())
 		case "{second}":
 			return fmt.Sprintf("%02d", t.Second())
+		case "{uuid}":
+			return util.GenUUID()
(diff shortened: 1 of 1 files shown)


boojack (approved):
LGTM

PR #2698, opened 2024-01-02
chore: improve resource internal_path migrator

lincolnthalles:
- handle mixed path styles
- handle Windows paths
- add tests
- run the migration in a separate goroutine, so server startup doesn't take too long on large databases and/or low-end hardware
- log basic post-migration stats like `INFO    store/migrator.go:52    migrated 34219 local resource paths in 42.8559848s`


boojack (approved):
LGTM

Issue #2691, opened 2024-01-01
hide database url (dsn) from logs in prod mode


moinologics:
### Describe the solution you'd like

hide database url when runs in prod mode as it is a good security practise

### Type of feature

API

boojack:
`DSN` will not output in the API.

PR #2699, opened 2024-01-03
Remove logging of DSN

jhademcconnell:
DSN in both dev/prod includes sensitive info used to log into the DB instance you're connecting to: https://github.com/usememos/memos/issues/2691


boojack:
Logging is more conducive to debugging and does not need to be removed for now.

PR #2716, opened 2024-01-08
feat: export Memos as Markdown FIles

nalderto:
Allows for exporting individual Memos as Markdown files from the "Share Memo" dialog


boojack (approved):

PR #2732, opened 2024-01-10
fix: pnpm install failed in docker

athurg:
This PR resolved a problem, that the `pnpm install` may failed with `ERR_PNPM_LINKING_FAILED` in some case.

This is because *pnpm* is works by create hard link, which is not working well in docker container.

So we use `npm` to replace `pnpm`.


athurg:
Still working

PR #2756, opened 2024-01-14
chore: use webp compression on logo

lincolnthalles:
- Logo size reduced from 311 KB to 36 KB.
- Point metadata image URL to local logo instead of remote.


boojack (approved):
LGTM

Issue #2211, opened 2023-09-12
Inconvenient backups / exports


maticboncina:
### Is your feature request related to a problem?

Since people invest heavily into their journals, and this is a self-hosted software, it would be nice to see if we could get some option to export markdown PDF exports. I used to use the MiniDiary app on MacOS and they had pretty nice options. 

We need to have some system that allows us to, independently of our hosting provider, make backups and ensure data integrity.

### Describe the solution you'd like

Probably the simplest integration would be to just use the already existing storage configuration used to store uploaded assets, and also allow to set "daily/weekly/monthly" internal when SQLite database would be copied to the destination (for example S3 compatible storage), thus allowing quick, reliable, and painless restores.

I am not sure how would it go with importing, but I am sure that even if an option exists to just overwrite the old database via the SFTP/Database tool, that will be fine by most users.

maks:
Given that there is a REST API available, it should be possible to do both imports and exports via the API.

For example the curl example from the docs:
```
curl http://localhost:5230/api/v1/memo \
   -H "Accept: application/json" \
   -H "Authorization: Bearer $TOKEN"
```
will list memos, while this [python script](https://discord.com/channels/1017113925776588902/1045138348165050409/1149939344258367609) posted by someone on the Discord shows how to do a very basic import using the API.

Admittedly this requires some programming knowledge but perhaps a kind soul will come along and create some user friendly scripts or GUI tools to do imports/exports.

yaoniplan:
> Given that there is a REST API available, it should be possible to do both imports and exports via the API.
> 
> For example the curl example from the docs:
> 
> ```
> curl http://localhost:5230/api/v1/memo \
>    -H "Accept: application/json" \
>    -H "Authorization: Bearer $TOKEN"
> ```
> 
> will list memos, while this [python script](https://discord.com/channels/1017113925776588902/1045138348165050409/1149939344258367609) posted by someone on the Discord shows how to do a very basic import using the API.
> 
> Admittedly this requires some programming knowledge but perhaps a kind soul will come along and create some user friendly scripts or GUI tools to do imports/exports.

`vim ./batch_upload.sh` # Go into your unzipped directory and edit it
```bash
#!/bin/bash

# Set your access token
[...]

PR #2757, opened 2024-01-14
feat: Export all user Memos as a .zip of Markdown files

nalderto:
This PR adds an "Export Memos" button to the Setting => My Account page.  This will download a zip folder, generated by the server, containing all of the user's Memos in Markdown files.  The individual memos filenames are named after their creation date in RFC3339 format.  This PR is should partially address this [open enhancement](https://github.com/usememos/memos/issues/2211) to improve exports with Memos.

I also resolved some memory leaks with downloading images and files in the share dialog.

image


web/src/components/ShareMemoDialog.tsx
```diff
@@ -59,14 +61,9 @@ const ShareMemoDialog: React.FC<Props> = (props: Props) => {
...
+    const url = URL.createObjectURL(blob);
     downloadFileFromUrl(URL.createObjectURL(blob), `memos-${getDateTimeString(Date.now())}.md`);
-  };
-
-  const downloadFileFromUrl = (url: string, filename: string) => {
-    const a = document.createElement("a");
-    a.href = url;
-    a.download = filename;
-    a.click();
+    URL.revokeObjectURL(url);
```
nalderto, line 66: This cleans the memory allocate for createObjectURL

boojack:
Thanks for the PR, there are some points I would like to clarify about import/export:

* I think import/export operations should be very infrequent;
* It's more convenient to import/export directly through the database;
* Memos now is mainly suit for person/family or friends, not for strangers on the internet. All data is stored on the user's machine.

So I don't think we need to do export in UI. Are there any other reasons about this feature? @nalderto

nalderto:
* I agree import/export is infrequent, but don't necessarily think that means that it shouldn't be included in the UI.  Folks don't change things like "Allow user signup" that often, but that doesn't mean it shouldn't be in the UI.
* Presently in Memos, yes, it is easier to import/export via the database.  However, I think that is something we should strive to make easier.  If folks decide they want to move to another Notes app, we shouldn't expect them to have to access the DB directly to export their data.  This is a feature that is [expected](https://www.chrisblunt.com/notes-portability/) and [desired](https://www.reddit.com/r/ios/comments/18jw2nl/apple_we_are_begging_you_let_us_export_our_notes/) with other notes apps.
* Less tech savvy friends/family should be able to export their data without having to have their Memos admin export their data manually.

For example, Joplin has this functionality 
image

I foresee a separate tab in Settings for Import/Export instead of always having it in "My Account", however since we would only have export functionality at the moment, I figured "My Account" is fine for now.

boojack:
Not the same as Joplin, memos is deployed by someone and can be used by others. In case one user has a large amount of data, it may take up a lot of resources when exporting or even cause the service to hang. So perhaps export should be an administrator's function.

nalderto:
That's fair, although with the infrequency of exports occurring and friends/family use case of Memos, I don't see that being a major issue.  Making it an admin only is better than nothing, although that makes it difficult to export other users' data.  The file creation is all done in memory, so I would hope that performance wouldn't be impacted significantly with an export

boojack:
SGTM. It's better to use APIv2(protobuf + gRPC) to implement the export api instead of the deprecated one.

nalderto:
Gotcha!  I wasn’t sure if I should use v1 or v2.  I can re-implement this with the v2 API in a couple weeks if you’d like.

erbanku:
@boojack  @nalderto

**Merging the PR is highly advisable for those like me needing reliable export features.** 

I frequently use Obsidian and various other note-taking apps for my day-to-day tasks, and for quick notes, I turn to Memos because it's incredibly user-friendly & handy. In this case, I aim to export my memos to a local storage solution to unify all my notes in a single location. Having this export capability is a critical requirement for me.

Previously, before the Open API was phased out, I could export all my memos seamlessly using the [memos-export service](https://memos-export.vercel.app/). Regrettably, this is no longer possible.

erbanku:
I can backup database files, but the database files are very inconvenient for further interactions with notes.

I think we can not implement the import function, but the export function is necessary. There are already many users requesting a similar function in GitHub Issues, I hope it can be implemented.

Tslmuun:
thank you for your work.Memos as a distributed platform, I believe the export/import feature is very important, users can freely migrate to another site. thank you again.

nalderto:
I'm closing this PR, because I made a [new PR](https://github.com/usememos/memos/pull/2854) to implement the streaming gRPC endpoint to download the zip folder.

PR #2760, opened 2024-01-15
fix: change use-set priority

thlstsul:
The user settings(locale, appearance) are not in use when restart broswer


boojack (approved):
LGTM

PR #2768, opened 2024-01-16
feat: Add URL paste handler

nalderto:
I added this with [this PR](https://github.com/usememos/memos/pull/2763/files), but it was removed with [this commit](https://github.com/usememos/memos/commit/5a723f00fa9d9a6a10e5a929c62abfe31ae2a751).

This PR hooks onto the paste event, if text is highlighted, and there is a valid URL in the clipboard, then the editor will use markdown syntax to turn the highlighted text into a hyperlink with the URL in the clipboard.  This matches the behavior of GitHub.  This especially makes it easier to paste URLs on mobile.

### DIFF
--- a/web/src/components/MemoEditor/handlers.tsx
+++ b/web/src/components/MemoEditor/handlers.tsx
@@ -15,20 +15,7 @@ export const handleEditorKeydownWithMarkdownShortcuts = (event: React.KeyboardEv
   }
 };
 
-const styleHighlightedText = (editor: EditorRefActions, delimiter: string) => {
-  const cursorPosition = editor.getCursorPosition();
-  const selectedContent = editor.getSelectedContent();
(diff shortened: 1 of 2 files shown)


boojack:
> we supported automatically turn URLs into links. That means if you type https://www.usememos.com, it will automatically be turned into a link even though you haven’t used brackets.

FYI, so the Markdown standard link syntax is optional.

https://www.usememos.com/docs/getting-started/content-syntax#links

nalderto:
[Screencast from 2024-01-15 17-33-07.webm](https://github.com/usememos/memos/assets/25762130/950d6e90-1e1c-47de-8440-936ef8329bba)

This addresses a different issue I show in the screencast.  This is where you want to paste a URL, but want it to be a hyperlink to an existing typed string

boojack (approved):
LGTM

PR #2769, opened 2024-01-16
chore: fix store tests on Windows

lincolnthalles:
It's just a matter of explicitly closing the database, so that TempDir.removeAll doesn't fail. This doesn't affect other systems.

Screenshot_1 => Screenshot_2


boojack (approved):
LGTM

Issue #2765, opened 2024-01-15
DatePicker displays memos in UTC date.


harveyghq:
### Describe the bug

`Timeline`'s datepicker displays memos in UTC date (calculated on server side). Users can get confused by this gap.

### Steps to reproduce

1. Write a memo.
2. Adjust created time to the previous UTC day (e.g. for a `UTC+8` user, adjust to 7 a.m. localtime.)
3. Switch to `Timeline`, click the `DatePicker`.
4. Now it display a badge for yesterday.

### The version of Memos you're using

[0.18.2](https://github.com/usememos/memos/commit/ad94e8e3c6cae4c76938a15d66ad156ca6d6b054)

harveyghq:
Maybe we send local timezone to the server to calculate, or require server to return all raw timestamps (at a higher cost of course).

harveyghq:
sry, the bug still exists. And the reason is that server side lacks knowledge of local timezone required for converting timestamps  into accurate dates. Currently the `GetUserMemosStats` api returns an array of converted-yet-standardlized UTC dates. I'm afraid there's insufficient data to make further adjustments during the client's rendering process.

nalderto:
@harveyghq You're right, this is still a bug.  I misunderstood the issue.  With `GetUserMemosStats` just returning UTC date, we'll need to change the API.  @boojack This should be re-opened

meizigz:
still exist.

RoccoSmit:
Fixed with pr https://github.com/usememos/memos/pull/4046

PR #2770, opened 2024-01-16
fix: DatePicker by passing in Timezone to API

nalderto:
This should actually resolve [this bug](https://github.com/usememos/memos/issues/2765).  I pass in timezone to `GetUserMemosStats` so that the dates generated match the appropriate timezone of the client.

[Video demonstrating the fix](https://github.com/usememos/memos/assets/25762130/9fef9e69-4f93-490b-ad00-ee11aedeaaf3)


boojack (approved):

PR #2771, opened 2024-01-16
fix: server overrides user's locale/appearance

### LINKED ISSUE #2767: 0.18.2 Language set invalid
### Describe the bug

After updating to docker version 0.18.2, set the language to ‘简体中文’, then refresh the page, and the language returns to the default English.

### Steps to reproduce

1.updating to docker version 0.18.2
2.set the language to ‘简体中文’
3.refresh the page
4.the language returns to the default English

docker v0.18.2

_No response_

harveyghq:
Close #2767.

Locale usage sequence: `storage locale` > `user locale` > `server's default locale`. (same for appearance)

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -82,10 +82,10 @@ const App = () => {
   }, [systemStatus.customizedProfile]);
 
   useEffect(() => {
-    document.documentElement.setAttribute("lang", locale);
     const { locale: storageLocale } = storage.get(["locale"]);
     const currentLocale = storageLocale || userStore?.userSetting?.locale || locale;
     i18n.changeLanguage(currentLocale);
+    document.documentElement.setAttribute("lang", currentLocale);
     if (currentLocale === "ar") {
       document.documentElement.setAttribute("dir", "rtl");
     } else {
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -83,11 +83,10 @@ const App = () => {
 
   useEffect(() => {
     document.documentElement.setAttribute("lang", locale);
-    i18n.changeLanguage(locale);
-    storage.set({
-      locale: locale,
-    });
-    if (locale === "ar") {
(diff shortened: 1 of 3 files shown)


web/src/App.tsx
```diff
@@ -83,24 +83,20 @@ const App = () => {
 
   useEffect(() => {
     document.documentElement.setAttribute("lang", locale);
```
boojack, line 85: Maybe move this line after `i18n.changeLanguage(currentLocale);`

boojack (approved):
LGTM otherwise

harveyghq:
**Test: user can change user locale in `Preferences` / specify a new locale when logging in. Storage locale overrides user locale.**

test1

**Note: currently when a user doesn't set locale, the default user setting `en` and `follow system` will override server's setting.**
Maybe in the future, we need to use server's setting locale/appearance as default user setting. (see [here](https://github.com/usememos/memos/blob/15e6542f0df709611eb27d5f454537e5f0dddabd/api/v2/user_service.go#L233C47-L233C47))

web/src/App.tsx
harveyghq: Done :)

boojack:
FYI, fixed in https://github.com/usememos/memos/commit/9ffd827028d572c24db7d54e29952b95862e25e8

harveyghq:
Some scenarios to consider:
- A fresh start: A new user signs up and **specifies their preferences** on the sign-up page. These preferences should remain **unchanged** later.
- A signed-up user opens a **new** browser and logs in without modifying their preference settings on the login page. In this case, the server's default preferences should change to **match** the user's previously specified preferences.
- A signed-up user **specifies new preferences** during login, or if they haven't cleared their browser's storage (with last used preferences). These preferences should **not** be replaced by their preferences stored in the database, and may even require an update to the user's preferences.

Issue #2724, opened 2024-01-09
Add hint/info/tooltip what the button "Vacuum" does


q2apro:
### Describe the solution you'd like

Under "/setting" on the top right, there is a button labeled "Vacuum". 

Add hint/info/tooltip what the button "Vacuum" does. 

Most users have never seen a "vacuum" option, so have no idea what it does.

I do not know what it does.

### Type of feature

User Interface (UI)

MehadND:
I believe I can work on this issue but not sure exactly what the "vacuum" option does. However, if anyone can explain then I can resolve this issue. From some google searches I think this button deletes unused data/files from database. Again not entirely sure so please guide me.

Thanks,
Mehad

MehadND:
Sorry for being a bit stupid but I am getting this error and not sure exactly what to do?

![image](https://github.com/usememos/memos/assets/34424878/2877bbc7-79f0-4c43-abb7-21600a228fd9)


This is my first time with Vite so maybe I'm doing something wrong there.

The steps to produce this error:
- Clone Repo
- open the project in vs code and then navigate to `web` directory and run `npm install`
- then run `npm run dev`

MehadND:
@boojack, when I run the memos in dev I cannot go past the auth for some reason, even with the default login credentials.

boojack:
> But I would like to do it outside docker.

@MehadND You can build a binary with source code for your OS. Or follow https://github.com/lincolnthalles/memos-builds

MehadND:
Ok will look into that but if you can tell me how to fix the vite issue with authentication. Because whenever I change code and want to see my change I can't because I can't go past the sign in/up page

boojack:
@MehadND This issue is not appropriate for your problems. Let's take a deeper look at them on discord: https://discord.com/invite/tfPJa4UmAv. You can DM to me.

MehadND:
#2778

PR #2778, opened 2024-01-17
chore: added tooltip for vacuum button

MehadND:
A lot of users were confused about the use of 'Vacuum' button in the settings->system. Therefore I added a tooltip for this.

At the moment the tooltip text is only in English language.

From [issue #2724](https://github.com/usememos/memos/issues/2724)

### DIFF
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -256,6 +256,7 @@
       "disable-public-memos": "Disable public memos",
       "max-upload-size": "Maximum upload size (MiB)",
       "max-upload-size-hint": "Recommended value is 32 MiB.",
+      "vacuum-hint": "Cleans up unused or obsolete data.",
       "additional-style": "Additional style",
       "additional-script": "Additional script",
       "additional-style-placeholder": "Additional CSS code",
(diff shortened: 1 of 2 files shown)


web/src/locales/en.json
```diff
@@ -256,6 +256,7 @@
       "disable-public-memos": "Disable public memos",
       "max-upload-size": "Maximum upload size (MiB)",
       "max-upload-size-hint": "Recommended value is 32 MiB.",
+      "vacuum-hint": "Cleans up unused or obsolete data.",
```
jxpsert, line 259: I think obsolete is (nearly) a synonym of unused. I personally think only 'unused' would look better.

jxpsert (commented):
> At the moment the tooltip text is only in English language.

Was afraid you'd hardcoded the text from reading this, but you used i18n- good!

One comment on the text itself- not entirely important.
Code looks good, though!

web/src/locales/en.json
MehadND: Ok, thanks. Appreciate it

boojack (approved):

Issue #858, opened 2022-12-25
Email reminders and SMTP


sumitmamoria:
### Is your feature request related to a problem?

I want something like a "remind me later" and "recurring reminders" buttons against all individual entries, specially for TODOs. The alerts/reminders can be pushed over email via any public or private SMTP server. 

### Describe the solution you'd like

Adding SMTP server configuration to the application along with  "remind me later" and "recurring reminders" buttons on the UI might be worthwhile.

ChasLui:
I think we can abstract the push notification service, like [novuhq/novu](https://github.com/novuhq/novu), which already supports future push from various providers, such as email, chat, etc.

sumitmamoria:
> I think it is possible to abstract a push notification service, like [novuhq/novu](https://github.com/novuhq/novu). Support future push from multiple providers, such as email, chat, etc.

Thats right. I like "novuhq/novu", but even a basic SMTP support could be really helpful.

ItsNoted:
I think implementing [Apprise](https://github.com/caronc/apprise) would be amazing. This way we can choose whatever we want to use as a notification be it email or any of the hundreds of push notification services.

ItsNoted:
Any thoughts on this @boojack

rounakdatta:
+1, a schedule for later / reminder system would be awesome!

PR #2791, opened 2024-01-19
feat: add a webhook action for deleted memos

ertuil:
Add a webhook when the memos are deleted. Currently, the webhooks are only triggered when the memos are created (C) or updated (U). But the webhook for deleting (D) is missing. I think it is better for the webhook to activate for the entire CURD procedures.

I love memos, and use it heavily. However, I need it to send me a notice as a reminder, like issue #858. I realize that this function is in the roadmap, but I cannot wait to have it. As a result, I recently worked on my community tool [memos-reminder](https://github.com/ertuil/memos-reminder) as a temporary solution. It reads memos using the webhook API, and reads some date/time tags in it, and then sends me reminder emails.

However, I find that the current webhook is not completed, as it will not notify my memos-reminder to delete the related timers when the memos are deleted.

### DIFF
--- a/api/v1/memo.go
+++ b/api/v1/memo.go
@@ -632,7 +632,6 @@ func (s *APIV1Service) DeleteMemo(c echo.Context) error {
 		}
 	}
 
-
 	if err := s.Store.DeleteMemo(ctx, &store.DeleteMemo{
 		ID: memoID,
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM

PR #2792, opened 2024-01-19
fix: incorrect timeline month display

### LINKED ISSUE #2787: Timeline month display incorrect & Heatmap in timeline page not functioning?
### Describe the bug

It seems that the timeline page in the new version displays the month incorrectly? I have been using two computers and two browsers to test that, both my docker and the demo page, the month always shows one month ago.

That is, for the memos created in Jan 2024, the timeline will show it as created in Dec 2023. Please see the screenshot for demonstration.

Besides, I see that there is the heatmap relocated to the timeline page, but I do see any color variations or click to navigate to that date function. Is that still developing?

Thank you very much.

### Steps to reproduce

1. Go to timeline page
2. Month is not the current month, but current month - 1

0.18.2

_No response_

xwjdsh:
Close #2787


boojack (approved):
LGTM for now. Maybe we can use some lib to do such things.

jon-f-novastor:
Thanks for the fix guys.

PR #2806, opened 2024-01-20
fix: double click to edit memo

MehadND:
The user was not able to edit the memo by double-clicking in v18.0.2, however, this was available in previous versions.

Hence this PR aims to fix this bug.

Note: You are only able to edit the memo (by double click) on the homepage, as the UI is more convenient in the Memo Detail page.

### DIFF
--- a/web/src/components/MemoView.tsx
+++ b/web/src/components/MemoView.tsx
@@ -84,9 +84,13 @@ const MemoView: React.FC<Props> = (props: Props) => {
       }
     }
 
-    if (e.detail === 2) {
-      handleEditMemoClick();
-    }
+    // Use setTimeout for double-click recognition, as pointed out by @daveclinton on GitHub.
+    setTimeout(() => {
+      if (e.detail === 2) {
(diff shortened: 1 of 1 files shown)


webysther:
Hi @boojack any chance to review this one? As a old google keep user I use a lot double click and I really miss this one.

web/src/components/MemoView.tsx
```diff
@@ -163,6 +163,10 @@ const MemoView: React.FC<Props> = (props: Props) => {
         showPreviewImageDialog([imgUrl], 0);
       }
     }
+
+    if (e.detail === 2) {
```
daveclinton, line 167: You could use a timeout to ensure a double-click is recognized correctly, even if clicks are slightly close together.

daveclinton (changes requested):

web/src/components/MemoView.tsx
MehadND: Thank you for your suggestion!

PR #2809, opened 2024-01-20
feat: add notice when sharing private links

moehanabi:
I am not familiar with ts coding, so please make a double check before merge.

Besides, I think it is a good idea to support accessing private memos anonymously, as long as you have a link and the author of the memo agrees. And I think password access is also a nice feature. To accomplish these function, it may need to add some aurguements. What do you think about it?


boojack:
> And I think password access is also a nice feature. To accomplish these function, it may need to add some aurguements. What do you think about it?

IMO, let's keep it as simple as possible for now.

PR #2819, opened 2024-01-22
fix: increase MySQL blob size

### LINKED ISSUE #2818: After the database is switched to MySQL, the large file is uploaded incorrectly
### Describe the bug

In the default database, you can upload files larger than 700 MB, but when the database is switched to mysql, an error is reported when the upload file exceeds the database limit. Change the database size limit again max_allowed_packet = 2048 MB

After that, the error message is "Data too long for column 'blob' at row 1" is returned

### Steps to reproduce

{"time":"2024-01-21T19:07:06Z","latency":"1.969964607s","method":"POST","uri":"/api/v1/resource/blob","status":500,"error":"code=500, message=Failed to create resource, internal=Error 1105 (HY000): Parameter of prepared statement which is set through mysql_send_long_data() is longer than 'max_allowed_packet' bytes"}

{"time":"2024-01-21T19:12:49Z","latency":"2.77445071s","method":"POST","uri":"/api/v1/resource/blob","status":500,"error":"code=500, message=Failed to create resource, internal=Error 1406 (22001): Data too long for column 'blob' at row 1"}

Version: v0.18.2

_No response_

lincolnthalles:
Although 4GB from a LONGBLOB is probably too much, current MySQL's MEDIUMBLOB is only 16MB, and SQLite default to 1GB.

Closes #2818


boojack:
Refer to https://github.com/usememos/memos/issues/2818#issuecomment-1903936243

PR #2820, opened 2024-01-22
chore: added link preview card for YouTube (video + short) and Twitter

MehadND:
Added link preview card for YouTube (video + short) and Twitter


web/src/components/MemoView.tsx
```diff
@@ -216,7 +216,7 @@ const MemoView: React.FC<Props> = (props: Props) => {
               <span className="btn more-action-btn">
                 <Icon.MoreVertical className="icon-img" />
               </span>
-              <div className="more-action-btns-wrapper">
+              <div className="more-action-btns-wrapper z-10">
```
MehadND, line 219: `z-10` is needed becuase it causes issues if there are preview cards in a memo on top and it makes the action menu not useable, in other words, cannot edit, delete a memo otherwise.

This makes the action menu appear on top of preview cards, hence allowing user to access the action menu

MehadND:
please do take into consideration that the Twitter/X cards do tend to take a few seconds to display data. Maybe this can be improved with caching, etc.

Issue #2837, opened 2024-01-26
url parse dsn change mysql password to lowercase


zwpaper:
### Describe the bug

https://github.com/usememos/memos/blob/f8b9a83d4ad99accd6c7f66b9258c0656c67e9de/store/db/mysql/mysql.go#L117

use url to parse dsn which may recognise password as host, and set it lower cases

we should use the MySQL lib to parse it and set the options.

### Steps to reproduce

1. use a password with upper case in MySQL dsn
2. you will find it would be set to lowercase

### The version of Memos you're using

0.19

PR #2838, opened 2024-01-26
fix: use mysql to parse dsn

zwpaper:
fix https://github.com/usememos/memos/issues/2837


boojack (approved):

PR #2843, opened 2024-01-27
feat: add buf plugin to generate openapiv2 spec

lincolnthalles:
I was unable to use the plugin opt `openapi_configuration` to supply the additional API configuration file (it simply doesn't find the file), but it's a matter of API users replacing the first 4 lines of `v2.swagger.yaml` with something like this:

```yaml
swagger: "2.0"
host: "demo.usememos.com"
schemes:
  - https
  - http
securityDefinitions:
  Bearer:
    type: apiKey
    in: header
    name: Authorization
    description: >-
      Enter the Memos token with the `Bearer` prefix, e.g. "Bearer abcde12345"
security:
  - Bearer: []
info:
  title: Memos API
  version: 2.0.0
``` 

This enables interacting with v2 API using Insomnia built-in Swagger-UI or even using the web-hosted [Swagger Editor](https://editor.swagger.io).
image

By the way, there's not much control of file name, the plugin will always append ".swagger.yaml".

It's just a matter of running `buf generate` after changes on v2 API and commiting `v2.swagger.yaml` to keep the spec updated.

### DIFF
--- a/api/v2/v2.swagger.yaml
+++ b/api/v2/v2.swagger.yaml
@@ -1,0 +1,2113 @@
+swagger: "2.0"
+info:
+  title: api/v2/activity_service.proto
+  version: version not set
+tags:
+  - name: ActivityService
+  - name: UserService
+  - name: AuthService
+  - name: InboxService
+  - name: MarkdownService
+  - name: ResourceService
+  - name: MemoService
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM

PR #2854, opened 2024-01-29
feat: Export all user Memos as a .zip of Markdown files

nalderto:
This PR adds an "Export Memos" button to the Setting => My Account page. This will download a zip folder, generated by the server, containing all of the user's Memos in Markdown files. The individual memos filenames are named after their creation date in RFC3339 format. This PR is should partially address this https://github.com/usememos/memos/issues/2211 to improve exports with Memos.

I also resolved some memory leaks with downloading images and files in the share dialog.

This is a second attempt after [this PR](https://github.com/usememos/memos/pull/2757).  This time, I am using a gRPC steaming endpoint to download the .zip file.
Screenshot from 2024-01-28 22-16-00


proto/api/v2/memo_service.proto
```diff
@@ -90,6 +90,10 @@ service MemoService {
     option (google.api.http) = {get: "/api/v2/memos/stats"};
     option (google.api.method_signature) = "username";
   }
+
+  rpc MemosExport(MemosExportRequest) returns (stream MemosExportResponse) {
```
boojack, line 94: Why did it change from `ExportMemos` to `MemosExport`? IMO, I prefer `ExportMemos`.

boojack (approved):
LGTM otherwise

proto/api/v2/memo_service.proto
nalderto: Good call!  I switched it back

boojack (approved):
Thanks for your contribution!

nalderto:
@boojack I saw you [changed](https://github.com/usememos/memos/commit/1e07b70d2391c18e1b4eba2a21d21eb5a823ee44#diff-3a5c5d01f39f818e7d0a853984fcb5abd26bdef31466138de9087938b0cc6ccf) exporting to be unary instead of streaming.  The reason I made it streaming in the first place was to support larger file sizes.  I believe gRPC only support 4MB payloads by default.  If/when resource export support is added, it is likely that the exports zip folders will be larger than 4MB.

nalderto:
@boojack Ah, you're right about it just being for requests.  Thanks for fixing that!

ocean-haiyang:
where did this feature go??

ocean-haiyang:
Is there any existing way to export?

PR #2860, opened 2024-01-29
fix: the same-storage check in the new `pre-sign` feature

ertuil:
I tried the recent `pre-sign` feature in #2855 , but it does not work for my Aliyun OSS. In particular,  I tried this feature with the latest test docker image with Aliyun OSS. I set presign to true and restarted the container several times. However, it seems that it doesn't work out, as the ExternalUrls are not changed after I saw a 'links pre-signed' message in my log.

After an examination, I find the debug. In the original code, it checks if the endpoint matches the ExternalLinks's hostname. However, it does not work out in my Aliyun OSS cases, and I return the pre-sign link 'As-Is' to the original one.

My endpoint is `https://oss-cn-shanghai.aliyuncs.com`, while my ExternalLinks's hostname is `<bucket-name>.oss-cn-shanghai.aliyuncs.com`. Therefore, they can not match with a simple `strings.Contain()`.

# What I did:
* add additional codes to match the endpoint and the resource's ExternalLinks's hostname.
* disable set the new resource's ACL to 'public_read' if `presign` (#2855 ) feature is enabled.


boojack (approved):
LGTM

PR #2861, opened 2024-01-29
fix: month grouping error in timeline page

### LINKED ISSUE #2817: Timeline stuck on the year 2023
### Describe the bug

Timeline is stuck on Dec 2023 (as the latest date), even though there are Jan 2024 created Memos entries in Memos v18.0.2. Entries created in Nov 2023 show in the Timeline as Dec 2023. I'm seeing ("QSnexus" user, as well as one other user "Stayer", in the Memos Discord > Questions chat on 01/20/24) that in my memos v0.18.2 instance, with a few memos created in Jan 2024, the Timeline is still stuck on Dec 2023. I even tried restarting the instance and did a vacuum but that didn't help. I'm not seeing a bug for that one on the GitHub. They redid this feature in this version, so probably it is a bug.

### Steps to reproduce

1.Make sure your Linux machine where the Memos instance is already installed on is using Pacific Time Zone and the clock and date is showing correct, at least that it is already in the 2024 year.
2. Have at least one Memo note added that has a created date of January 2024. (Not sure if this step is required but this is what I had in place).
3. Upgrade to Memos v0.18.2 from v0.18.1 on Ubuntu Linux x64 for the Docker container variant. I'm not sure if this issue would affect clean installs or only upgrade installs.
4. Login to Memos via Chrome or Firefox web browser.
5. Click on the Timeline and it will be stuck at December 2023 with no way of moving to January 2024. That is even though you already have a January 2024 created note.
6. Go back to the Home page of Memos.
7. Create one more Memos note while in the year 2024.
8. Make sure that new note saves and lists in Home tab.
9. Click on Timeline.
10. You should see while in Timeline view it is still stuck on December 2023. Restarting the instance again and performing a vacuum at this point won't help.

0.18.2

Timeline view, viewed on 01/21/24:

Screenshot_20240121-094728

Viewing individual Memos created in January 2024 in Home tab, by clicking on each one to view the entire entry to see the date timestamp of each entry:
Screenshot_20240121-094650~2

Screenshot_20240121-094642~2

Even the official Memos Demo instance website (https://demo.usememos.com/timeline) has the same bug, seen here when checking on 01/22/24:
Screenshot_20240122-184451

xwjdsh:
Close #2817

---

For example, if user in the New York timezone, then `new Date('2024-01').getFullYear()` returns 2023, we should treat it as UTC time or local time with an offset.

### DIFF
--- a/web/src/components/ActivityCalendar.tsx
+++ b/web/src/components/ActivityCalendar.tsx
@@ -1,6 +1,6 @@
 import { Tooltip } from "@mui/joy";
 import classNames from "classnames";
-import { getNormalizedDateString } from "@/helpers/datetime";
+import { getNormalizedDateString, getDateWithOffset } from "@/helpers/datetime";
 
 interface Props {
   // Format: 2021-1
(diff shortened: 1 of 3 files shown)


boojack (approved):
LGTM

PR #2865, opened 2024-01-30
chore: add en-GB language

lincolnthalles:
en-GB is just a copy of en-US, at least for now, but it solves an issue with date format that every now and then some user complain.

`en` will use the format mm/dd/yyyy, and simply selecting `en-GB` will format dates as dd/mm/yyyy.

British English
image

English
image


boojack:
If the translation of `en-GB` will be always the same as `en`, maybe we can do some hacks. https://github.com/usememos/memos/pull/2865/files#diff-69661a91c7ad27e32735ba8890bfe6547e961e462b5ecf7fc7b11f3329701c79R52

lincolnthalles:
> If the translation of `en-GB` will be always the same as `en`, maybe we can do some hacks. https://github.com/usememos/memos/pull/2865/files#diff-69661a91c7ad27e32735ba8890bfe6547e961e462b5ecf7fc7b11f3329701c79R52

Not always, as there are a few differences in wording between British English and American English.

But it's possible to simply leave the json empty and rely on the default `en` fallback provided by the i18next extension, it just works. This also leaves room for updating in wording, if some one think it's worth it.

The `fallbacks` object doesn't work in this case. The language must be listed in `availableLocales` to allow user selection.

Issue #2885, opened 2024-02-01
Date reversed in the timeline in the resource page


ertuil:
### Describe the bug

The resource page's timeline seems to have the wrong order: 2024.1 -> 2024.2 -> 2023.12

![屏幕截图 2024-02-01 161611](https://github.com/usememos/memos/assets/14366322/98639ca9-bc07-4412-b527-9b4c29d1c2bd)

### Steps to reproduce

1. open the resource page

### The version of Memos you're using

0.19.0

Issue #2623, opened 2023-12-15
Webhook not triggered 


lyfar:
### Describe the bug

when I create a new entry from telegram, web hook is not triggered 

### Steps to reproduce

1. post in telegram
2. post is created in memo
3. web hook doesn't work

IDDQD69:
Just noticed this too.

PR #2886, opened 2024-02-01
fix: wrong order of the timeline in the resource page & add webhook when create memos using Telegram bot

ertuil:
What I did:
1. In this PR, I fix the wrong order of the timeline on the resource page (#2885 ), by ordering the resources by its created time. See commit 5c50fc2
2. fix some typos, also in commit 5c50fc2
3. Also, I add a webhook when memos are created by the telegram bot (#2623 ). See commit 5825017


boojack (approved):
LGTM

PR #2902, opened 2024-02-03
feat: filter all untagged memos

### LINKED ISSUE #2879: missing tag filter

Is there a way to filter memos that do not have any tags, so that one can easily find the ones where one forgot to add a tag?

### Type of feature

User Interface (UI)

### Additional context

_No response_

aabiji:
Adding an "Uncategorized" filter allows us to get all memos that aren't tagged. Close #2879.


web/src/components/TagList.tsx
```diff
@@ -83,6 +84,7 @@ const TagList = () => {
         </button>
       </div>
       <div className="flex flex-col justify-start items-start relative w-full h-auto flex-nowrap">
+        <TagItemContainer tag={untagged} tagQuery={filter.ignore} filterIgnore={true} />
```
boojack, line 87: Because the `Uncategorized` is not a tag and won't be used very often, it shouldn't be shown on the top of the tag list. A better way to show this action is to add a dropdown. WDYT?

image

aabiji: Yeah, that's valid. I'll work on it.

Issue #2868, opened 2024-01-30
Warn user if sharing private/workspace memos


NtWriteCode:
### Describe the solution you'd like

It would be great if the memos would notify me if I'm trying to share a private/workspace memos, because to unauthenticated users it'll return error code 403 (obviously).

So, it'd be good to either offer changing visibility to public
or
create a link with a special token to allow seeing that post only regardless of publicity setting.

### Type of feature

User Experience (UX)

### Additional context

I'd prefer the latter solution, since it may let me share private memos with unauthenticated friends and I wouldn't have to make it entirely public.

Betyho:
#2942

Betyho:
Perrilla

PR #2931, opened 2024-02-07
feat: add visibility change dialog

moehanabi:
I am not familiar with ts/react coding, so please make a double check before merge.

Show visibility change dialog when sharing private memos, including workspace memos. In this dialog, user can change the visibility of the memo to public, so that the memo can be viewed by anyone with the link.

related issue:  #2868


boojack:
Let's not make things too complicated. If you want to change the visibility, please edit the memo.

boojack:
Or we can show a visibility selector in memo sharing dialog.

image

moehanabi:
> Or we can show a visibility selector in memo sharing dialog.
> image

I think it is a good idea. I will close the pr and open a new pr for it.

PR #2940, opened 2024-02-08
POC use ent as dbv2

zwpaper:
I have divided the PR into 3 commits to show how ent could be used.

1. c7648670c839539ef94a4af9d15546de34929aa4 define the db schema, memo, memo_relation
2. 2d4ae64791340180034b7d817a54a89e6484daeb ent auto-generate the db client codes
3. ef5b193e682f13970fdcbe114550bbda9bab8816 add a dbv2 to store and use it for get memo and memo relations.


zwpaper:
I mark it as draft, and we can discuss it here.

the PR should actually work for API v2 to get memos.

ent/schema/memorelation.go
boojack, line 1: Maybe `memo_relation`

api/v2/memo_relation_service.go
```diff
@@ -44,23 +46,26 @@ func (s *APIV2Service) SetMemoRelations(ctx context.Context, request *apiv2pb.Se
 
 func (s *APIV2Service) ListMemoRelations(ctx context.Context, request *apiv2pb.ListMemoRelationsRequest) (*apiv2pb.ListMemoRelationsResponse, error) {
 	relationList := []*apiv2pb.MemoRelation{}
-	tempList, err := s.Store.ListMemoRelations(ctx, &store.FindMemoRelation{
-		MemoID: &request.Id,
-	})
+	tempList, err := s.Store.V2.MemoRelation.
```
boojack, line 49: Should we wrap a method in the `store` package?

boojack:
Sorry for the late review. Overall, I think it is a good try!

api/v2/memo_relation_service.go
zwpaper: that's also my first try wrapping the db inside store.

but there are a few concerns, and we can discuss:
1. using the origin store interface has no way to distinguish v1 or v2, we can only replace the origin v1 directly in each function step by step.
2. it seems not worth creating a new store interface only for ent operations
3. the ent client seems to be easy enough to use

PR #2941, opened 2024-02-08
feat: add visibility select in ShareMemoDialog

moehanabi:
I am not familiar with ts/react coding, so please make a double check before merge.

In ShareMemoDialog, user can change the visibility of the memo, so that the memo can be set to public to be viewed by anyone with the link.

related issue: #2868
related pr: #2931

### DIFF
--- a/web/src/components/ShareMemoDialog.tsx
+++ b/web/src/components/ShareMemoDialog.tsx
@@ -1,33 +1,40 @@
-import { Button, IconButton } from "@mui/joy";
+import { Button, IconButton, Select, Option } from "@mui/joy";
 import copy from "copy-to-clipboard";
 import React, { useEffect, useRef } from "react";
 import { toast } from "react-hot-toast";
 import { getDateTimeString } from "@/helpers/datetime";
 import { downloadFileFromUrl } from "@/helpers/utils";
+import useCurrentUser from "@/hooks/useCurrentUser";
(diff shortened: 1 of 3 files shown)


web/src/components/ShareMemoDialog.tsx
```diff
@@ -75,6 +82,22 @@ const ShareMemoDialog: React.FC<Props> = (props: Props) => {
...
+          id: memo.id,
+          visibility: visibility,
+        },
+        ["visibility"],
+      )
+      .then((newMemo: Memo) => {
+        if (newMemo.visibility == visibility) {
+          toast.success(t("common.changed"));
+        }
+      });
```
boojack, line 98: Don't need to use `then` in `async/await` function.

```suggestion
    const updatedMemo = await memoStore
      .updateMemo(
        {
          id: memo.id,
          visibility: visibility,
        },
        ["visibility"],
      )
      .then((newMemo: Memo) => {
        if (newMemo.visibility == visibility) {
          toast.success(t("common.changed"));
        }
      });
```

boojack (approved):
LGTM

web/src/components/ShareMemoDialog.tsx
moehanabi: I force-pushed a new version, please check it.

boojack (approved):
LGTM

PR #2946, opened 2024-02-10
feat: translate statistics

smallnest:
translate statistics to Chinese


smallnest:
closed because we must translate all languages

PR #2947, opened 2024-02-10
chore: allow all 20x response status code in webhook

ImSingee:
Some website may return 201/204 to indicate a more detailed success


boojack (approved):
LGTM

Issue #2901, opened 2024-02-03
Can't input whitespace in create memo relation dialog


vinicius507:
### Describe the bug

When searching for memos in `CreateMemoRelationDialog` the user can't input whitespaces. 

### Steps to reproduce

1. Click on the `Add reference` button in Memo editor
2. Input search string containing whitespace (e.g. `React Server Components`)

### The version of Memos you're using

0.19.1

### Screenshots or additional context

The issue happens because the `Autocomplete` component calls `setSearchText` with `value.trim()`. While useful, the same component uses `searchText` as `inputValue`, which removes any whitespace before the user can enter the next word.

https://github.com/usememos/memos/blob/57510ddee579ab714b2da4dad06b2078e2e5059d/web/src/components/CreateMemoRelationDialog.tsx#L106-L109

A simple fix should be moving the `trim()` call to the `useDebouce` hook, although it would make the hook be called when the input inserts trailing spaces.

PR #2950, opened 2024-02-12
Allowing whitespace in create memo relation dialog

TheMaksym:
Proposing a solution for the issue [regarding not being able to input whitespace in create memo relation dialog](https://github.com/usememos/memos/issues/2901).

gif whitespace working

### DIFF
--- a/web/src/components/CreateMemoRelationDialog.tsx
+++ b/web/src/components/CreateMemoRelationDialog.tsx
@@ -1,7 +1,6 @@
 import { Autocomplete, AutocompleteOption, Button, Checkbox, Chip, IconButton } from "@mui/joy";
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { toast } from "react-hot-toast";
-import useDebounce from "react-use/lib/useDebounce";
 import { memoServiceClient } from "@/grpcweb";
 import { DEFAULT_MEMO_LIMIT } from "@/helpers/consts";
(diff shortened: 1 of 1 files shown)


TheMaksym:
Not exactly sure what error occurred on the front end test. This is in fact my first open source contribution so this is for sure a learning process.

MehadND:
> Not exactly sure what error occurred on the front end test. This is in fact my first open source contribution so this is for sure a learning process.

Hi @TheMaksym, the frontend test failed because you need to have new line at the very end of file after `export default`.

### Steps to resolve this
1. cd `web`
2. run `npm run lint --fix`

### or 

manually enter a new line after `export default ...`

Then push the changes. 

A general advice that I would give to you would be to run `npm run lint` and `npm run lint --fix` before you push any changes. This way any linter errors are easy to catch and fixable.

web/src/components/CreateMemoRelationDialog.tsx
```diff
@@ -106,7 +116,7 @@ const CreateMemoRelationDialog: React.FC<Props> = (props: Props) => {
           inputValue={searchText}
           value={selectedMemos}
           multiple
-          onInputChange={(_, value) => setSearchText(value.trim())}
+          onInputChange={(_, value) => setSearchText(value)}
```
boojack, line 119: Maybe we can only remove `.trim()` that will fulfill your needs. The other changes don't seem too important.

Issue #2575, opened 2023-12-05
Add Delete, Archive and Pin button to memo detail pages


gitkeniwo:
### Is your feature request related to a problem?

When you click a marked / linked memo from another memo, you are navigated to the detail page of that memo (`your.site/m/memo_id`) and if you would like to delete / pin this memo, you'll have to go back and search for it again.

### Describe the solution you'd like

Add Delete, Archive and Pin button to memo detail pages

MehadND:
@boojack are you accepting PR for this issue or for any issue in milestone? Or are these just for your self?

PR #2959, opened 2024-02-15
chore: added delete, archive and pin buttons to memo detail page

MehadND:
This PR aims to resolve the following [issue](https://github.com/usememos/memos/issues/2575).


boojack:
Updated with https://github.com/usememos/memos/commit/d22b7722328afc202e835bb3856e38e2707a73b9

Since we already have `MemoActionMeny`, we can reuse the component.

PR #2966, opened 2024-02-16
chore: update ja.json

Somme4096:
## Updated Japanese translation
- Rearranged `ja.json` order based on latest `en.json`. 
- Added missing lines (And their translations, of course) .
- Fix grammatical inconsistencies between sentences in similar context.
- Added spaces before and after English words that are in Japanese sentences.

*Before*
PRbefore
*After*
PRafter


boojack (approved):

PR #2998, opened 2024-02-25
fix: Incorrectly checking whether registration is allowed in the login v1 api

Cufoon:
As mentioned, this made it impossible to log in via apiv1, simply fixed it.


boojack (approved):
LGTM, thanks for your contribution

PR #3003, opened 2024-02-26
fix: telegram callback query handler dereferencing nil pointer if memo not found

ericd23:
If a memos added through telegram is deleted and we try to change the visibility in telegram, server will dereference a nil pointer in the handler when dispatching webhook.


boojack (approved):
LGTM

Issue #3026, opened 2024-03-02
Pressing down "Tab" collides with IME when editing.


jjaychen1e:
### Describe the bug

https://github.com/usememos/memos/blob/9a8a1d017e06afe289362fd3b3515abcb5a522fa/web/src/components/MemoEditor/index.tsx#L134-L145

Pressing down "Tab" will force IME to stop working and insert temporary characters into the editor.

<img width="147" alt="image" src="https://github.com/usememos/memos/assets/31304335/c91501e3-4b6e-413b-86a0-544ef1bd93c1">

### Steps to reproduce

1. Focus on an editor area
2. Use IME such as Chinese Pinyin
3. Type some characters then press down "Tab" Key

### The version of Memos you're using

v0.20.0

PR #3027, opened 2024-03-02
fix: ignore "Tab" key down event when is composing in editor(#3026)

jjaychen1e:
Fix https://github.com/usememos/memos/issues/3026


boojack (approved):
LGTM

PR #3032, opened 2024-03-03
feat: improve theming

Dubzer:
This PR adds a color-scheme property so that the browser applies the appropriate user agent styles, such as scrollbar color
Learn more about this property: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

Also, I've decided to make an experimental change to the scrollbar color in the dark mode, so it will use the corrent dark shades

Here's an example of improvement in scrollbar style for dark themes on Windows in MS Edge:

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/web/src/css/global.css
+++ b/web/src/css/global.css
@@ -5,3 +5,10 @@ body {
 #root {
   @apply relative w-full min-h-full;
 }
+html.dark {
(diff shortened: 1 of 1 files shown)


boojack (approved):
LGTM

Issue #3036, opened 2024-03-04
PUBLIC memo disallowed, but can be created in Telegram


redsubmarine:
### Describe the bug

Even though you have `Disable public memos` turned on with the host account, public memos can still be created successfully from your Telegram bot.

### Steps to reproduce

1. Connect the Telegram bot.
2. Enable the "disable public memos" option with the host account.
3. Create a memo via the Telegram bot.
4. Change the visibility through the "PUBLIC" button on the Telegram bot.

### The version of Memos you're using

v0.20.0

PR #3037, opened 2024-03-04
fix: check disallow public memo in Telegram(#3036)

redsubmarine:
I have retrieved the 'workspaceSetting' and compared it, then handled to expose the error message. Please review the code.

#3036


boojack (approved):
LGTM

PR #3043, opened 2024-03-05
feat: implemented link preview cards for URLs

MehadND:


ctoli:
good, but I personally think that url preview finish when save memos is better it Helps to improve  the page loading speed

MehadND:
> good, but I personally think that url preview finish when save memos is better it Helps to improve  the page loading speed

So you mean when I create a memo then if there is a url so I should do the preview then like fetch and display the data and then save/cache it?

MehadND:
@ctoli Also I am working on making some changes so that the preview card is only visible on link hover, that way the page loading speed is not that affected I think

onnttf:
I don't think early caching is necessary. I think the preview should reflect the current situation rather than a certain moment in the past.

MehadND:
> I think the preview should reflect the current situation rather than a certain moment in the past.

Sorry I don't really understand what you mean by this

PR #3122, opened 2024-03-23
feat: add shortcut to edit the previous memo

Dubzer:
hi!
sometimes you save a memo, and only after realize you've made a typo.
I've added the ability to edit a previous memo simply by pressing the arrow down key in the empty editor, similar to the arrow up shortcut in Telegram.


web/src/components/MemoEditor/index.tsx
```diff
@@ -161,6 +162,12 @@ const MemoEditor = (props: Props) => {
       }
       return;
     }
+
+    if (!!props.onEditPrevious && event.key == "ArrowDown" && !state.isComposing && editorRef.current.getContent() === "") {
```
boojack, line 166:
```suggestion
    if (!!props.onEditPrevious && event.key === "ArrowDown" && !state.isComposing && editorRef.current.getContent() === "") {
```

boojack (approved):
LGTM

Issue #3093, opened 2024-03-13
Can't load images when sharing notes


wzc90:
### Describe the bug

当使用图片分享笔记时，如果笔记中含有`![]()`这样的图片链接，这个图片无法在分享的图片中加载。

### Steps to reproduce

1. 编写一条笔记，含有`![]()`图片链接，这时图片可见
2. 点击分享笔记，选择图片，此时图片仍然可见，点击图片下载
3. 打开下载的图片，发现此图片无法加载

### The version of Memos you're using

0.20.1

YohannKovacs:
From what i have observed, this happens when the servers serving the requested image have CORS disabled.
When the CORS headers are present on the requested resource, there are no issues.
This might be a problem since most images are hosted on CDN's, which do not set the CORS headers.

YohannKovacs:
@boojack I would like to contribute, but would require some help since I'm not sure how to resolve this. Going through the code, it seems like there is a problem with the sanitization of the image content. 
I've tested without the Sanitization code and found that it works fine. Since we are using a third party sanitizer, i think the problem would be resolved if we switch over to something else or implement our own.

Image with sanitization code
![memos-3_27_2024, 10_44_08 PM](https://github.com/usememos/memos/assets/72270440/44cbca80-f13e-4b6d-8fbe-90b8b49b17df)


Image without sanitization code
![memos-3_27_2024, 10_43_18 PM](https://github.com/usememos/memos/assets/72270440/a7859155-9931-4674-84be-a6e5b7fa823b)

YohannKovacs:
I'd like to add that this issue only occurs if the image is being fetched from CDN's which haven't enabled CORS , since we do sanitization after running into CORS policy error

wzc90:
@YohannKovacs @boojack I found that the bug still exists now when using Safari on MacBook,  but on other browsers the bug is fixed!

YohannKovacs:
@wzc90 have you tried with other image uri's as well? I just tested on an older version of safari (15.6), and it works there.

wzc90:
@YohannKovacs That's the problem.

YohannKovacs:
@wzc90 Can you check the live demo as well? Surprisingly ,I tested on the live demo and found it to be working.

wzc90:
@YohannKovacs Just checked the live demo.It doesn't work for my Safari , the downloaded image was transparent too.
[...]

PR #3147, opened 2024-03-28
removed sanitization logic

YohannKovacs:
Solving issue https://github.com/usememos/memos/issues/3093
I have removed the sanitization code currently being used for images.
Current logic uses an external html sanitization library i.e. bluemonday, however since we are passing the binary image content and not html/markdown, the process damages the image bytes leading to a malformed image. This library filters out data that are not the standard html tags, which would obviously work against raw image bytes. 
I think sanitizing binary image data is unnecessary in that sense.


boojack (approved):
LGTM

PR #3165, opened 2024-03-30
feat: handle restore memo when is in archived memo detail page

ntheanh201:
Currently, when going to the link of an archived memo, the action still shows Archive instead of Restore


boojack (approved):
LGTM

PR #3172, opened 2024-04-01
cmd: use the built-in slices library

stayweek:
In the current go 1.21 version used in the project, slices are no longer an experimental feature and have entered the standard library


michaelien (approved):

PR #3180, opened 2024-04-05
chore: add short uuid to avoid overwritten same files

### LINKED ISSUE #3177: The image with the same name will be overwritten
### Describe the bug

Uploading a file with the same name in the resources will cause the previous file to be overwritten. I'm using Cloudflare R2 as a backend here, and it's not clear at this time whether the other way would result in an override.

The usual practice in this place should be to upload the image and then change the name of the image to its SHA value, timestamp or some other random string to avoid being overwritten by the upload.

### Steps to reproduce

1. Use Cloudflare R2 Storage as a storage service.
2. Prepare an image named 1.png
3. Upload that image to the repository.
4. In preparing a completely different image, also named 1.png
5. Upload that image to the repository.
6. Observe that the previously uploaded 1.png was overwritten.

Version: v0.20.1

image

ntheanh201:
Closes: #3177

### DIFF
--- a/server/route/api/v1/resource.go
+++ b/server/route/api/v1/resource.go
@@ -223,7 +223,7 @@ func (s *APIV1Service) UploadResource(c echo.Context) error {
 	create := &store.Resource{
 		UID:       shortuuid.New(),
 		CreatorID: userID,
-		Filename:  file.Filename,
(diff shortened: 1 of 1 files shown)


boojack:
Refer to https://github.com/usememos/memos/issues/3177#issuecomment-2041505312

PR #3186, opened 2024-04-07
Set navigation icons to not shrink

imikod:
There is a problem with spanish, portuguese and other languages where even when the words are truncated the icons are shrinking.


boojack (approved):

michaelien (approved):
LGTM

PR #3224, opened 2024-04-13
chore: replace util.Min helper with built-in min

Juneezee:
We can use the built-in `min` function since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max


boojack (approved):

PR #3253, opened 2024-04-20
fix: specify go toolchain in go.mod (fix #3252)

### LINKED ISSUE #3252: Go 1.22 toolchain is not available
### Describe the bug

I want to run memos locally, but when running `go` run or any `go` command, it says `go` can't find the go1.22 toolchain. The root cause of this is the go directive in the `go.mod` file acts as an input to select which go toolchain version will be used for this project. Because there's no go toolchain directive specified in `go.mod`, go select go1.22 as the toolchain version, and go can't find it because it is an invalid go toolchain version (it should be in the form of `go1.N.P`).

The solution for this is we can specify the go toolchain with toolchain directive in `go.mod`, or just specify the minimum of go release version with `1.N.P` syntax.

### Steps to reproduce

1. Download `go` with the version is below `1.22`
2. Run any `go` command inside the project with `go` below of `1.22`. It will start downloading the `go1.22` toolchain (I'm using mac apple silicon).

master (at the time I opened this issue, the latest version is v0.21.0)

image

kbiits:
Closes #3252

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/usememos/memos
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	github.com/aws/aws-sdk-go-v2 v1.26.1
 	github.com/aws/aws-sdk-go-v2/config v1.27.11


michaelien (approved):
LGTM

PR #3311, opened 2024-05-05
fix(tag_deletion): fix deleting tag from tag lists

### LINKED ISSUE #3309: Delete tag not working
### Describe the bug

Tag deletion seems not to be working. If the bug is verified, I'm happy to contribute to fixing this.

### Steps to reproduce

1. Go to the tag list
2. Hover on # symbol
3. Click and delete
4. Confirm deletion

The tag should be deleted if the user confirms to delete it, and the frontend should auto-refresh the tag list after deletion

latest master (#da535c8)

https://github.com/usememos/memos/assets/67781184/579b9682-d265-49aa-b0db-5e50f07e9d5c

kbiits:
Fix #3309 

What to know:
- Delete tag with or without children should be working after this fix, but the tag is still present in the memo
- the tag deletion still not working if the tag that wants to be deleted is a parent tag (need to discuss how is the expected behaviour if we delete the parent tag)
ex: tag is `#parent/sub1`. The tag is not deleted if the user hovers on #parent and deletes it. But if the user deletes #parent/sub1, it's working as expected. `#parent/sub1` will be deleted with `#parent` still exists.

https://github.com/usememos/memos/assets/67781184/e11432d3-2fba-480a-9e0c-e2c470c0a342

### DIFF
--- a/server/router/api/v1/tag_service.go
+++ b/server/router/api/v1/tag_service.go
@@ -137,13 +137,7 @@ func (s *APIV1Service) RenameTag(ctx context.Context, request *v1pb.RenameTagReq
 }
 
 func (s *APIV1Service) DeleteTag(ctx context.Context, request *v1pb.DeleteTagRequest) (*emptypb.Empty, error) {
-	userID, err := ExtractUserIDFromName(request.Tag.Creator)
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "invalid user name: %v", err)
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM

boojack:
And I have a todo for your question, see more in https://github.com/orgs/usememos/projects/5/views/1?pane=issue&itemId=61918870

PR #3330, opened 2024-05-07
Fix: Set max height to 50vh for better viewing for larger screens.

jackwander:
### Description
Setting max height to 50vh than limiting to 256px only for better editing in larger screens.

image


boojack (approved):

Issue #3348, opened 2024-05-11
ERROR failed to migrate database after I updated memos


wayner6:
### Describe the bug

Today I updated memos, but the web cnannot load, so I run `docker-compose up` to check something. 

```
memeos is up-to-date
Attaching to memeos
memeos   | ---
memeos   | Server profile
memeos   | version: 0.22.0
memeos   | data: /var/opt/memos
memeos   | dsn: /var/opt/memos/memos_prod.db
memeos   | addr: 
memeos   | port: 5230
memeos   | mode: prod
memeos   | driver: sqlite
memeos   | ---
memeos   | succeed to copy a backup database file
memeos   | start migrate
memeos   | applying migration for 0.22.0
[...]

PR #3350, opened 2024-05-12
Fix MySQL database migration 0.21.0 -> 0.22.0

leo9800:
This pull request addresses #3348, which implements the fix I worked out, mentioned in the issue.

SQLite migrations and PostgreSQL migrations seems to be fine, as they are not trying `DROP`ping non-exist columns.

Besides, a version bump (e.g. 0.22.0 -> 0.22.1) is suggested to aware the downstream users to upgrade.

### DIFF
--- a/store/db/mysql/migration/prod/0.22/01__memo_tags.sql
+++ b/store/db/mysql/migration/prod/0.22/01__memo_tags.sql
@@ -1,4 +1,3 @@
 ALTER TABLE `memo` ADD COLUMN `tags_temp` JSON;
 UPDATE `memo` SET `tags_temp` = '[]';
-ALTER TABLE `memo` DROP COLUMN IF EXISTS `tags`;
 ALTER TABLE `memo` CHANGE COLUMN `tags_temp` `tags` JSON NOT NULL;
--- a/store/db/mysql/migration/prod/0.22/01__memo_tags.sql
+++ b/store/db/mysql/migration/prod/0.22/01__memo_tags.sql
@@ -1,4 +1,4 @@
 ALTER TABLE `memo` ADD COLUMN `tags_temp` JSON;
 UPDATE `memo` SET `tags_temp` = '[]';
-ALTER TABLE `memo` DROP COLUMN `tags`;
+ALTER TABLE `memo` DROP COLUMN IF EXISTS `tags`;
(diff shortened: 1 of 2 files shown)


boojack (approved):

wzc90:
In MySQL versions prior to 8.0.16, the ALTER TABLE ... DROP COLUMN IF EXISTS syntax is not supported. However, starting from MySQL version 8.0.16 and onward, the IF EXISTS clause is supported to conditionally drop a column only if it exists, which prevents errors from occurring when the specified column does not exist.

So, if you are using MySQL 8.0.16 or later, the statement:

```ALTER TABLE `memo` DROP COLUMN IF EXISTS `tags`;```

will work as expected, and it will drop the tags column if it exists without causing an error. If you are using an older version of MySQL, this statement would result in an error, and you would need to use a different approach to check for the existence of the column before attempting to drop it.

So the best approach is to directly remove this line of code. @boojack @leo9800

File: store/db/mysql/migration/prod/0.22/01__memo_tags.sql [hunk 1]
```diff
@@ -1,4 +1,4 @@
 ALTER TABLE `memo` ADD COLUMN `tags_temp` JSON;
 UPDATE `memo` SET `tags_temp` = '[]';
-ALTER TABLE `memo` DROP COLUMN `tags`;
+ALTER TABLE `memo` DROP COLUMN IF EXISTS `tags`;
```
boojack, line 3: Let's remove this line as it shouldn't exist.

boojack, line 3:
```suggestion
```

store/db/mysql/migration/prod/0.22/02__memo_payload.sql
```diff
@@ -1,4 +1,4 @@
 ALTER TABLE `memo` ADD COLUMN `payload_temp` JSON;
 UPDATE `memo` SET `payload_temp` = '{}';
-ALTER TABLE `memo` DROP COLUMN `payload`;
+ALTER TABLE `memo` DROP COLUMN IF EXISTS `payload`;
```
boojack, line 3:
```suggestion
```

boojack (changes requested):

PR #3355, opened 2024-05-12
Add scroll restoration when navigates back

### LINKED ISSUE #3277: Remember memos list browsing location

I now have more than 500 memos. When I enter the details page of a memo and then return to the homepage, it will go back to the top and I have to scroll down again, which is very troublesome. I hope that when returning to the homepage from the details page, the browser location can be remembered instead of starting from the beginning. This will enhance the user experience, especially when there are many memos records.

### Type of feature

User Experience (UX)

### Additional context

_No response_

kbiits:
This PR fixes #3277 issue.

I refactor the Home component to make the react-router-dom ScrollRestoration work. There are some useless re-rendering that makes the ScrollRestoration not work.

Here's the demo 
https://www.veed.io/view/d774a53a-fcd9-4c9f-afe2-d145e204dcc2?panel=share

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,7 +1,7 @@
 import { useColorScheme } from "@mui/joy";
 import { useEffect } from "react";
 import { useTranslation } from "react-i18next";
-import { Outlet } from "react-router-dom";
+import { Outlet, ScrollRestoration } from "react-router-dom";
 import useLocalStorage from "react-use/lib/useLocalStorage";
 import { getSystemColorScheme } from "./helpers/utils";
 import useNavigateTo from "./hooks/useNavigateTo";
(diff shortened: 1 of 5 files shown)


kbiits:
Sorry, I've just noticed there's a bug. When we navigate back from memo detail and the Home page already has more than 1 page of memos in the store (I mean the zustand store), the next page token is reset. 

We need to move it out to the zustand store so it doesn't reset when navigating back.

kbiits:
I encountered many problems because we reused the same store on multiple pages (the homepage, the explore page, and others).

I think we can do a workaround. By using full page reload when navigating pages between home, explore, timeline, and other pages that show memo lists and use the same store, we can prevent the bug and enable the scroll restoration feature.

PR #3399, opened 2024-05-19
chore: update i18n from Weblate

michaelien:


quaintdev:
@michaelien I added full translations for Marathi language yesterday. This PR has only 22% of them which I had the day earlier. Also the branch weblate-i18n is deleted so where are the full translations?

Issue #3443, opened 2024-05-23
No data after windows or docker restart


kyxap:
### Describe the bug

Hi all, after restart of windows 11 or docker there no saved data. 
But user, and api tokens are still there. 

### Steps to reproduce

1. `docker run -d --name memos -p 5230:5230 -v ~/.memos/:/var/opt/memos neosmemo/memos:stable`
2. create note
3. restart docker/windows
4. check if created note is still there 

### The version of Memos you're using

Version: v0.22.0

### Screenshots or additional context

![image](https://github.com/usememos/memos/assets/3080529/dbb09a1b-c244-4141-bed3-da1d4552f11f)

I was expected that `~/.memos/` will not work on windows (cmd/ps) but decided to try and it worked (no error) and was working fine. But after 2w of usage I went into restart and realized that data is disappeared. 

After the quick look I've found that you actually can use `~/` in `PS` v5 but not `cmd` in windows.
But docker does not create `memos` in user home and if I create it manually it will still not place db files in it. Seems like it will use some sort of in memory db which will be removed after restart. 

Some ideas on the issue:
1) It's make sense to updated a readme for windows users to avoid this issue  (see #3444)
2) I think it make sense to add verification if application is actually using/has access to the folder and inform user if it does not (UI error will be nice). And maybe disable in memory db to avoid such issue? 

I'm using koreader plugin for saving notes from ereader and for some reason after notes are disappeared I can't upload them back again. It seems like there some other way to prevent adding notes multiple times to prevent duplicates? Is there quick way to reset it or I will be better with creating new user?

Thanks

boojack:
README updates: https://github.com/usememos/memos/commit/d39cacf1f94e2f8f34877797e430fb387e7eac0b

PR #3444, opened 2024-05-23
Windows cmd and ps command 

kyxap:
Also added auto start option

#3443


boojack:
The README should be kept as simple and easy as possible, we'd like to put more detailed information into the documentation. 

And I'll add a hint in this section, as well as update the documentation. However, thanks for the contribution!

PR #3459, opened 2024-05-26
feat: add panic recovery middleware for grpc and echo server

RyoJerryYu:
A panic would lead to a server shutdown now.
This PR is trying to make the server more robust, by adding panic recover interceptor for echo server and grpc server.


boojack (approved):
LGTM

PR #3491, opened 2024-05-30
Added logic to delete memo relation and memo comments recursively whe…

mmacdo54:
…n a memo is deleted

- Added logic that deletes the memo relation if exists
- Added logic that recursively calls the DeleteMemo function for all the Memo comments

### DIFF
--- a/server/router/api/v1/memo_service.go
+++ b/server/router/api/v1/memo_service.go
@@ -352,6 +352,11 @@ func (s *APIV1Service) DeleteMemo(ctx context.Context, request *v1pb.DeleteMemoR
 		return nil, status.Errorf(codes.Internal, "failed to delete memo")
 	}
 
+	// Delete memo relation
+	if err := s.Store.DeleteMemoRelation(ctx, &store.DeleteMemoRelation{MemoID: &id}); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to delete memo relations")
+	}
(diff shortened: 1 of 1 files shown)


server/router/api/v1/memo_service.go
```diff
@@ -363,6 +368,17 @@ func (s *APIV1Service) DeleteMemo(ctx context.Context, request *v1pb.DeleteMemoR
 		}
 	}
 
+	// Delete memo comments
+	relations, err := s.Store.ListMemoRelations(ctx, &store.FindMemoRelation{RelatedMemoID: &id})
```
boojack, line 372: `Type: MemoRelationComment`

```suggestion
	relations, err := s.Store.ListMemoRelations(ctx, &store.FindMemoRelation{RelatedMemoID: &id})
```

mmacdo54: Not sure what you are asking here? The suggested change looks identical to the existing code?

boojack: Your comment removes the memo comment, but there is no relation type specified here?

mmacdo54: Ah I see what you mean. Sorted so it targets the comments - also added in logic to delete the any references to this deleted Memo

boojack (approved):
LGTM

PR #3515, opened 2024-06-04
chore: update fetch tags args

mmacdo54:
…ponents to pass in those params needed

- Updated the useTagStore fetchTags function to now accept a user and location and it handles the filtering internally
- updated all the places fetchTags was called to pass through the user and location where applicable

Looks to solve the issue where in certain circumstances it finds any public tags

### DIFF
--- a/web/src/store/v1/tag.ts
+++ b/web/src/store/v1/tag.ts
@@ -23,17 +23,17 @@ export const useTagStore = create(
         .sort((a, b) => b[1] - a[1])
         .map(([tag]) => tag);
     },
-    fetchTags: async (options: { skipCache?: boolean; user?: User; location?: Location<any> }) => {
+    fetchTags: async (params: { user?: User; location?: Location<any> }, options?: { skipCache?: boolean; }) => {
       const { tagAmounts: cache } = get();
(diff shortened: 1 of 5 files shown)


web/src/store/v1/tag.ts
```diff
@@ -20,12 +23,21 @@ export const useTagStore = create(
         .sort((a, b) => b[1] - a[1])
         .map(([tag]) => tag);
     },
-    fetchTags: async (filter?: string, options?: { skipCache: boolean }) => {
+    fetchTags: async (options: { skipCache?: boolean; user?: User; location?: Location<any> }) => {
```
boojack, line 26: It's better to split it into two args.

```suggestion
    fetchTags: async (params?: {user?: User; location?: Location<any> }, options: { skipCache?: boolean }) => {
```

boojack (changes requested):

web/src/store/v1/tag.ts
mmacdo54: Yeah good shout.

boojack (approved):
LGTM

PR #3543, opened 2024-06-09
feat: return not found error instead internal error for local file not exist

RyoJerryYu:
Sometimes there may be some resource store on file system disappeared, as the user here described https://github.com/RyoJerryYu/obsidian-memos-sync/issues/14 .  (Though I don't know how this happened)

Returning a gRPC NotFound error would be better than returning Internal error. Then we could got a 404 status code when we use gRPC-web or gRPC-gateway API if this was happened.

### DIFF
--- a/server/router/api/v1/resource_service.go
+++ b/server/router/api/v1/resource_service.go
@@ -197,6 +197,9 @@ func (s *APIV1Service) GetResourceBinary(ctx context.Context, request *v1pb.GetR
 		}
 
 		file, err := os.Open(resourcePath)
+		if os.IsNotExist(err) {
+			return nil, status.Errorf(codes.NotFound, "file not found for resource: %s", request.Name)
+		}
 		if err != nil {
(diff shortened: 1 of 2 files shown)


server/router/api/v1/resource_service.go
```diff
@@ -197,6 +197,9 @@ func (s *APIV1Service) GetResourceBinary(ctx context.Context, request *v1pb.GetR
 		}
 
 		file, err := os.Open(resourcePath)
+		if os.IsNotExist(err) {
+			return nil, status.Errorf(codes.NotFound, "file not found for resource: %s", request.Name)
+		}
```
boojack, line 202: Move into `if err != nil {`

boojack (approved):
LGTM, otherwise

server/router/api/v1/resource_service.go
RyoJerryYu: done

PR #3589, opened 2024-06-18
feat: add public flag instead of system setting to enable signup

johnnyjoy101:
Because disabling registration as well as disabling password login may lead to some unforeseen problems: users can't sign in and need to modify the database. So it is recommended to put it into an environment variable.


boojack (approved):
LGTM

evanlihou:
Hey there! Definitely agree with the change for disabling signups to be an env variable, but this PR removes the ability for an instance to allow only SSO logins. Is that no longer a supported use case?

Also, the new flag is not documented

evanlihou:
@johnnyjoy101 Nah I don't see it as a security flaw, it's more a matter of providing customizability to instance administrators and reducing confusion for the users.

For example, a small business that uses memos to share snippets internally. All users log in via SSO, maybe once every few years an admin needs to log in with a password because something broke in SSO. The users not needing to think to themselves "Should I type in my Active Directory username and password here, or should I click this 'Sign in with...' button?" It's a small thing for sure, but it is a regression from the previous version.

Not trying to throw work at you btw! I'm happy to open a PR myself, I just wouldn't want to step on any toes. Also I wanted to give @boojack a few days to chime in and disagree with me

PR #3673, opened 2024-07-09
Let Members edit WorkSpace's memos

Iziram:
# Collaborative Workspace Memos

## Description

I use `memos` as a replacement for Google Keep and I'm very happy with it. Recently, however, I wanted to share a shopping list with my partner.
Using `workspace` visibility allowed my partner to see my list, but she couldn't edit it.
I couldn't find a way to set this up, so I decided to code it.

For the moment it's a simple version that allows all `workspace` members to edit `memos` published in the `workspace`.

In the future, I'm thinking of adding a collaboration feature. We could then add members to our `memos` to enable them to modify it.

I'm making this pull request in case this particular need is requested by others. 

## What does it do

This `pull request` adds a new parameter to the workspace memo relative settings. This parameter, once active, allows workspace members to modify any `memo` published on the workspace. This makes it easy to create collaborative notes.

settings-memo

## Code

I'm not very familiar with React, so it's possible that my modifications aren't the most optimized. Feel free to modify/optimize my code.

I've run the server-side unit tests and everything passes.
On the web side, I linted and tested by hand.

For translations, I've written the description for English and French (as these are the only two languages I know).


boojack:
Hi @Iziram. It seems that the new setting name `enable_members_edit` is not very suitable. I have a bit of a naming compulsion, so I suggest not adding this setting for now. Instead, we could allow admin users to edit other people's data through the UI. WDYT

Iziram:
> Hi @Iziram. It seems that the new setting name `enable_members_edit` is not very suitable. I have a bit of a naming compulsion, so I suggest not adding this setting for now. Instead, we could allow admin users to edit other people's data through the UI. WDYT

Well, if you have any idea for the setting's name feel free to say. I will happily change my code.
If you need i've allowed edits by maintainers.

Your solution might be suitable for moderation purposes, but not really for my problem, as most of the members of my workspace are not administrators.

In the end, it may be a good idea to keep both solutions : 

- Administraton can edit any memo
- Members of workspace can edit workspace/protected memo

When I coded this request, my first idea was to make a `collaborator` system where when creating the memo, you could choose to add members to the memo. But i scrapped the idea because I don't really know how the code base works.

But if you're interested, i can write you a doc on how i would have implemented that.

Iziram:
I've joined the discord if needed.

PR #3687, opened 2024-07-14
chore: update disallow public visibility setting name

johnnyjoygh:
`disallow_public_visible` to `disallow_public_visibility`


boojack (approved):
Good catch

PR #3711, opened 2024-07-18
fix: code blocks of unknown languages cause HTML injection

andrigamerita:
A code block of unknown language (that is, a language not treated as special by Memos and not handled by highlight.js) should fall back on rendering its plaintext content. However, the content is never properly escaped before it is appended to the DOM, and thus any string that happens to contain HTML is unsafely rendered. This commit fixes the issue by ensuring that, when none of the previous cases handle the text, any HTML entities are escaped first.

### DIFF
--- a/web/src/components/MemoContent/CodeBlock.tsx
+++ b/web/src/components/MemoContent/CodeBlock.tsx
@@ -42,7 +42,10 @@ const CodeBlock: React.FC<Props> = ({ language, content }: Props) => {
       // Skip error and use default highlighted code.
     }
 
-    return content;
+    // escape any HTML entities when rendering original content
+    return Object.assign(document.createElement('span'), {
+      textContent: content,
+    }).innerHTML;
   }, [formatedLanguage, content]);
 
   const handleCopyButtonClick = useCallback(() => {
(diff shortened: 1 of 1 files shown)


web/src/components/MemoContent/CodeBlock.tsx
```diff
@@ -42,7 +42,10 @@ const CodeBlock: React.FC<Props> = ({ language, content }: Props) => {
       // Skip error and use default highlighted code.
     }
 
-    return content;
+    // escape any HTML entities when rendering original content
+    return Object.assign(document.createElement("span"), {
```
boojack, line 46: How about div

boojack (approved):

web/src/components/MemoContent/CodeBlock.tsx
andrigamerita: Because of div being a block element and span an inline one? This here will set just the text content, not the HTML, to then extract an HTML content that will only have a single text node, so everything is technically within standard and working in practice; and in this sense, since we're only writing text inside the element, I would argue span is more correct, since it's a generic element designed specifically for text. Although a div would work the exact same way.
Would you like this changed anyway?

PR #3717, opened 2024-07-19
improve: s3 store prefix url

### LINKED ISSUE #3575: current version of S3 Storage cannot set url domain prefix

In previous versions it was possible to set a prefix, but when the Endpoint domain was different from the CDN domain, it was not possible to set a prefix which prevented the use of the CDN domain.

In fact, CDN image addresses only need to be spliced with `URL Prefix` and `Filepath template`, and do not need to using pre-signed url.

### Type of feature

Integrations

### Additional context

If can't using CDN, some images are too large, take longer to load.

kuusei:
Close #3575


proto/store/workspace_setting.proto
```diff
@@ -75,6 +75,7 @@ message StorageS3Config {
   string endpoint = 3;
   string region = 4;
   string bucket = 5;
+  string prefix_url = 6;
```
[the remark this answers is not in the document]
kuusei: > > Presigned URLs can only be used with the .r2.cloudflarestorage.com S3 API domain and cannot be used with custom domains. Instead, you can use the [general purpose HMAC validation feature of the WAF](https://developers.cloudflare.com/ruleset-engine/rules-language/functions/#hmac-validation), which requires a Pro plan or above.预签名 URL 只能用于 .r2.cloudflarestorage.com S3 API 域，不能用于自定义域。相反，您可以使用 WAF 的通用 HMAC 验证功能，该功能需要 Pro 计划或更高版本。
> 
> Maybe presigned urls cannot fit the custom domain in cloudflare r2. Did you test this locally?也许预签名的 url 无法适应 cloudflare r2 中的自定义域。你在本地测试过吗？
> 
> https://developers.cloudflare.com/r2/api/s3/presigned-urls/#limitations

Yes, for R2, presigned URLs cannot be used with custom domains. Therefore, when setting the prefixURL, the file path is composed of prefixURL/Filepath without using the presigned feature. 
I have tested it on both local and VPS deployed server environments, and it can display images correctly in both cases.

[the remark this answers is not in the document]
kuusei: I have updated my pr, `custom_domain` is a great name, and we can use `custom_domain` to represent `prefix_url` in r2 as well.

server/router/api/v1/resource_service.go
```diff
@@ -355,7 +355,7 @@ func SaveResourceBlob(ctx context.Context, s *store.Store, create *store.Resourc
 		if err != nil {
 			return errors.Wrap(err, "Failed to upload via s3 client")
 		}
-		presignURL, err := s3Client.PresignGetObject(ctx, key)
+		presignURL, err := s3Client.PresignGetObject(ctx, key, s3Config)
```
boojack, line 358: We should check custom domain here instead of inside `s3Client.PresignGetObject`.

boojack (approved):
LG otherwise

server/router/api/v1/resource_service.go
kuusei: There are two options: 1) Change the function name to s3Client.GetObject, 2) Check the custom domain here. I updated the code to use option 2, but maybe option 1 is better

server/service/s3_object_presigner/s3_object_presigner.go
```diff
@@ -67,15 +68,21 @@ func (p *S3ObjectPresigner) CheckAndPresign(ctx context.Context) {
 			continue
 		}
 
-		presignURL, err := s3Client.PresignGetObject(ctx, s3ObjectPayload.Key)
-		if err != nil {
-			return
+		var referenceURL string
```
boojack, line 71: Could we skip this? This is the scheduler that refreshes the presigned time. For resources that use custom domain, we don't need to reference its URL.

kuusei: I kept this because when I set/unset custom domains, this scheduler will automatically update and switch to using the custom domain URL/presigned URL after the previous signature expires.

Euruson:
Could someone share why this hasn’t been merged? From a user perspective, custom domain/CDN support for S3 resources would greatly improve performance and access speed, and is quite important in real use cases.

PR #3739, opened 2024-07-24
chore: build release containers from tags rather than branches

### LINKED ISSUE #3645: Premature Pushing of Stable Docker Tags
### Describe the bug

I host all of my infrastructure in a GitOps repository which uses Renovate to automate Docker image updates. I've noticed Memos that frequently triggers update PRs before releases are officially available. For instance, I've had a [PR updating to Memos v0.22.3](https://github.com/gabe565/home-ops/pull/710) since 2024-06-21, even though v0.22.2 is the latest release. It appears that the `release/0.22.3` branch is being pushed, which is causing this issue.

I would expect stable Docker tags to only be created when a new release is officially available. Pushing release tags early can cause confusion and potentially unstable updates in automated systems, as users may receive updates that are not officially released.

### Steps to reproduce

Any hosting strategy that automates updates will have this issue. Renovate makes it obvious in my case, but there are other times this could happen more silently. For example, if a user uses Watchtower, they will receive the update much earlier than they should without warning.

v0.22.2

Here is a screenshot of the latest package tags:
image
I would expect `latest` to match `stable` and `0.22.2`, while `0.22.3` should not exist or should have a suffix indicating it is not yet stable.

nikdoof:
Fixes #3645, currently the release containers are build from the release branches, which are created in preparation for a release. This causes some users with automated CI/CD to either deploy a pre-release version or be notified of a release version before it has actually been released. 

This change switches the workflow to be triggered by tags, which are created on the release version.

### DIFF
--- a/.github/workflows/build-and-push-release-image.yml
+++ b/.github/workflows/build-and-push-release-image.yml
@@ -2,9 +2,9 @@ name: build-and-push-release-image
 
 on:
   push:
-    branches:
-      # Run on pushing branches like `release/1.0.0`
-      - "release/*.*.*"
+    tags:
+      # Run on pushing a tag in the format of 'v0.0.0'
+      - "v[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
   build-and-push-release-image:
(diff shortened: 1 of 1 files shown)


boojack (approved):
LGTM

PR #3750, opened 2024-07-28
fix: dead link in README.md

nalderto:
The previous page no longer exists.  The new one takes users to the Docker directions for Windows.


boojack (approved):
LGTM, good catch

PR #3783, opened 2024-08-04
Allow memo date change Firefox

RoccoSmit:
in v 0.22.4 the ability to update memo dates was reinstated. This feature works in Edge and Chrome but not Firefox

This is due to Firefox needing a specific date string format provided as the input value otherwise the time section of the datetime is not populated.

This becomes an issue as Firefox only fires the `change` event needed to update the date state if the complete date and time is set on the input

This pr is to provide the required date string format of `yyyy-mm-ddThh:mm:ss`

Tested in Firefox and Chromium and works as expected.


RoccoSmit:
Suggested change was to use MUI datetime picker. This will be a new concept to what this PR is doing so closing this PR and will create a new one once MUI datetime picker is implemented

PR #3785, opened 2024-08-06
fix: tag query for mysql

tiltshiftfocus:
In version v0.22.4, filtering timeline via tags when using MySQL as database driver will cause the `ListMemos` API to fail due to an invalid query. 

This pull request removes the `%` in the query and should resolve the issue.


boojack (approved):
LGTM, good catch

Issue #3791, opened 2024-08-07
Filter Timezone Inconsistency


Yamico:
### Describe the bug

In the GMT+8 timezone, the creation time of memos is displayed correctly. However, when applying the displaytime filter, information before 8 AM on the same day appears under the previous day. 

The suspected reason is that the filter is using UTC time. This issue can be reproduced in the latest version. please help with that.

### Steps to reproduce

1. Create a memo with a time earlier than the current timezone's hour.
2. Create a memo with a time later than the current timezone's hour.
3. Click on a date in the activity calendar.

Result: Only memos with a time later than the current timezone's hour are displayed.

### The version of Memos you're using.

latest

Yamico:
![image](https://github.com/user-attachments/assets/ffc53a87-6da3-43bc-b906-997d443a8c94)
![image](https://github.com/user-attachments/assets/46ed22eb-b35b-4876-b0c0-51205304d4d3)
![image](https://github.com/user-attachments/assets/17c82dd9-d297-4f5c-98ed-3327af83db69)

Issue #3760, opened 2024-07-31
Filtering by date is incorrect


hchengting:
### Describe the bug

When filtering memos by date, memos that do not belong to that date will also be displayed.
![image](https://github.com/user-attachments/assets/baadeba3-89b1-4fba-bff1-f6d142901bf5)

### Steps to reproduce

1. Set the device time zone to GMT+8.
2. Create a memo and set the creation time to 12:19 AM.
3. Click calendar to filter by date.

### The version of Memos you're using.

v0.22.4

PR #3794, opened 2024-08-08
fix: corrected the issue where an incorrect timestamp was applied when using the displaytime filter

Yamico:
fix about issue #3791  #3760


RoccoSmit:
I believe this issue is now resolved. See https://github.com/usememos/memos/pull/4046

Yamico:
I'm glad if what I did was helpful.

PR #3844, opened 2024-08-28
perf: reduce bundle size by 21% with direct icon imports

michelheusschen:
Reduce the size of the final bundle by using direct imports instead of including all icons. This decreases the size of `index.js` by 21% (from 3,798.92 kB to 3,014.38 kB).


boojack (approved):
Great!

PR #3856, opened 2024-08-31
fix: thumbnail creation race condition

RoccoSmit:
When uploading resources to a memo, Firefox sends the creation request and then sends the get resourcebinary request 2 times
firefox double network call

This causes a race condition where request 1 creates the thumbnail file and while saving the data to the file, request 2 sees the file exists and tries to loads the data from the incomplete file

e.g. of request (ref 3) returning the image before request (ref 1) has completed saving the file
image

The result is that the UI shows a broken image
Broken image

Refreshing the page loads the fixed image

Changes made:
- Created a temporary random thumbnail file path so that other requests have a very low chance of referencing the thumbnail file being created.
- After processing renamed the random file to the required thumbnail file name for future reference  

e.g. of both requests now following the full flow
flow sync

### DIFF
--- a/server/router/api/v1/resource_service.go
+++ b/server/router/api/v1/resource_service.go
@@ -456,7 +456,7 @@ func (s *APIV1Service) getOrGenerateThumbnail(resource *store.Resource) ([]byte,
 		}
 
 		if err := os.Rename(tempFilePath, filePath); err != nil {
-			return nil, errors.Wrap(err, "failed to rename thumbnail file")
+			return nil, errors.Wrap(err, "failed to rename temp thumbnail file")
 		}
 	}
 
--- a/server/router/api/v1/resource_service.go
+++ b/server/router/api/v1/resource_service.go
@@ -8,6 +8,7 @@ import (
 	"image"
 	"io"
 	"log/slog"
+	"math/rand/v2"
 	"os"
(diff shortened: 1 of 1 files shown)


boojack:
It seems to be working fine on chrome, so why is firefox sending the request twice?

image

RoccoSmit:
I think it has something to do with how Chrome prevents duplicate requests from being sent

e.g. if you add a resource and then try adding the same resource a second time to the same memo you get the following in Chrome vs Firefox

Chrome only allows 1 create and display request so you will see 1 resource displayed
image

Firefox allows both create and display requests to go through so you see 2 resources displayed
image

You can also replicate this by adding a resource to a memo, clicking the `x` and you Firefox will allow you to add the resource back in where Chrome wont.

I'm assuming that if I add a cache buster query string to the request URL it will confirm my suspicions, but I'll look into it some more tonight.

The PR will fix scenarios where the same thumbnail requests come through quickly one after the other

RoccoSmit:
Did some further testing, I cant find why Firefox sends 2 requests. The react code renders the image once only and the server still shows 2 requests coming in.

I'm also not able to replicate the broken images any more. 

Closing this PR and will revisit if this issue shows up again in the future

PR #3876, opened 2024-09-04
test: add test for searching memos by tags

onee-only:
Added test for searching memos by tags, which has been fixed in #3785.


boojack (approved):
LGTM

PR #3889, opened 2024-09-07
chore: replace `unmatchedEmailAndPasswordError` with unmatchedUsernameAndPasswordError`

haohanyang:
When the user fails to login using username and password on `/api/v1/auth/signin`, the error message should be `unmatched username and password`, instead of `unmatched email and password`


boojack (approved):
LGTM, good catch!

PR #3911, opened 2024-09-11
feat: support disable change username and nickname

thezzisu:
When using Single Sign-On (SSO) for a restricted environment (eg. in colleges or institutions), username and nickname should be set to SSO-provided value and users should not be able to edit them.

This PR implements this functionality by adding two workspace settings.

image


server/router/api/v1/user_service.go
```diff
@@ -196,11 +201,17 @@ func (s *APIV1Service) UpdateUser(ctx context.Context, request *v1pb.UpdateUserR
 	}
 	for _, field := range request.UpdateMask.Paths {
 		if field == "username" {
+			if workspaceGeneralSetting.DisallowChangeUsername {
+				continue
```
RoccoSmit, line 205: Can these checks return an error. if no response is returned the use will never know why their value doesnt get saved

RoccoSmit (commented):
Can we add some form of UI visualization showing that the values can't be changed. e.g. showing label or disabled text inputs instead of text input.

proto/api/v1/workspace_setting_service.proto
```diff
@@ -50,6 +50,11 @@ message WorkspaceGeneralSetting {
   // 0: Sunday, 1: Monday, 2: Tuesday, 3: Wednesday, 4: Thursday, 5: Friday, 6: Saturday
   // Default is Sunday.
   int32 week_start_day_offset = 6;
+
+  // disallow_change_username disallows changing username.
+  bool disallow_change_username = 7;
+  // disallow_change_nickname disallows changing nickname.
+  bool disallow_change_nickname = 8;
```
boojack, line 57: Disabling username change is a nice feature, but the nickname part is a bit over-designed. And I haven't seen a platform that restricts changing nicknames. Is there more context behind this?

server/router/api/v1/user_service.go
thezzisu: Thank you! The latest commit will gray out corresponding field, and returns an error in api instead

proto/api/v1/workspace_setting_service.proto
thezzisu: You are right, restricting nickname is always a bad idea. However this option is made for some restrictive environment, like in my university (Peking Univ.), government regulations forces everyone to use their real name as nickname to avoid malformed nicknames.

thezzisu: So I think it's reasonable to leave this options to the user ;D

boojack (approved):

PR #3916, opened 2024-09-11
fix: iframe rendering

ti777777:
By default, DOMPurify removes the content of iframe, but I think iframe is a commonly used tag


boojack (approved):
LGTM

xhocquet:
Thanks for this fix, just ran into the bug and was wondering why I couldn't see the issue on `main`'s code!

PR #3930, opened 2024-09-15
fix: fixed the code block layout in compact mode on iOS/MacOS

durianice:
When using `compact mode`, if the collapsed content contains code blocks, the layout displays abnormally on **iOS browsers (Safari and the latest Chrome) and also macOS Safari**. Therefore, I used the `dotdotdot-js` library to implement content folding functionality, which nicely replaces `-webkit-line-clamp`. And add collapse button in the `compact mode` when show more.

eg:
MacOS Safari
Snipaste_2024-09-16_03-04-41

iOS Chrome120

### DIFF
--- a/web/src/utils/ellipsis.ts
+++ b/web/src/utils/ellipsis.ts
@@ -1,0 +1,656 @@
+/* eslint-disable */
+// @ts-nocheck
+
+/*!
+ *	dotdotdot JS 4.1.0
+ *
+ *	dotdotdot.frebsite.nl
+ *
+ *	Copyright (c) Fred Heusschen
+ *	www.frebsite.nl
+ *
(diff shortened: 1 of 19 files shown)


web/src/utils/ellipsis.ts
```diff
@@ -0,0 +1,656 @@
...
+
+/*!
+ *	dotdotdot JS 4.1.0
+ *
+ *	dotdotdot.frebsite.nl
+ *
+ *	Copyright (c) Fred Heusschen
+ *	www.frebsite.nl
+ *
+ *	License: CC-BY-NC-4.0
```
RoccoSmit, line 12: The copyright license for this code is more restrictive than the MIT license memos uses particularly the noncommercial part.

PR #3932, opened 2024-09-16
fix: fixed the block layout in compact mode on iOS/MacOS

durianice:
When using `compact mode`, layout display is abnormal on **iOS browsers (Safari and latest Chrome) and macOS Safari** when the folded content contains code blocks. Use `mask-image` instead of `-webkit-line-clamp`.

Issue eg:
MacOS Safari
Snipaste_2024-09-16_03-04-41

iOS Chrome120

**After:**
SmartSelect_20240917_070810_Chrome
SmartSelect_20240917_070722_Chrome


boojack:
Thanks for your contribution, updated with a simpler implementation https://github.com/usememos/memos/commit/ea881338a9fc6740580ab794c84973872e27bf1c

image

PR #3935, opened 2024-09-18
feat: update Slovenian translation

thehijacker:
Added all that was missing and updated few existing ones.


boojack (approved):

Issue #3892, opened 2024-09-07
Case sensitive search results


deuts:
### Describe the bug

I just noted that when searching for entries, you have to follow the exact case that you used to write it to look them up. Wish we could non-case-sensitive search. Thanks

### Steps to reproduce

1. Go to search
2. Search for a word in a different case
3. You can't find that post

### The version of Memos you're using.

Version: v0.22.5

deuts:
I just checked search in a SQLite installation of Memos and search is not case sensitive. The bug above was noted in a Postgres installation of Memos.

PR #3938, opened 2024-09-19
fix: case insensetive memo search for postgres

sriniwas14:
This is a fix for the bug https://github.com/usememos/memos/issues/3892

Case insensitive search was not working for postgres driver, to mitigate this I changes `LIKE`  to `ILIKE`


boojack (approved):
LGTM, good catch!

PR #3946, opened 2024-09-22
fix: show future relative dates

RoccoSmit:
As per the below Discord comment, 
image

memos with a date set in the future have their relative time display as `now`

e.g. of future date (1 day in the future) showing `now`
image

After code change date references the future
image


boojack (approved):
LGTM

Issue #3947, opened 2024-09-23
Show number of memos with tag


alehmann76:
### Describe the solution you'd like

It would be probably helpful if i can see how many memos have which tag assigned:
e.g.
#home     2
#holiday  23
#job         345

This could be integrated in the "Tag" view on the right sidebar when the tree-mode is active

### Type of feature

User Experience (UX)

RoccoSmit:
There is already a count next to tags (qty > 1). Is this what you are referring to?
![image](https://github.com/user-attachments/assets/c260e270-c9fc-4a65-90e0-9bfe43bca6e3)

alehmann76:
Big sorry.... my mistake.... i used a rather old veriosn on my test server!
--> Everyting fine now! Great product and keep on fantastic work!!

zodac:
Looks like there is no count for tags in tree mode? Only in default mode?

PR #3970, opened 2024-09-30
feat: add tag count to tree view

RoccoSmit:
As per issue https://github.com/usememos/memos/issues/3947#issuecomment-2381853439 the tag list does not show usage numbers in tree view. This PR aims to add usage numbers

Design decisions:
Parent nodes only has a number next to them if they are uniquely used, they do not contain the sum of their child nodes

Examples:
image

Notes from the above:
- Parent1 has no direct reference so no usage number is displayed
- Parent2 has a direct reference so a usage number is displayed
- Parent3 child1 has no direct reference so no usage number is displayed
- Parent2 child1 has a direct reference so a usage number is displayed

### DIFF
--- a/web/src/components/TagTree.tsx
+++ b/web/src/components/TagTree.tsx
@@ -6,11 +6,12 @@ import { useMemoFilterStore } from "@/store/v1";
 interface Tag {
   key: string;
   text: string;
+  amount: string;
   subTags: Tag[];
 }
 
 interface Props {
-  tags: string[];
+  tags: [string, number][];
 }
 
 const TagTree = ({ tags: rawTags }: Props) => {
@@ -21,13 +22,15 @@ const TagTree = ({ tags: rawTags }: Props) => {
     const root: Tag = {
       key: "",
       text: "",
+      amount: "",
       subTags: [],
     };
 
     for (const tag of sortedTags) {
-      const subtags = tag.split("/");
+      const subtags = tag[0].split("/");
       let tempObj = root;
       let tagText = "";
+      let tagAmountText = "";
 
       for (let i = 0; i < subtags.length; i++) {
         const key = subtags[i];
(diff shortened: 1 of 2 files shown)


web/src/components/TagTree.tsx
```diff
@@ -111,7 +118,7 @@ const TagItemContainer: React.FC<TagItemContainerProps> = (props: TagItemContain
             <HashIcon className="w-4 h-auto shrink-0 mr-1 text-gray-400 dark:text-gray-500" />
           </div>
           <span className="truncate cursor-pointer hover:opacity-80" onClick={handleTagClick}>
-            {tag.key}
+            {tag.key} {tag.amount}
```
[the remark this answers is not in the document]
RoccoSmit: Logic updated

web/src/components/TagTree.tsx
```diff
@@ -6,11 +6,12 @@ import { useMemoFilterStore } from "@/store/v1";
 interface Tag {
   key: string;
   text: string;
+  amount: string;
```
boojack, line 9: number?

boojack (changes requested):

web/src/components/TagTree.tsx
RoccoSmit: I can see how this naming is misleading, its a string because it contains the formatted number text that contains the parentheses around the number. I've updated the name to be more descriptive

boojack: It's better to use number instead of formated string. For example: amount > 1 then show amount string.

RoccoSmit: Move in https://github.com/usememos/memos/pull/3970/commits/e2c96a39b77ae01bbde0d22ee11b041cf9d406c5

boojack (approved):
LGTM

PR #3991, opened 2024-10-06
Allow single letter user id.

rkj:
They do work - I use one (by login using Authelia), but then I am not able to update the profile to update avatar or add comment because "Invalid username: r" errors.

I don't know `go` at all, so not sure how to add proper test for this.

### DIFF
--- a/internal/util/resource_name.go
+++ b/internal/util/resource_name.go
@@ -3,5 +3,5 @@ package util
 import "regexp"
 
 var (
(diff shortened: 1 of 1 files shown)


RoccoSmit:
This change allows for 1 char usernames but not 2 and then > 2 again
Ideally a new regex expression (or other validation as regexes are hard to decipher) would be created for usernames. `UIDMatcher` is for checking user IDs which has a different structure to usernames

rkj:
Added a test and updated the regexp, I have zero context for this project, so not really sure how to make some more complex change.

Issue #4005, opened 2024-10-10
Toggling system dark theme requires reloading


rc2dev:
### Describe the bug

First of all, thanks for this awesome application!

I noticed that if I toggle my system dark mode, Memos doesn't follow the change, unless I refresh the page.

OS: Debian 12
Browser: Chromium 129
Memos: v0.22.5

### Steps to reproduce

1. Open Memos on your desktop browser.
2. Make sure Memos is set to follow system theme.
3. Toggle system theme (to dark or light).

Expected result: Memos theme also switches (to dark or light).
Actual result: Nothing happens.
Workaround: Manually refreshing the page changes the theme.

### The version of Memos you're using.

v0.22.5

RoccoSmit:
PR https://github.com/usememos/memos/pull/4010

PR #4010, opened 2024-10-11
fix: apply system theme dynamically

RoccoSmit:
As per issue https://github.com/usememos/memos/issues/4005 when toggling the operating system them between light and dark mode the web UI only takes effect after page reload

This PR allows for the them to change as the operating system theme changes without refreshing the page

e.g. 
Dark
image

Light
image

This change work with Browser theme settings as well


boojack (approved):
LG

Issue #4022, opened 2024-10-15
Can we disable the markdown key mapping?


d0zingcat:
### Describe the solution you'd like

I am a macOS/unix/linux user and I frequently use the unix-like keymapping day and night. 
But memos Editor register the keymap by default, e.g. Ctrl-B to generate a bolder syntax of markdown instead of moving cursor backforward.
It's quite annoying.

### Type of feature

User Interface (UI)

### Additional context

I will try to submit an enhancement branch which includes:
1. Settings/Memos: Add one option "Disable Shortcut"
2. Persistent 1. option into database(and do the db migration)
3. In web editor, read the options, if there's the disable config, then don't register the markdown keydown handler.

RoccoSmit:
Can the default be set as `true` (shortcut keys are in use).

With the default set to true users will be none the wiser of the change and their user experience won't be affected and those who want the change can toggle the setting.

If the default is set to `false` the are going to be users who use to have a feature that is going to be missing after upgrading and I'd imaging GitHub issues will follow.

RoccoSmit:
Fair point, my line of thinking is "how many users do/don't want shortcuts" (I for one like using `ctrl` + `enter` to quick save) and which is the bigger group being negatively impacted by removing the feature vs giving the ability to remove the feature

d0zingcat:
I share the same thoughts with @RoccoSmit as I think there maybe more users who is using the shortcuts more than those don't. A better way is to keep things as they were and add a new choice to change, namely like I've commented, maybe the name should be like `"Disable editor shortcut"` with the default set to `"false"`, just like the `"Disable public memos
"` option.

PR #4025, opened 2024-10-16
feat: allow to disable markdown shortcuts

d0zingcat:
Related: #4022


d0zingcat:
Except for issue has discussed about, I have to give a brief illustration about this PR.
Although the best way(to keep code clean and easy to understand) is that we should use `"EnableKeymap"` like option which defaults to `false`, if so most users who are used to the keymaps and have upgraded their memos will have to enable this option manually(even though we can issue one new notice to notice this update).
So I have to use another way, whereas we can give the option to disable this keymap.
BTW, I've added some locales I've been known of but the others I don't have any ideas. Should I use translator to fill them or wait for another PR to update?

proto/api/v1/workspace_setting_service.proto
```diff
@@ -115,6 +115,8 @@ message WorkspaceMemoRelatedSetting {
   string default_visibility = 9;
   // reactions is the list of reactions.
   repeated string reactions = 10;
+  // disable_markdown_keymap disallow the registration of markdown keymap.
+  bool disable_markdown_keymap = 11;
```
boojack, line 119: How about `disable_keyboard_shortcuts`?

image

https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-shortcuts-editor

boojack (approved):
LGTM, otherwise

proto/api/v1/workspace_setting_service.proto
d0zingcat: Make sense. How about `disable_markdown_keyboard_shortcuts` as it only suppress the markdown related shortcuts(Yeah it does not affect the command+enter to submit)?

boojack: Then it's recommand to use `disable_markdown_shortcuts` which is simpler and easier to understand.

boojack (approved):
LGTM, thanks for the contribution

PR #4046, opened 2024-10-23
fix: use UTC date for calendar filter

RoccoSmit:
As per issue: https://github.com/usememos/memos/issues/2765#issuecomment-2431346655

Memos are stored with a UTC timestamp. When the calendar dates are selected the UTC offset is not used to create the timestamps between which the memo creation date should fall.

e.g. of incorrect memos being shown
image
image
image

This PR aims to include the UTC offset in filter dates

e.g. of correct memos being shown
image
image
image


boojack (approved):
LGTM

Issue #4060, opened 2024-10-26
Reactions must not be empty when saving settings


towarisch:
### Describe the bug

When I try to send settings I see the message - "reactions must not be empty". I am unable to save settings.

### Steps to reproduce

1. Go to settings
2. Change any item
3. press save

### The version of Memos you're using.

0.23.0

### Screenshots or additional context

<img width="897" alt="reactions" src="https://github.com/user-attachments/assets/0dfb0b89-7541-44a0-9edf-2fbbd756567e">

RoccoSmit:
There is a migrations script that creates and backdates reactions for previous version users (e.g. 0.22.5 -> 0.23). it does not look like the migration scrip runs for users starting at version 0.23

@towarisch this is the default list of reactions 👍👎💛🔥👏😂👌🚀👀🤔🤡❓. Add the ones you want one at a time

towarisch:
Hello, 
thank you for your answer. Unfortunately I can't add any reactions. When I press the button, there is no possibility to select a reaction. I can only assume it is because I try to react to my own memo?
How do I add a reaction then?
<img width="558" alt="reactions2" src="https://github.com/user-attachments/assets/73306eb0-a2dd-4c7f-b411-c6e293c52229">

RoccoSmit:
When you went to `settings` -> `memo` you couldnt save because you did not have a value for `reactions`
![image](https://github.com/user-attachments/assets/00d246cc-e48e-47a7-a57b-d2500b74774a)

add some reactions from the list above, save and then they will show up when you try to react on a memo

towarisch:
Thank you it worked!


> Am 27.10.2024 um 07:53 schrieb RoccoSmit ***@***.***>:
> 
> 👍

PR #4061, opened 2024-10-27
migration version

RoccoSmit:
as per https://github.com/usememos/memos/issues/4060 the `00__reactions.sql` migration for `v0.23` does not run when you start with a fresh install

`GetCurrentSchemaVersion` get the current version (`v0.23`) and passes it into the migration logic. `getSchemaVersionOfMigrateScript` then increases the minor version so the version is `v0.23.1` and because of this the migration files to run are not found.

From debugging and testing this change work, but I'm not confident how it will affect already running version that upgrade to the latest. Some insights from someone with a deeper understanding of how the migrations system works would be appreciated.


RoccoSmit:
Closing PR to look into failing tests, I dont have time at the moment so will resubmit one resolved

PR #4094, opened 2024-11-03
fix: handle missing reaction

RoccoSmit:
as per https://github.com/usememos/memos/issues/4060 the `00__reactions.sql` migration for `v0.23` does not run when you start with a fresh install

Update:
Comments suggested we go with a default value set in code rather than migration changes so reverted and made those changes instead.

e.g. 
Memo with no reactions in db shows `thumbs up` default reaction in settings and on memo reaction
image
image

After reaction update the new reaction shows and `thumbs up` is not used
image
image


RoccoSmit:
Not sure what is causing checks to fail. Will revisit tomorrow

boojack:
@RoccoSmit Expectedly, the newly started instance should not execute any migrations. So I think what's needed here is to handle the default `reactions` value for memo setting.

Reference: https://github.com/usememos/memos/blob/main/store/workspace_setting.go#L157

RoccoSmit:
Makes sense. I wasn't sure if the `LASTEST.sql` files were for creating the initial db structure and the migrations were to add data to the tables.

If migrations are not run on the first install, would it make sense to add the insertion of the initial reactions to the `LATEST.sql` files? If not, I'll add 👍 as a default reaction value in `workspace_setting.go` if no value is set during save and load.

boojack:
> I'll add 👍 as a default reaction value in workspace_setting.go if no value is set during save and load.

SGTM

store/workspace_setting.go
```diff
@@ -140,6 +140,8 @@ func (s *Store) GetWorkspaceGeneralSetting(ctx context.Context) (*storepb.Worksp
 const (
 	// DefaultContentLengthLimit is the default limit of content length in bytes. 8KB.
 	DefaultContentLengthLimit = 8 * 1024
+	// DefaultReaction = thumbs up.
+	DefaultReaction = "\U0001f44d"
```
boojack, line 144:
```suggestion
	DefaultReactions = []string{"👍"}
```

RoccoSmit: I've update the code to use the emoji rather than the string representation of the emoji. Strangly when I tried using the emoji before my ide showed a compile error stating something to do with utf8. Cant replicate again so all good.

I tried setting the reactions as an array bit you cant set an array as a constant. If we intend to add more default reactions I'll pull the list into a variable.
image

boojack (approved):
LGTM

Issue #4105, opened 2024-11-07
V0.23.0 + mysql + docker-compose can't filter tags


Lishuhao-lucky:
### Describe the bug

![1730941001541](https://github.com/user-attachments/assets/376d075e-bf81-4f70-abf7-ed1d78137f6c)
![1730941013218](https://github.com/user-attachments/assets/759fadd3-0006-482a-a21d-a02ce72d1f58)
v0.23.0 +mysql + docker-compose can't filter tags , but v0.22 works

### Steps to reproduce

1、create a memo with tag
2、filter memo with tag just created
3、but can‘t find it

### The version of Memos you're using.

v0.23.0 pre

activ8xpgithub:
I have the same problem, also with mysql in separate docker.

PR #4119, opened 2024-11-10
fix: added missing "

RoccoSmit:
Discussion: https://github.com/orgs/usememos/discussions/4062 and potentially this issue https://github.com/usememos/memos/issues/4105

The mysql syntax is missing a closing "
Note: I dont have a mysql db instance so this change is untested. Apologies for not testing this pr


boojack (approved):
Good catch

Issue #1789, opened 2023-06-02
能不能在memos图标旁边加个刷新按钮，或者下拉刷新？


pxd0207:
### Is your feature request related to a problem?

手机端PWA网页无法刷新

### Describe the solution you'd like

电脑端还好可以刷新网页，但是在手机上的PWA没有刷新按钮要滑后台重进才能刷新。

athurg:
It'll be cool to pin the nav bar always in the top, and refresh the page by 
- pull-down after scroll to the top of page.
- click on the nav bar

athurg:
@pxd0207 You can just click on the site title `MEMOS` to reload the page with the latest `test` version.

PR #4128, opened 2024-11-14
feat: add pull to refresh for PagedMemoList

nlfox:
https://github.com/usememos/memos/issues/1789
Although above issue is fixed, tap to reload still a bit hard to find and non-intuitive, I feel adding pull to refresh feature will make user experience better. 

image

Looks like this^

GIF:

image

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -41,6 +41,7 @@
     "react-leaflet": "^4.2.1",
     "react-redux": "^9.1.2",
     "react-router-dom": "^6.27.0",
+    "react-simple-pull-to-refresh": "^1.3.3",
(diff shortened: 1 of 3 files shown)


boojack:
Can we customize the element of loader, the current one's style doesn't fit Memos very well.

nlfox:
> Can we customize the element of loader, the current one's style doesn't fit Memos very well.

Yes we can, the library I'm using didn't expose the icon for customization, we can switch to 
https://www.npmjs.com/package/react-pull-to-refresh

what kind of style do you suggest? do you have some icon that you want to use?

boojack (approved):
LGTM

Issue #4127, opened 2024-11-13
start failure 0.23.0.rc1


dream7180:
### Describe the bug

Archlinux, start failure, either pre-built binary or self-complied binary.
message:
[...]

PR #4131, opened 2024-11-15
Added charset that allow for emojis

RoccoSmit:
As per https://github.com/usememos/memos/issues/4127, emojis are causing migration and website usage issues

MySQL required a specific charset to store emojis.
This PR adds the required charset to the field that store the full list of available emojis for reactions as well as the field that stores emoji reactions added to memos

### DIFF
--- a/store/migration/mysql/prod/LATEST.sql
+++ b/store/migration/mysql/prod/LATEST.sql
@@ -1,3 +1,8 @@
+-- Set character set and collation to MySQL recommended default
+ALTER DATABASE `memos_prod`
+DEFAULT CHARACTER SET utf8mb4
+DEFAULT COLLATE utf8mb4_0900_ai_ci;
+
 -- migration_history
 CREATE TABLE `migration_history` (
   `version` VARCHAR(256) NOT NULL PRIMARY KEY,
--- a/store/migration/mysql/prod/LATEST.sql
+++ b/store/migration/mysql/prod/LATEST.sql
@@ -7,7 +7,7 @@ CREATE TABLE `migration_history` (
 -- system_setting
 CREATE TABLE `system_setting` (
   `name` VARCHAR(256) NOT NULL PRIMARY KEY,
-  `value` LONGTEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
+  `value` LONGTEXT NOT NULL,
   `description` TEXT NOT NULL
 );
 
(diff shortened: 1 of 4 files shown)


store/migration/mysql/prod/LATEST.sql
```diff
@@ -127,6 +127,6 @@ CREATE TABLE `reaction` (
   `created_ts` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `creator_id` INT NOT NULL,
   `content_id` VARCHAR(256) NOT NULL,
-  `reaction_type` VARCHAR(256) NOT NULL,
```
boojack, line 130: How about `TINYTEXT` without COLLATE option? I found that it's ok to save emoji into `memo.content` which is `TEXT` type.

boojack: I didn't set any special charset. But I found:

> MySQL includes character set support that enables you to store data using a variety of character sets and perform comparisons according to a variety of collations. The default MySQL server character set and collation are utf8mb4 and utf8mb4_0900_ai_ci, but you can specify character sets at the server, database, table, column, and string literal levels. To maximize interoperability and future-proofing of your data and applications, we recommend that you use the utf8mb4 character set whenever possible.

https://dev.mysql.com/doc/refman/8.4/en/charset.html#:~:text=The%20default%20MySQL%20server%20character,column%2C%20and%20string%20literal%20levels.

RoccoSmit: In the issue OP mentioned their db charset is `utf8`, not `utf8mb4`. I can set the db default charset to `utf8mb4` so that all users who decide to use a MySQL db align with recommended defaults

RoccoSmit:
@boojack , the mySQL docs shows using `mysql_prod` as the db name, but previous migrations scripts alter db `memo`. I used `mysql_prod` in the pr. What is the db name to use in the scripts?

store/migration/mysql/prod/LATEST.sql
RoccoSmit: Adding recommended settings in the docs makes sense for future users.

The emoji reactions feature (and the migrations that go with it) will cause all existing mysql users with this incorrect charset to not be able to upgrade to v0.23 as the migration script will fail to add the default values.

If the migrations are removed then the same subset of users will still not be able to update their settings, use emojis in their memos or add reactions to memos.

For existing users, adding the update in the code seems like the better user experience and should avoid github issues regarding upgrades

RoccoSmit: It shouldn't be needed for new dbs but it looks like there are people who created their dbs back when utf8 (utf8mb3) was the default. 
Issue OP was one of those and today another user had the same issue https://github.com/usememos/memos/issues/4127#issuecomment-2487803176

Are you suggesting to take the db alter out of the latest file and leave it in the migration file?

RoccoSmit: Up to you and @boojack so will go with what you choose.

Out of interest, why do you prefer to not add the db updates?
I can understand that settings `utf8mb4` in `latest` could open users up to being out of date if/when for e.g. `utf8mb5` becomes the new default, but migrations dont have that issue. I feel like I'm missing something.

boojack: @RoccoSmit It seems to be related to the user's database encoding configuration. We can provide documentation specifying the supported databases and encoding, rather than imposing restrictions in the code. https://stackoverflow.com/a/50264108

Issue #4142, opened 2024-11-22
Let users choose whether to turn on the folding function.


ZDY321:
### Describe the solution you'd like

I hope that like the old version, users can choose whether to enable the folding function. Some notes are very long and need to be folded.

### Type of feature

User Interface (UI)

RoccoSmit:
Are you referring to `settings` -> `memos` -> `enable auto compact`

ZDY321:
> Are you referring to `settings` -> `memos` -> `enable auto compact`

Yes, it can be set, but it can't be folded back after clicking Show More, and you need to refresh it

PR #4143, opened 2024-11-23
Added show less action and language files

RoccoSmit:
As per issue https://github.com/usememos/memos/issues/4142. Currently large memos has an option to start compacted and show more, but can then not be set back to a compacted state.

This PR is to add the functionality to set an expanded memo back to a compacted state

Show more state:
show more

Show less state:
show less

### DIFF
--- a/web/src/components/MemoContent/index.tsx
+++ b/web/src/components/MemoContent/index.tsx
@@ -26,13 +26,15 @@ interface Props {
   onDoubleClick?: (e: React.MouseEvent) => void;
 }
 
+type compactModes = "more" | "less" | "none";
+
 const MemoContent: React.FC<Props> = (props: Props) => {
   const { className, contentClassName, nodes, memoName, embeddedMemos, onClick, onDoubleClick } = props;
   const t = useTranslate();
   const currentUser = useCurrentUser();
   const memoStore = useMemoStore();
   const memoContentContainerRef = useRef<HTMLDivElement>(null);
-  const [showCompactMode, setShowCompactMode] = useState<boolean>(false);
+  const [showCompactMode, setShowCompactMode] = useState<compactModes>("none");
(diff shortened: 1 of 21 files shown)


web/src/components/MemoContent/index.tsx
```diff
@@ -26,13 +26,15 @@ interface Props {
   onDoubleClick?: (e: React.MouseEvent) => void;
 }
 
+type compactModes = "more" | "less" | "none";
```
boojack, line 29: * types should always be upper cases.

```suggestion
type ContentCompactView = "ALL" | "SNIPPET";
```

web/src/components/MemoContent/index.tsx
```diff
@@ -26,13 +26,15 @@ interface Props {
...
+
 const MemoContent: React.FC<Props> = (props: Props) => {
   const { className, contentClassName, nodes, memoName, embeddedMemos, onClick, onDoubleClick } = props;
   const t = useTranslate();
   const currentUser = useCurrentUser();
   const memoStore = useMemoStore();
   const memoContentContainerRef = useRef<HTMLDivElement>(null);
-  const [showCompactMode, setShowCompactMode] = useState<boolean>(false);
+  const [showCompactMode, setShowCompactMode] = useState<compactModes>("none");
```
boojack, line 37:
```suggestion
  const [showCompactMode, setShowCompactMode] = useState<ContentCompactView | undefined>(undefined);
```

boojack (approved):
Some nits

web/src/components/MemoContent/index.tsx
RoccoSmit: Changes applied

web/src/components/MemoContent/index.tsx
RoccoSmit: Changes applied

PR #4148, opened 2024-11-25
fix #4144 now when there is no sso then the Disallow password auth is…

### LINKED ISSUE #4144: Disable `Disallow password auth` when SSO not configured.
### Describe the bug

When SSO is not configured the `Disallow password auth` toggle should be disabled or not saved. As a user can disable this and lock themselves out. We should make an effort to ensure the user does not shoot themselves in the foot by accidentally configuring this.

## Remediations
- Disable the toggle button when SSO is not configured.

### Steps to reproduce

1. Login as an admin
2. Go Settings
3. Go to SSO tab
4. Ensure no SSO is configured
5. Go to System tab
6. Toggle `Disallow password auth`
7. Click on profile picture -> Sign out

### The version of Memos you're using.

0.22.5

_No response_

X1Vi:
… disabled

### DIFF
--- a/web/src/components/Settings/WorkspaceSection.tsx
+++ b/web/src/components/Settings/WorkspaceSection.tsx
@@ -10,6 +10,7 @@ import { WorkspaceGeneralSetting } from "@/types/proto/api/v1/workspace_setting_
 import { WorkspaceSettingKey } from "@/types/proto/store/workspace_setting";
 import { useTranslate } from "@/utils/i18n";
 import showUpdateCustomizedProfileDialog from "../UpdateCustomizedProfileDialog";
+import { identityProviderServiceClient } from "@/grpcweb";
 
 const WorkspaceSection = () => {
   const t = useTranslate();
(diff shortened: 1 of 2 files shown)


X1Vi:
fix now if SSOs are not there then disallows password auth is disabled

web/src/components/Settings/WorkspaceSection.tsx
```diff
@@ -18,6 +20,7 @@ const WorkspaceSection = () => {
     workspaceSettingStore.getWorkspaceSettingByKey(WorkspaceSettingKey.GENERAL)?.generalSetting || {},
   );
   const [workspaceGeneralSetting, setWorkspaceGeneralSetting] = useState<WorkspaceGeneralSetting>(originalSetting);
+  const [identityProviderList, setIdentityProviderList] = useState<IdentityProvider[]>([]);
```
RoccoSmit, line 23: Can this be a simple variable? I dont believe there is and need to have it's state tracked.

web/src/components/Settings/WorkspaceSection.tsx
```diff
@@ -50,6 +53,15 @@ const WorkspaceSection = () => {
     toast.success(t("message.update-succeed"));
   };
 
+  useEffect(() => {
+    fetchIdentityProviderList();
+  }, []);
+
+  const fetchIdentityProviderList = async () => {
+    const { identityProviders } = await identityProviderServiceClient.listIdentityProviders({});
```
RoccoSmit, line 61: I've seen the code use a `useasync` pattern. Is this applicable here?

X1Vi: > I've seen the code use a `useasync` pattern. Is this applicable here?

I have already copied all this from the SSO section it is using the exact same code

PR #4151, opened 2024-11-27
chore: apply styles to the correct month

Yamico:
image

In the latest version, there's a small issue with the ActivityCalendar component where styles are being applied to the same date in the previous month. Here's some code to fix it.

### DIFF
--- a/web/src/components/ActivityCalendar.tsx
+++ b/web/src/components/ActivityCalendar.tsx
@@ -84,9 +84,10 @@ const ActivityCalendar = (props: Props) => {
                 "w-6 h-6 text-xs rounded-xl flex justify-center items-center border cursor-default",
                 "text-gray-400",
                 item.isCurrentMonth ? getCellAdditionalStyles(count, maxCount) : "opacity-60",
-                isToday && "border-zinc-400",
-                isSelected && "font-bold border-zinc-400",
-                !isToday && !isSelected && "border-transparent",
+                item.isCurrentMonth && isToday && "border-zinc-400",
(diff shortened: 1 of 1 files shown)


web/src/components/ActivityCalendar.tsx
```diff
@@ -84,9 +84,10 @@ const ActivityCalendar = (props: Props) => {
                 "w-6 h-6 text-xs rounded-xl flex justify-center items-center border cursor-default",
                 "text-gray-400",
                 item.isCurrentMonth ? getCellAdditionalStyles(count, maxCount) : "opacity-60",
-                isToday && "border-zinc-400",
-                isSelected && "font-bold border-zinc-400",
-                !isToday && !isSelected && "border-transparent",
+                item.isCurrentMonth && isToday && "border-zinc-400",
+                item.isCurrentMonth && isSelected && "bg-success font-bold",
```
RoccoSmit, line 88: Can we remove the `bg-success` from here. This is causing the date to indicate a memo is present event when there isn't one. With no memo there should only be a border set

expected:
image

this PR:
image

Yamico: You are correct. In my custom implementation, the logic for `isSelected` was modified, and the styles were also adjusted accordingly. Apologies for overlooking this and causing issues. Please review this new submission. #4153 

By the way, would it be possible to modify the way `selectedDate` is obtained? Based on the name of this variable, my understanding is that it should be set by selecting a date via a mouse click. However, obtaining it through `props` instead of state seems to prevent this variable from being updated by mouse clicks.

PR #4159, opened 2024-11-29
fix: search conditions affected by pull-to-refresh

freelylee:
Fix the failure of search conditions affected by pull-to-refresh.


boojack (approved):
Good catch!

Issue #4178, opened 2024-12-04
using tags among all users


non9mer:
### Describe the solution you'd like

Describe the solution you'd like
HELLO!
It would be great to be able to see and use tags in workspace that were created by another user.
So, let's say another user created a note in the workspace section, creating new tags for it, but the current implementation does not allow these tags to be shared between other users, and this seems very useful.

P.S. this is my second try to create this issue

Best wishes!

### Type of feature

User Interface (UI)

### Additional context

I'm trying to make it so that when new tags are created, they are shared among all users, appearing in the right column, under the TAGS section.

Issue #3985, opened 2024-10-03
using tags among all users


non9mer:
### Describe the solution you'd like

HELLO!
It would be great to be able to see and use tags in _workspace_ that were created by another user.
So, let's say another user created a note in the _workspace_  section, creating new tags for it, but the current implementation does not allow these tags to be shared between other users, and this seems very useful.

Best wishes!

### Type of feature

User Interface (UI)

### Additional context

I'm trying to make it so that when new tags are created, they are shared among all users, appearing in the right column, under the _TAGS_ section.

RoccoSmit:
@boojack, I'd like to implement this.

The 2 ways I was thinking about are:

1) Create a `useTagsdataStore` that leverages `useMemoMetadataStore`. `useMemoMetadataStore` would get a all the metadata info including tags (`private`, `protected` and `public`) for the user and `useTagsdataStore` would make another call and get all user's `protected` and `public` tags and merge the 2 lists of tags together.

This approach has some shortcoming in that it will retrieve the user's `public` and `protected` tags 2 times, but it is a smaller change in the code and uses existing mechanisms.

2) Add another filter list that will be mapped to `or` logical operators in the `.sql` file queries. This way the db query can get all memos that have a `creatorId` of x `or` `protected` and `public` visibility all in a single query.

This approach is more involved as it will require changes across FE, server and db store files for all 3 db types.

Thoughts?

non9mer:
up

Issue #3103, opened 2024-03-15
More transparent tags


mikrofyr:
### Describe the solution you'd like

It would be nice to be able to browse tags of other users, or have a more flexible search on tags/posts from other users. 
E.g search for shopping lists created by other members in my family.

Right now the only option is go to explore, and scroll until you get lucky. 
Or request someone to share the link, and keep the reference availeble until it is no longer intersting.

### Type of feature

User Experience (UX)

### Additional context

Improve cross user functionality.

PR #4198, opened 2024-12-15
Share tags

RoccoSmit:
As per https://github.com/usememos/memos/issues/3103, https://github.com/usememos/memos/issues/3985 and https://github.com/usememos/memos/issues/4178 users are currently not able to share tag across `workspaces`.

This PR is to allow for `workspace` and `public` tags to be shared.

**Design decision:**
- I tried to keep with the current design of using the `ListMemos` endpoint and setting the data that is needed via the `request.View` flag similar to what is done with memos vs metadata. I could not piggyback off the metadata flow as the metadata numbers would include other user's memos as well 
- I started going down the path of creating a generic filter structure and consumer that would allow for building filters that have multiple checks, but it got very involved so I went for the simpler adding of the `ShareTags` property on the `filter` object and adding extra SQL to go along with it.
- Users can `rename` and `edit` other user's tags. This will only take effect on memos that are marked as `workspace` or `public`
- I didnt want the PR to get too big so I did not add other languages for the `share tags` setting. I can do that separately if the PR gets approved.

**Experience:**

in the below examples there are 2 users (`User-1` and `User-2`). They both have 1 `private`, `workspace` and `public` memo with an identifying tag name in each. each user also has a `workspace` level `link`, `todo` and `code` memo.

Settings screen
image

Shared tags showing in the tag panel and metadata numbers are for user only
image

User only has access to their own private tags
image

User has access to shared tags
image

The explore page still functions as expected logged in and out
logged in
image

logged out
image

**Tested in:**
- `Firefox` :heavy_check_mark: 
- `Chromium` :heavy_check_mark: 
- `SQLite` :heavy_check_mark: 
- `Mysql` :heavy_check_mark: 
- `Postgres` :heavy_check_mark:


RoccoSmit:
Merged latest master into PR

@boojack, @johnnyjoygh, is this feature under consideration to be added? If it is, I'll keep it up to date with `main` till the PR gets reviewed. If not I'll remove the PR to clean up the PR list

PR #4238, opened 2024-12-28
refactor: enhance hyperlink handling in markdown shortcuts

chriscurrycc:
Recognize url in hyperlink shortcuts. This is similar to GitHub.


chriscurrycc:
Thank you for your prompt response. Here is my intended outcome: when I select a text segment resembling a URL and press `cmd+k`, I want it to be the link section wrapped by the parenthesis.

Dec-29-2024 22-36-04

PR #4239, opened 2024-12-30
feat: persist memo filters with url query

chriscurrycc:
- Implemented synchronization of memo filters with URL search parameters.
- Enhanced memo store to manage current request state for fetching memos, allowing for request cancellation, avoiding the race conditions.

This is very helpful in this situation: Applying filters on the list page, viewing a single memo detail page, and then returning will keep the applied filters intact.

We can also share or save the URL with filters to quickly access our filtered memos.


RoccoSmit:
For some context:

This use to be a feature in `v0.22.0` and was then removed when the filter UX changed.

Some code samples from `v0.22.0`: [MemoFilter.tsx](https://github.com/usememos/memos/blob/d7889d9903f0035bd983a97335e0d9f0337e9e2c/web/src/components/MemoFilter.tsx), [filter.ts](https://github.com/usememos/memos/blob/d7889d9903f0035bd983a97335e0d9f0337e9e2c/web/src/store/module/filter.ts), [useFilterWithUrlParams](https://github.com/usememos/memos/blob/d7889d9903f0035bd983a97335e0d9f0337e9e2c/web/src/hooks/useFilterWithUrlParams.ts)

This issue has been requested since the change, e.g. https://github.com/orgs/usememos/discussions/3999, https://github.com/orgs/usememos/discussions/3809 but has not got much traction.

Might need some feedback from @boojack or @johnnyjoygh on if this feature is part of their future vision of Memos

chriscurrycc:
Thank you for providing the context. However, I would like to know why this feature was removed since `v0.22.0`. I believe it is a core feature.

chriscurrycc:
By the way, I used to be a [Flomo](https://flomoapp.com/) user, but recently migrated all my notes to Memos. Thank you for building such a great open-source software.

I plan to contribute to Memos in the future. I have also written an article on [how I developed Memos](https://github.com/chriscurrycc/memos/issues/8). I will submit PRs for bug fixes, optimizations, etc. If it's something more personalized, I will modify it in my own repository. Thank you again for your hard work, and I hope to contribute to helping you as well.

Issue #4267, opened 2025-01-03
Missing german (de) translations


tungerno:
### Describe the bug

If I switch to German language, there some not translate phrases in the GUI.
Which keys are not translate could be found by compare en.json with de.json.
By fixing that issue, it would be nice to remove old unnecessary keys.

### Steps to reproduce

1. Start a clean docker image in v.0.23.0 and open a browser with http://localhost:5230
2. Register a new user
3. Go to Settings -> Preferences -> Basics -> Language and switch to "Deutsch"
4. E.g. go to  Einstellungen-> Administration -> System and there you should see the same like in the screenshot below 

### The version of Memos you're using

v0.23.0

### Screenshots or additional context

![image](https://github.com/user-attachments/assets/9d5a2a3d-c81c-40c6-8ead-e2af990fdd3b)

RoccoSmit:
@tungerno, feel free to add a PR for the missing German translations

PR #4268, opened 2025-01-03
add missing german translations

tungerno:
- Add missing translated keys with corresponding German translations. 
- See also issue [BUG] Missing german (de) translations #4267 
- remove unused keys (compare to en.json file)
I assumed that the English local file is the leading one and that keys that are not listed here are obsolete and can be deleted.


RoccoSmit (approved):
LGTM

PR #4289, opened 2025-01-11
fix #4257 fixed ui for mermaid diagrams in dark mode

### LINKED ISSUE #4257: Mermaid Diagrams Are Unreadble in Dark Theme
### Describe the bug

Have a look at screenshot below. The text is unreadable.

image

### Steps to reproduce

1. Enable dark theme
2. Use below mermaid diagram and view it for yourself
```
sequenceDiagram
    Alice->>+John: Hello John, how are you?
    Alice->>+John: John, can you hear me?
    John-->>-Alice: Hi Alice, I can hear you!
    John-->>-Alice: I feel great!
```

latest

_No response_

X1Vi:
fixes #4257 
The diagrams are now visible and can be seen easily in dark mode

image

### DIFF
--- a/web/src/components/MemoContent/MermaidBlock.tsx
+++ b/web/src/components/MemoContent/MermaidBlock.tsx
@@ -11,7 +11,8 @@ const MermaidBlock: React.FC<Props> = ({ content }: Props) => {
     // Dynamically import mermaid to ensure compatibility with Vite
     const initializeMermaid = async () => {
       const mermaid = (await import("mermaid")).default;
-      mermaid.initialize({ startOnLoad: false, theme: "default" });
+      mermaid.initialize({ startOnLoad: false, theme: 'dark', darkMode: true
+      });
       if (mermaidDockBlock.current) {
         mermaid.run({
           nodes: [mermaidDockBlock.current],


X1Vi:
When Dark theme
image

When light theme
image

RoccoSmit:
When you select `Follow System` as the theme the `commonContext` value is `system`, meaning that system set dark mode will use light mode styles.

Have a look [here](https://github.com/usememos/memos/blob/34c26a394a6f69e66085c543ebf358f243528246/web/src/App.tsx#L92-L99) to see how this has been handled in the past

X1Vi:
> When you select `Follow System` as the theme the `commonContext.appearance` value is `system`, meaning that system set dark mode will use light mode styles.
> 
> Have a look [here](https://github.com/usememos/memos/blob/34c26a394a6f69e66085c543ebf358f243528246/web/src/App.tsx#L92-L99) to see how this has been handled in the past

Can you elaborate more as to what you want exactly ? because if it is `system` it just goes to default.

web/src/components/MemoContent/MermaidBlock.tsx
```diff
@@ -1,17 +1,18 @@
...
+import { useCommonContext } from "@/layouts/CommonContextProvider";
 
 interface Props {
   content: string;
 }
 
 const MermaidBlock: React.FC<Props> = ({ content }: Props) => {
   const mermaidDockBlock = useRef<null>(null);
-
+  const commonContext = useCommonContext();
```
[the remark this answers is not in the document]
X1Vi: I think I am handling and getting the theme when user selects "Follow system" in the new commit. If you want to set certain specific themes at certain conditions then do tell me and dont forget to ping me.

X1Vi:
@johnnyjoygh I have made a new and better solution I wasn't completely aware that how the theme works completely (Didn't knew there was another hook as well). I have selected themes which I sought suitable if you feel like they are not suitable then feel free to ping me. I think we will have to just adjust the themes.

web/src/components/MemoContent/MermaidBlock.tsx
```diff
@@ -1,17 +1,35 @@
 import { useEffect, useRef } from "react";
+import { getSystemColorScheme } from "@/helpers/utils";
+import { useCommonContext } from "@/layouts/CommonContextProvider";
```
RoccoSmit, line 3:
```suggestion
import { useColorScheme } from "@mui/joy";
```

web/src/components/MemoContent/MermaidBlock.tsx
```diff
@@ -1,17 +1,35 @@
...
+    } else {
+      if (getSystemColorScheme() === "dark") {
+        return "dark";
+      } else if (getSystemColorScheme() === "light") {
+        return "neutral";
+      } else {
+        return "default";
+      }
+    }
+  };
```
RoccoSmit, line 27: It should be possible to simplify the theme selection logic to this. 

```suggestion
const { mode } = useColorScheme();

const handleMermaidTheme = () => {
  return mode == 'dark'
    ? 'dark'
    : 'default'
};
```

`app.tsx` checks `commonContext.appearance` and `getSystemColorScheme()` to determine the theme and uses `setMode(currentAppearance);` to set `mui-joy`'s `mode` var with that value so no need to redo it.

Note:
`mui`s `mode` is the only variable out of the vars that are set for theme that get dynamically updates after page load. You can see this in the browser's local storage view, where a copy of this values is stored, as the OS/browser theme changes. 
I was going to suggest adding `mode` to the `useEffect` variables to track so that any OS/browser changes take effect as the user makes the change, but it doesn't work. `Mermaid.js` doesnt allow for `re-initialization` out of the box and some workarounds are needed to get it to happen. see [here](https://github.com/mermaid-js/mermaid/issues/1945). I dont think is worth the effort if a user can refresh to see the new theme and the user potentially doesn't ever change their theme once its set.

X1Vi:
I have fixed the changes as told by @RoccoSmit

RoccoSmit (approved):
LGTM

X1Vi:
> Please use `const` instead `function` call.

Done

PR #4305, opened 2025-01-21
refactor(sql): refactor sql queries to use prepared statements

fernandoonrails:
in this commit, a the sql queries for the resource domain in the database were refactored to use prepared statements instead of simple queries, as it usually, makes the project safer when exposed to public internet

### DIFF
--- a/store/db/mysql/resource.go
+++ b/store/db/mysql/resource.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


fernandoonrails:
will take a look at it later, any help or critic to improve it would be much appreciated 😁

store/db/mysql/resource.go
```diff
@@ -30,8 +31,13 @@ func (d *DB) CreateResource(ctx context.Context, create *store.Resource) (*store
 	}
 	args := []any{create.UID, create.Filename, create.Blob, create.Type, create.Size, create.CreatorID, create.MemoID, storageType, create.Reference, payloadString}
 
-	stmt := "INSERT INTO `resource` (" + strings.Join(fields, ", ") + ") VALUES (" + strings.Join(placeholder, ", ") + ")"
-	result, err := d.db.ExecContext(ctx, stmt, args...)
+	query := "INSERT INTO `resource` (" + strings.Join(fields, ", ") + ") VALUES (" + strings.Join(placeholder, ", ") + ")"
+	stmt, err := d.db.PrepareContext(ctx, query)
```
RoccoSmit, line 35: Are these `db.PrepareContext` calls needed? 

The way I understand the [docs](https://go.dev/doc/database/prepared-statements) is that they are intended to wrap SQL statements and are added to some form of a SQL statement file from where they can be called and reused or passed around between functions.

The way they are used in this PR does not allow for reuse and looks to me like a step that can be omitted.

PR #4229, opened 2024-12-24
refactor: move tags from property to payload

johnnyjoygh:

PR #4327, opened 2025-01-29
fix: use correct separator in migration file name

l2dy:
Fixes: e913271f15ef4fb3116bcc17f291622a15094dfc ("refactor: move tags from property to payload (#4229)")

See:
https://github.com/usememos/memos/blob/75fd4223bff4f4aa068a87ad76170098e688f5bd/store/migrator.go#L28-L30


l2dy:
It seems that the file was added in mistake, because the corresponding LATEST.sql file was not updated. Maybe we should just delete the file instead?

Issue #4330, opened 2025-01-30
Calendar may not be colored even on days with notes


shungiku:
### Describe the bug

Calendar may not be colored even on days with notes, in both light and dark mode.

### Steps to reproduce

1. Go to https://demo.usememos.com/explore
2. See calendar

### The version of Memos you're using

Version: v0.23.1

### Screenshots or additional context

![Image](https://github.com/user-attachments/assets/62fa84b1-4828-4c4a-ae72-b0ccd6826f66)

shungiku:
Hello. I found additional conditions that reproduce this issue.
1. On calendar before previous month
2. Number of memos created on that day is 3 or more

I would expect it to be related to this part.
https://github.com/usememos/memos/blob/66db662e33dda6b1fbd04495d7fd353d15c95f64/web/src/components/ActivityCalendar.tsx#L20

![Image](https://github.com/user-attachments/assets/633510d3-cce0-4702-b085-5ebc7bc436c2)

PR #4352, opened 2025-02-02
fix: calendar coloring issue

shungiku:
Fixed issue with calendar color not being painted according to conditions.
https://github.com/usememos/memos/issues/4330

### DIFF
--- a/web/src/components/ActivityCalendar.tsx
+++ b/web/src/components/ActivityCalendar.tsx
@@ -17,11 +17,11 @@ const getCellAdditionalStyles = (count: number, maxCount: number) => {
   if (count === 0) {
     return "";
   }
-  if (count >= 3) {
-    const ratio = count / maxCount;
+  const ratio = count / maxCount;
+  if (count >= 3 && ratio > 0.4) {
     if (ratio > 0.7) {
       return "bg-primary-darker/80 text-gray-100 dark:opacity-80";
-    } else if (ratio > 0.4) {
+    } else {
       return "bg-primary/80 text-gray-100 dark:opacity-80";
     }
   } else {


web/src/components/ActivityCalendar.tsx
```diff
@@ -17,11 +17,11 @@ const getCellAdditionalStyles = (count: number, maxCount: number) => {
   if (count === 0) {
     return "";
   }
-  if (count >= 3) {
-    const ratio = count / maxCount;
+  const ratio = count / maxCount;
+  if (count >= 3 && ratio > 0.4) {
```
RoccoSmit, line 21: This change will cause dates with more than `3` memos but at a `ration < 0.4` to render as dates with less than `3` memos i.e using `bg-primary/70` rather than `bg-primary/80`

Can we stay with the initial code design of dates with `>= 3` memos using `bg-primary/80`
I'd suggest either going down the middle and using `ration > 0.5` rather than `0.4` and `0.7`
or up the opacity to 90 for `ratios > 0.7`

RoccoSmit: I dont have any context on why the ratios are used after 3 notes.
@boojack , @johnnyjoygh? The github contributions method looks like a good option

shungiku: @RoccoSmit @johnnyjoygh 
Light mode was adjusted to be darker when there are a lot of memos and dark mode to be lighter.
{08945F36-C4AD-4E0F-96AE-628BF993513F}
{E5A24CDD-69F2-490F-A0E0-B8AC3B86DDA6}

web/tailwind.config.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
           DEFAULT: "#0d9488", // Teal 600
           dark: "#0f766e", // Teal 700
           darker: "#0d5a56", // Teal 800
+          lighter: "#39d353", // Green
```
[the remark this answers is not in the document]
shungiku: It looks nice!
{753BADA0-A5B5-4852-937D-68DDA2B474C1}

PR #4383, opened 2025-02-08
Add id to RSS feed items

hartlco:
This change adds the permalink as GUID to RSS feed items. [micro.blog](https://micro.blog) requires feed items to provide a GUID for content ingestion.

**Please note**: I was not able to verify this change locally as the `u/USER/rss.xml` 404s before and after this change locally.


johnnyjoygh (approved):
LGTM

PR #4397, opened 2025-02-11
Chinese localization enhancements

xtzero:
I add some Chinese localization config to file: web/src/locales/zh-Hans.json, and modified the code that referenced them on the page.


johnnyjoygh (approved):
LGTM

PR #4405, opened 2025-02-14
chore: update chinese i18n at member setting section

taosin:
chore: update chinese i18n at member setting section
image


johnnyjoygh (approved):
LGTM

PR #4408, opened 2025-02-14
Allow storing resource references >256 chars long in mysql

### LINKED ISSUE #4322: Cloudflare Storage: Link too long to be stored to `resource`.`reference`
### Describe the bug

When setting up Memo to use Cloudflare R2, upon saving, it shows the following error on the screen:

Image
 

Here's the console log:

```
memos  | 2025/01/27 14:54:55 ERROR server error method=/memos.api.v1.ResourceService/CreateResource error="rpc error: code = Internal desc = failed to create resource: Error 1406 (22001): Data too long for column 'reference' at row 1"
```

Image

After digging around, I think i know the issue: the `reference` column in `resource` name is not sufficient to store signed URL from Clouadlare R2. The usual Cloudflare R2 URL uses the `https://[bucket].[account-id].r2.cloudflarestorage.com`. Here's R2 sample link:

```
https://bucket-name.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.r2.cloudflarestorage.com/assets/aaaaaaaaaaaaaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa.jpg?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa%2Faaaaaaaa%2Fauto%2Fs3%2Faws4_request&X-Amz-Date=20250123T110829Z&X-Amz-Expires=432000&X-Amz-SignedHeaders=host&x-id=GetObject&X-Amz-Signature=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

It has 427 charactes while the column were set only to `varchar(256)`. I think we can just create a new migration to update the column to be `TEXT` for mysql. Waht do you think? I kinda want to send a PR for this.

### Steps to reproduce

1. Install & Setup memo with the following info:
  - Install using Docker Compose
  - Use MariaDB.
2. Use Cloudflare R2 for storage option, as described in the following documentations:
  https://www.usememos.com/docs/advanced-settings/cloudflare-r2
3. Save the configurations
4. Create a new Memo, and upload an image. The screen immediately shows error.

v0.23.0

_No response_

cdzombak:
This PR  fixes #4322.

I have verified this on my own Memos instance, running MySQL 8.4, by running: 

```
ALTER TABLE `resource` MODIFY `reference` TEXT NOT NULL DEFAULT ('');
```

After this change, B2-backed file uploads appear to work as expected.

Caveat: I am not familiar with how this project likes to handle version numbering and DB migrations, so please accept my apologies if this migration isn't in quite the right place at this time.

### DIFF
--- a/store/migration/mysql/0.25/00__s3_reference_length.sql
+++ b/store/migration/mysql/0.25/00__s3_reference_length.sql
@@ -1,0 +1,3 @@
+-- https://github.com/usememos/memos/issues/4322
+
+ALTER TABLE `resource` MODIFY `reference` TEXT NOT NULL DEFAULT ('');
--- a/store/migration/mysql/LATEST.sql
+++ b/store/migration/mysql/LATEST.sql
@@ -77,7 +77,7 @@ CREATE TABLE `resource` (
   `size` INT NOT NULL DEFAULT '0',
   `memo_id` INT DEFAULT NULL,
   `storage_type` VARCHAR(256) NOT NULL DEFAULT '',
-  `reference` VARCHAR(256) NOT NULL DEFAULT '',
+  `reference` TEXT NOT NULL DEFAULT (''),
   `payload` TEXT NOT NULL
 );
 


johnnyjoygh (approved):
LGTM

PR #4411, opened 2025-02-16
feat: Trim space on webhook url

ProjectOrangeJuice:
This removes spaces that can cause invalid URLs. While I was using the webhooks I kept pasting urls with spaces and discovering the problem in the logs later.

Eg,
`http://192.168.1.10:8080/webhook ` is accepted, but will return 404 when used. This change removes the space at the end


johnnyjoygh (approved):
LGTM

Issue #4413, opened 2025-02-17
CVE-2025-22952 memos SSRF


joe912693295:
### Describe the bug

elestio/memos:latest and neosmemo/memos:latest both have SSRF vulnerability, which allowing attackers to access to intranet IP addresses and protocols. Despite efforts to limit access protocols to HTTP and HTTPS, attackers can still accessing other ports running on localhost. This vulnerability enables attackers to access any asset on the internal network, sniff web services on the internal network, scan hosts on the internal network, and potentially access weak internal endpoints. The vulnerability is due to insufficient validation of user-supplied URLs, which can be exploited to perform SSRF attacks.

### Steps to reproduce

┌──(root㉿kali)-[~/Downloads]
└─# cat docker-compose.yaml 
version: '3.3'
services:
    memos:
        image: 'elestio/memos:latest'
        restart: always
        ports:
            - '172.17.0.1:5230:5230'
        volumes:
            - './memos/:/var/opt/memos'

┌──(root㉿kali)-[~/Downloads]
└─# docker-compose up -d
WARN[0000] /root/Downloads/docker-compose.yaml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
[+] Running 2/2
 ✔ Network downloads_default    Created                                                                                                                            0.1s 
 ✔ Container downloads-memos-1  Started                                                                                                                            0.2s 

┌──(root㉿kali)-[~/Downloads]
└─# curl --head 172.17.0.1:5230
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 1162
Content-Type: text/html; charset=utf-8
Date: Mon, 17 Feb 2025 06:18:20 GMT

┌──(root㉿kali)-[~/Downloads]
└─# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 00:0c:29:0d:08:46 brd ff:ff:ff:ff:ff:ff
[...]

PR #4421, opened 2025-02-19
fix: Prevent previewing internal network web pages.

CallMeMhz:
#4413 It doesn’t seem like a complete solution; it would be better if page preview could be done from the frontend.


johnnyjoygh:
> it would be better if page preview could be done from the frontend.

Due to cross-origin restrictions on the browser(frontend), many websites can't be reliably accessed.

johnnyjoygh (approved):
LGTM

PR #4427, opened 2025-02-21
feat: support YouTube video thumbnail in link preview

CallMeMhz:
image
image

### DIFF
--- a/plugin/httpgetter/html_meta.go
+++ b/plugin/httpgetter/html_meta.go
@@ -2,6 +2,7 @@ package httpgetter
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


plugin/httpgetter/html_meta.go
```diff
@@ -121,3 +123,14 @@ func validateURL(urlStr string) error {
 
 	return nil
 }
+
+func enrichSiteMeta(url *url.URL, meta *HTMLMeta) {
+	if url.Hostname() == "www.youtube.com" {
+		if url.Path == "/watch" {
```
johnnyjoygh, line 129:
```suggestion
	if url.Hostname() == "www.youtube.com" && url.Path == "/watch" {
```

johnnyjoygh (approved):
LGTM

plugin/httpgetter/html_meta.go
CallMeMhz: Thank you for your review. Originally, I wanted to use a switch case for each domain, but in order to pass the static check, I had to use an if condition instead.

PR #4428, opened 2025-02-21
Prevent attackers from exploiting redirect attack GetLinkMetadata API

CallMeMhz:
The previous fix for issue #4413  (commit #4421) still has some flaws. An attacker could potentially exploit a crafted redirection address to redirect to the internal network, thereby leveraging the GetLinkMetadata API to probe internal network information.


johnnyjoygh (approved):
LGTM

Issue #4396, opened 2025-02-11
Editor ignores current indentation when creating new list item


mbaum0:
### Describe the bug

The text editor ignores the current indentation level when adding an item to a list. After pressing enter, the next list item is created with all indentation removed. This makes writing sub-lists tedious, as you have to manually add whitespace.

### Steps to reproduce

1. Create a non-list item
2. Start a sub-list and add the first item
3. Press enter to add another item
4. See that the cursor creates another item at the parent's indentation level.

If the parent is a list item, the indentation will be correct but the sub-list symbol will not be added

### The version of Memos you're using

v0.24.0

PR #4433, opened 2025-02-22
Match sublist indentation when adding a new item

mbaum0:
Closes: https://github.com/usememos/memos/issues/4396

It looks like the indentation match was overwritten instead of being appended to the sub list symbol.

I fixed this so the matched spacing is properly appended to the symbol when enter is pressed when editing a sublist.

I also discovered an issue where if the parent node was a list, the child nodes would be indented properly, but the symbol was not included after pressing enter. I fixed this by recursively digging into the child nodes.

Kudos to the easy dev environment setup. I was able to start debugging the issue in <5 minutes 😄


johnnyjoygh (approved):
LGTM

PR #4445, opened 2025-02-26
feat: optimize initial load performance by implementing lazy loading and code splitting

nbb2025:
Here is the pagespeed score!
image


johnnyjoygh (approved):
LGTM, good to go!

PR #4488, opened 2025-03-11
feat: update i18n for zh-hans

taosin:
merge this feature ASAP.

### DIFF
--- a/web/src/components/MemoFilters.tsx
+++ b/web/src/components/MemoFilters.tsx
@@ -25,20 +25,19 @@ const MemoFilters = () => {
     }
     if (filter.factor.startsWith("property.")) {
       const factorLabel = filter.factor.replace("property.", "");
-			switch (factorLabel) {
-				case "hasLink":
-					return t("filters.has-link");
-				case "hasCode":
-					return t("filters.has-code");
-			 	case "hasTaskList":
-				 	return t("filters.has-task-list");
-			  default:
-					return factorLabel;
-			}
+      switch (factorLabel) {
+        case "hasLink":
+          return t("filters.has-link");
+        case "hasCode":
+          return t("filters.has-code");
+        case "hasTaskList":
+          return t("filters.has-task-list");
+        default:
+          return factorLabel;
+      }
     }
     return filter.factor;
   };
-
 
   if (filters.length === 0) {
     return undefined;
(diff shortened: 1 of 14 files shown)


web/src/components/StatisticsView.tsx
```diff
@@ -95,7 +95,7 @@ const StatisticsView = observer(() => {
       <div className="pt-1 w-full flex flex-row justify-start items-center gap-x-2 gap-y-1 flex-wrap">
         <div
           className={cn("w-auto border dark:border-zinc-800 pl-1.5 pr-2 py-0.5 rounded-md flex justify-between items-center")}
-          onClick={() => memoFilterStore.addFilter({ factor: "property.hasLink", value: "" })}
+          onClick={() => memoFilterStore.addFilter({ factor: "property.hasLink", value: t("filters.has-link") || "" })}
```
johnnyjoygh, line 98: We should handle it [here](https://github.com/usememos/memos/blob/main/web/src/components/MemoFilters.tsx#L24), using a switch or something else that should works better.

johnnyjoygh (changes requested):

web/src/components/StatisticsView.tsx
taosin: I completely agree with your statement and have made the necessary adjustments. Please confirm.

web/src/components/MemoFilters.tsx
```diff
@@ -22,7 +24,8 @@ const MemoFilters = () => {
       return filter.value;
     }
     if (filter.factor.startsWith("property.")) {
-      return filter.factor.replace("property.", "");
+      const factorLabel = filter.factor.replace("property.", "");
```
johnnyjoygh, line 27: As the property factors are enumerable, how about using `switch` to avoid change [`TranslationType`](https://github.com/usememos/memos/pull/4488/files#diff-b86dfa196026c60adc127b84e168d6ce8fcdbcc0982f693a2422bf94096e47cbR43)?

taosin: yes, resolved it

johnnyjoygh (approved):
LGTM, fix the linter error then

johnnyjoygh (approved):
LGTM, thanks!

PR #4490, opened 2025-03-11
Adds ids and classes for easier CSS/JS selectors

### LINKED ISSUE #4481: Ids & classes to make it easier to customize Memos with CSS and Javascript

It would be great if the `div` for each memo/note contained a unique id and classes to make it easier to target them.

Right now this is how the `div` container for a note renders:
```html
<div class="group relative flex flex-col justify-start items-start w-full px-4 py-3 mb-2 gap-2 bg-white dark:bg-zinc-800 rounded-lg border border-white dark:border-zinc-800 hover:border-gray-200 dark:hover:border-zinc-700">
  [note]
</div>
```

I'd like it if the note included a unique id and classes like "memo" and "tagged-warning" for easier targeting. It would be great if the div looked something like this:

```html
<div id="memo-123456" class="memo tagged-personal tagged-warning tagged-web-dev …">
  <div class="memo-header"><div>
  <div class="memo-body"></div>
</div>

```

Then it would be easy to customize notes with CSS and Javascript like these:

```css
.memo {
  background: lightgray;
}

.tagged-warning {
  border-left: 5px solid red;
}
```

```javacript
let memo123456 = document.getElementById('memo-123456');

let memos = document.querySelectorAll('.memo');
let warnings = document.querySelectorAll('.tagged-warning');
```

### Type of feature

Other

### Additional context

_No response_

samglover:
This PR adds ids and classes for easier CSS/JS selectors.

Added ids:
* #memo-list
* #memo-comments
* #home-sidebar

Added classes:
* .memo
  * .tagged-[tag] for all tags
  * .memo-header
  * .memo-content
* .memo-detail-sidebar

Closes #4481

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -86,5 +86,6 @@
     "onlyBuiltDependencies": [
       "@bufbuild/buf"
     ]
-  }
+  },
+  "packageManager": "pnpm@10.6.2+sha512.47870716bea1572b53df34ad8647b42962bc790ce2bf4562ba0f643237d7302a3d6a8ecef9e4bdfc01d23af1969aa90485d4cebb0b9638fa5ef1daef656f6c1b"
 }
--- a/web/src/components/HomeSidebar/HomeSidebar.tsx
+++ b/web/src/components/HomeSidebar/HomeSidebar.tsx
@@ -71,7 +71,7 @@ const HomeSidebar = observer((props: Props) => {
   );
 
   return (
-    <aside className={cn("relative w-full h-full overflow-auto flex flex-col justify-start items-start", props.className)}>
+    <aside id="home-sidebar" className={cn("relative w-full h-full overflow-auto flex flex-col justify-start items-start", props.className)}>
       <SearchBar />
       <div className="mt-2 w-full space-y-1">
         {navLinks.map((navLink) => (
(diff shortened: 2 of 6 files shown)


samglover:
I'm getting a formatting error, but I'm not sure there's actually a problem. And part of the error relates to existing code that I didn't feel comfortable modifying.

johnnyjoygh:
> I'm getting a formatting error, but I'm not sure there's actually a problem. And part of the error relates to existing code that I didn't feel comfortable modifying.

Try run `pnpm lint --fix` in your dev environment to solve these format errors.

johnnyjoygh:
@samglover We should ideally standardize to a specific attribute, such as `data-classname`, instead of mixing in `id` and `class`. e.g.,

```html
<div data-classname="header-sidebar"></div>
```

samglover:
I'm not sure I understand why you'd want to add a data attribute instead of just adding ids and classes. Just makes it harder to write selectors for CSS. `.memo-header` is regular CSS and easier to write than `[data-classname="memo-header"]`.

johnnyjoygh:
@samglover Since memos uses Tailwind, the `className` is already occupied by things like `px-2`, so it shouldn't be a good place to save custom class. Additionally, IDs should not conflict, so a better approach is to place it in a `data-` attribute, and to avoid breaking the existing code style. 

This is an open-source project, and I think that code maintainability is more important than the the convenience of code for custom styles.

samglover:
Fair enough. I just want to be able to actually use the CSS and JS customizations. Right now they're not very useful because it's not possible to reliably target memos or memo content elements. 

If someone wants to take over this PR and change it to use data attributes instead of ids and classes I can work with that.

PR #4497, opened 2025-03-13
chore(i18n): tidy

lincolnthalles:
- sort all keys alphabetically for easier side-by-side translation

- remove keys from other locales that don't exist anymore in `en.json`

- remove keys from other locales that have missing {{interpolations}}

- add missing entries to pt-BR


boojack (approved):
LGTM, thanks!

PR #4504, opened 2025-03-14
fix: user stats changes based on selected tab

### LINKED ISSUE #4500: Tags don't update when switching between Home and Explore
### Describe the bug

When switching between the Home and Explore pages, the tags don't update to reflect the memos available. It works once when switching from Home to Explore, but not back the other way. The page needs to be refreshed for this to take effect.

### Steps to reproduce

Have different tags on private and public memos, or other users that use different tags on their posts.
Switch from Home to Explore page, the tags on the left side update.
Switch back to Home and the tags remain the same, leaving tags visible that don't have any memos attached.

v0.24.1

_No response_

ehsundar:
- set partial only the relevant user stats instead of only adding up stats

closes #4500

### DIFF
--- a/web/src/store/v2/user.ts
+++ b/web/src/store/v2/user.ts
@@ -179,7 +179,6 @@ const userStore = (() => {
     }
     state.setPartial({
       userStatsByName: {
-        ...state.userStatsByName,
         ...userStatsByName,
       },
     });


ehsundar:
Not sure why state change was only additive 🫤. Tested the front a little bit and seems it hasn't broken anything yet

johnnyjoygh (approved):

PR #4506, opened 2025-03-14
Build process automation with makefile

Cadavanaugh:
I'm new to contributing to Memos and the build process was quite a fricction for the fresh start of developing new features.
I automated this simple, yet annoying and repetitive process, by creating a makefile that do the job for us.
**Make** is a software that comes preinstalled in most linux flavors (such as Ubuntu, Debian, Fedora, Red Hat, CentOS, Arch Linux, openSUSE etc) and it's widely used in the IT industry as a tool for automating scripts and therefore building processes. This change will reduce the mental workload for new and current contributors.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+.PHONY: build_backend backend frontend
+
+build_backend:
+	sh scripts/build.sh
+
+backend: build_backend
+	./build/memos --mode dev
+
+frontend:
+	cd web && pnpm i && pnpm dev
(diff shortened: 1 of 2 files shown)


ehsundar:
Make is a limited/obsolete build system for C/C++ projects. It describes how a missing object can be built to be used by the linker so that it constructs the executable. Following the principle, you should have named the `build_backend` -->  `build/memos` to construct the missing object accompanying a `clean`. However, I've found the Makefile irrelevant in Go or Python projects as it comes with additional complexity over time, and adds nothing to the project. I personally believe that .sh files serve the purpose best and they run anywhere without the need of learning extra syntax.

Cadavanaugh:
Besides all the cons you mentioned don’t you think it’s a good workaround to simplify the build process by reducing it to a single command? What other tools you suggest we use to achieve the same goal?

ehsundar:
Why not `scripts/build.sh && build/memos --mode=dev`. Don't see the necessity to add a new dependency to the project.

PR #4508, opened 2025-03-14
Create webpack.yml

rumman157:


ehsundar:
@boojack
Pretty damn dangerous! Hope you don't have secrets exposed there :))). I think you should avoid running arbitrary workflows by moving it to another repo.

boojack:
@rumman157 thanks for the attention, I will check the permission later.

PR #4514, opened 2025-03-15
fix: omit empty `next_page_token`

### LINKED ISSUE #4512: Next page token presence does not behave as documented
### Describe the bug

The [documentation](https://github.com/usememos/memos/blob/6b042fe16eea42506190e761ff5b9a81308643f0/proto/gen/apidocs.swagger.yaml#L2794C1-L2797C10) says this:

> | nextPageToken | string | A token, which can be sent as `page_token` to retrieve the next page. If this field is omitted, there are no subsequent pages. | No |

When I make a ListMemos request that will have no next page, e.g. `curl https://memos.example.com/api/v2/memos?pageSize=50&filter=row_status%20%3D%3D%20'NORMAL'` , the returned page token is an empty string rather than being absent:

```
{..."resources":[], "relations":[], "reactions":[]}], "nextPageToken":""}
```

The documentation implies that the field will be absent. This is incongruent.

Personally I think the fix should either be ommitting the field or setting it to `null` when there is no extra page. Empty string is not an acceptable stand-in for `null`.

### Steps to reproduce

Run a ListMemos query with a page size that is higher than the number of memos in your instance.

v0.24.1

_No response_

ehsundar:
closes #4512

I'm quite sure that changing to optional without changing the id is wire compat. But I decided to reassure the idea with the following piece:

```go
	r, _ := proto.Marshal(response)
	sh := sha256.New()
	sh.Write(r)
	slog.Info(hex.EncodeToString(sh.Sum(nil)))
```

With optional field:
`
2025/03/15 03:25:58 INFO 2a150f54f322b23ffae9ae7aaec39fcd28a0c6df7d20cd67395d04be8c007c51
`

Without it:
`
025/03/15 03:27:57 INFO 2a150f54f322b23ffae9ae7aaec39fcd28a0c6df7d20cd67395d04be8c007c51
`


johnnyjoygh:
> The next_page_token field, which supports pagination, must be included on all list response messages. It must be set if there are subsequent pages, and must not be set if the response represents the final page. For more information, see AIP-158.
> 
> https://github.com/aip-dev/google.aip.dev/blob/master/aip/general/0132.md#response-message

@ehsundar According to Google AIP, the `next_page_token` should not be optional.

radiozradioz:
Ok, if the field must be present, we should make it nullable

ehsundar:
Hmm. I also read AIP158 regarding pagination and you're right. It says 

> If the end of the collection has been reached, the next_page_token field must be empty. This is the only way to communicate "end-of-collection" to users.

So nullable is irrelevant and I close the PR accordingly.

radiozradioz:
@ehsundar @johnnyjoygh 

I don't entirely agree that "must be empty" necessarily means "must be an empty string". In any case, I think an empty string is a bad way to represent this state.

I agree that we should keep the field present in all responses, in line with AIP-132, because changing a field value is more ergonomic than changing the response structure (for example, if the client is in a statically typed language).

So we want to use the field to represent the fact that there is no next page. Here is what I think is meant by the two approaches:

- Nullable field: There is no next page, so there is no next page token.
- Empty string: There is no next page, so we give the user a special `""` next page token that they should interpret as meaning there is no next page.

The user needs to check the value either way, I think the nullable field far more well-founded.

AIPs are just guidelines after all, not standards. For example Google's own Ad Manager API omits the `nextPageToken` field entirely: https://developers.google.com/ad-manager/api/beta/pagination . So this is definitely up to interpretation and I think nullability is the better approach.

PR #4520, opened 2025-03-15
Clocktime of created memos

Cadavanaugh:
When reading old memos I didn't had the vision of when I did that specific post because the display date only showed weekday followed by date (Sun, Mar 9). Now the display time also shows the clocktime of memo creation (Sun, Mar 9, 1:40 PM) while keeping the relative time for newest memos such as "1 hour ago". Different time formats for different countries is already accomplished via i18n.

(en-US time)
image

(pt-BR time)
image

### DIFF
--- a/web/src/components/MemoContent/EmbeddedContent/EmbeddedMemo.tsx
+++ b/web/src/components/MemoContent/EmbeddedContent/EmbeddedMemo.tsx
@@ -80,7 +80,7 @@ const EmbeddedMemo = ({ resourceId: uid, params: paramsStr }: Props) => {
               minute: "numeric",
             })
           ) : (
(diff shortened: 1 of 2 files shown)


ehsundar:
There's no need to implement such feature as it's been already implemented. If you hover the date, the tooltip would show you the exact moment. It is also available in the PWA through the sidebar menu.

It's worth mentioning that I haven't read any code to know such feature exists as it is a common design practice and respected in many application.

Cadavanaugh:
Don’t you agree that showing the clocktime right after the date it’s just better UX? Honestly I didn’t even know about the tooltip nor the sidebar. That’s why I wanted to code this feature.

PR #4521, opened 2025-03-15
Preventing future package manager error

Cadavanaugh:
New contributors like me might miss the package manager and end up running `npm install` instead of `pnpm install`. Which generates a `package-lock.json` that should not be uploaded to the remote repository since the correct lockfile is `pnpm-lock.yaml`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,3 @@ dist
 
 # VSCode settings
 .vscode
-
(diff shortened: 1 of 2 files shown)


ehsundar:
If it is to be merged, I think the suitable .gitignore to change is the one inside the /web directory. See https://github.com/usememos/memos/blob/main/web/.gitignore

PR #4532, opened 2025-03-17
Fix: Allow admins to uncheck disallow password if applicable

### LINKED ISSUE #4469: Password auth is not allowed.
### Describe the bug

I did not enable SSO login, nor did I disable password login, but the login page shows "Password auth is not allowed."

### Steps to reproduce

Image
Image
This is a screenshot from when I was able to log in; now I can no longer access it.

v0.24.0

_No response_

ehsundar:
- closes #4469

### DIFF
--- a/web/src/components/Settings/WorkspaceSection.tsx
+++ b/web/src/components/Settings/WorkspaceSection.tsx
@@ -128,7 +128,10 @@ const WorkspaceSection = () => {
       <div className="w-full flex flex-row justify-between items-center">
         <span>{t("setting.workspace-section.disallow-password-auth")}</span>
         <Switch
-          disabled={workspaceStore.state.profile.mode === "demo" || identityProviderList.length === 0}
+          disabled={
+            workspaceStore.state.profile.mode === "demo" ||
+            (identityProviderList.length === 0 && !workspaceGeneralSetting.disallowPasswordAuth)
+          }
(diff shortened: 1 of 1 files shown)


web/src/components/Settings/WorkspaceSection.tsx
```diff
@@ -128,7 +128,10 @@ const WorkspaceSection = () => {
       <div className="w-full flex flex-row justify-between items-center">
         <span>{t("setting.workspace-section.disallow-password-auth")}</span>
         <Switch
-          disabled={workspaceStore.state.profile.mode === "demo" || identityProviderList.length === 0}
+          disabled={
+            workspaceStore.state.profile.mode === "demo" ||
+            (identityProviderList.length === 0 && !workspaceGeneralSetting.disallowPasswordAuth)
```
boojack, line 133: Disabling password login requires setting up SSO first; otherwise, there would be no way to log in. So I think the logic here is correct.

ehsundar: Don't you agree that one should be able to enable password login regardless?

I reproduced the erroneous scenario and mentioned it in the issue:
1- Add an OAuth login
2- Disallow passwords
3- Remove OAuth
4- You can no longer enable password login as the switch is always disabled

boojack: Gocha

boojack (approved):

PR #4547, opened 2025-03-20
MemoView support double click to edit

taosin:
MemoView support double click to edit  **like flomo**


johnnyjoygh:
You can double-click the content to edit.

taosin:
> You can double-click the content to edit.

You are right, I found double-click the content to edit. 

Thanks again!

PR #4566, opened 2025-03-27
feat: update ItalicNode to use repeated children instead of content

singhanubhavme:
This PR updates the markdown rendering logic to align with recent changes in the gomark package, specifically the refactored Italic node structure.

The upstream parser now uses a child-based representation (Children []Node) for ast.Italic instead of a simple Content string. This allows for nested inline elements within italic text (e.g., links inside italic).

Key Changes:
1. Updated rendering logic to handle the new Italic.Children structure
2. Adjusted any logic that previously relied on Italic.Content

Related PR:
Companion PR in the gomark repo (must be merged first):
👉 [Refactor Italic Node Structure for Nested Inline Support](https://github.com/usememos/gomark/pull/20)


johnnyjoygh:
@singhanubhavme you need to upgrade gomark version in this PR to avoid linter errors.

johnnyjoygh (approved):
LGTM

PR #4596, opened 2025-04-02
Fix hashtag parsing

snake-lvyonghao:
Fixes an issue where hashtags incorrectly include trailing punctuation such as ,, :, and .... Now hashtags are parsed correctly, ignoring special characters at the end.

Also adds support for multi-language and emoji tags like #标签 and #🚀go.


nick16180:
Does this issue address situation where a tag is created because of a hex code (like `#000000;`)? Would it be possible for it to not create tags from within code blocks as well?

PR #4635, opened 2025-04-17
feat: set locale based on browser language when user is not logged in

### LINKED ISSUE #4557: Hardcoded language when logged out
### Describe the bug

There [seems to be](https://github.com/usememos/memos/blob/main/web/src/store/v2/workspace.ts#L11) a hardcoded language for logged out visitors set to `"en"`, and I cannot find a way to set it to a different language from the settings, and also it doesn't try if browser language is supported. It just goes to English no matter what.

I tried a custom javascript snippet:

```javascript
localStorage.setItem("locale", `"${navigator.language}"`);
```

And this doesn't work. First time visitors only get to see their language after a second refresh of the page. Probably this script gets executed too late. It should be executed *before* the site reads this localStorage value, but it gets executed afterwards.

Besides, this probably won't work well if the language of the browser doesn't have a translation in Memos. Not tested though.

The least I could probably hope for at this point, is an environment variable, if not a setting in the GUI. But that's still a bit naff, because I don't want to impose *any single language* upon my visitors. Why would I? That's what the available translations are for.

### Steps to reproduce

Visit the site anonymously, with a browser that requests websites in a language other than English.

Actual result: website still displays in English
Expected result: try a supported matching language, and then revert to English as a last resort. And even that should be configurable.

0.24.0

NB: I'm calling this a bug, because it's a hardcoded user preference. That is a bug. It'd be a feature request if translations weren't a thing at all yet.

Alwayswithme:
### Description
This PR adds functionality to detect and set the locale based on the browser's language settings (`navigator.language`) when the user is not logged in.
Let me know if anything else needs to be updated. Thanks!

### Related Issue
Closes #4557

### DIFF
--- a/web/src/store/v2/user.ts
+++ b/web/src/store/v2/user.ts
@@ -224,6 +224,7 @@ export const initialUserStore = async () => {
       appearance: userSetting.appearance,
     });
   } catch {
+    // find the nearest matched lang based on the `navigator.language` if the user is unauthenticated or settings retrieval fails.
     const locale = findNearestMatchedLanguage(navigator.language);
     workspaceStore.state.setPartial({
       locale: locale,
--- a/web/src/store/v2/user.ts
+++ b/web/src/store/v2/user.ts
@@ -3,6 +3,7 @@ import { makeAutoObservable } from "mobx";
 import { authServiceClient, inboxServiceClient, userServiceClient } from "@/grpcweb";
 import { Inbox } from "@/types/proto/api/v1/inbox_service";
 import { Shortcut, User, UserSetting, UserStats } from "@/types/proto/api/v1/user_service";
+import { findNearestMatchedLanguage } from "@/utils/i18n";
 import workspaceStore from "./workspace";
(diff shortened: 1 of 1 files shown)


web/src/store/v2/user.ts
```diff
@@ -223,7 +224,10 @@ export const initialUserStore = async () => {
       appearance: userSetting.appearance,
     });
   } catch {
-    // Do nothing.
+    const locale = findNearestMatchedLanguage(navigator.language);
```
johnnyjoygh, line 227: Add a comment about how to reach this case. Maybe "For unauth users, we'll find the nearest matched lang from navigator language."

johnnyjoygh (approved):
LGTM

web/src/store/v2/user.ts
Alwayswithme: Thanks, comment added

johnnyjoygh (approved):
LGTM

PR #4691, opened 2025-05-14
Move memo filters component to the bottom of the sidebar

wenyg:
### **Optimized Layout Proposal for MemoFilter**  

We recommend relocating the **MemoFilter** beneath the **Tags** component for the following reasons:  

1. **Enhanced Usability**: The new positioning reduces the spatial distance between MemoFilter and Tags, improving operational efficiency.  
2. **Layout Consistency**: This adjustment ensures the Sidebar’s structure remains unaffected by the presence or absence of filters, maintaining a stable and uniform interface.  

https://github.com/user-attachments/assets/252a30fd-4115-42e6-b764-91c9c9687a7e

### DIFF
--- a/web/src/components/HomeSidebar/HomeSidebar.tsx
+++ b/web/src/components/HomeSidebar/HomeSidebar.tsx
@@ -90,12 +90,12 @@ const HomeSidebar = observer((props: Props) => {
           </NavLink>
         ))}
       </div>
-      <MemoFilters />
       <div className="px-2 w-full">
         <StatisticsView />
         {currentUser && <ShortcutsSection />}
         <TagsSection />
       </div>
+      <MemoFilters />
     </aside>
   );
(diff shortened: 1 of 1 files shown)


johnnyjoygh:
If there are many tags, the filter will appear at the very bottom and require scrolling to be visible. It's unexpected.

wenyg:
> If there are many tags, the filter will appear at the very bottom and require scrolling to be visible. It's unexpected.

Thank you for your response. I'm currently facing a key pain point regarding frequent switching between tags when searching for content. The typical workflow is:

1. Select Tag A, fail to locate target content, then deselect Tag A
2. Select Tag B, still can't find the content, then deselect Tag B
3. ......

I'm hoping to improve the tag-switching experience, or explore whether there might be more efficient solutions available.

PR #4699, opened 2025-05-18
Hide "or" when password auth disabled

RealOrangeOne:
This makes the UI cleaner, instead only showing the "Sign in with ..." buttons.


johnnyjoygh (approved):
LGTM

Issue #4683, opened 2025-05-10
Infinite scroll


Cadavanaugh:
### Describe the solution you'd like

Automatic "load more" trigger when timeline reaches the end.

### Type of feature

User Experience (UX)

elonpart2:
what if i can done this

PR #4715, opened 2025-05-22
feat: add infinite scrolling for memos

liltrendi:
## Overview

For users with too many notes, it feels unintuitive to keep pressing the "Load More" button. This PR aims to ease navigation by allowing the user to opt in to infinite scrolling, whereby the next posts (if any) are fetched as soon as the user reaches the bottom of the screen.

References: #4683 

### Changes

- Implemented infinite scrolling for memos in the home, archive, explore and profile pages
- Added a dropdown under user preferences (in settings) for users to opt in or out of this feature
- Persistence on the backend

### Demo

https://github.com/user-attachments/assets/3b04f8ec-58c2-4ed5-9f3f-fc9a69c0dfa6

### DIFF
--- a/proto/api/v1/user_service.proto
+++ b/proto/api/v1/user_service.proto
@@ -122,6 +122,11 @@ service UserService {
   }
 }
 
+enum ScrollMode {
+  BUTTON = 0;
+  INFINITE = 1;
+}
+
 message User {
   // The name of the user.
   // Format: users/{id}, id is the system generated auto-incremented id.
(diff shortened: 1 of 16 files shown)


proto/api/v1/user_service.proto
```diff
@@ -244,6 +249,8 @@ message UserSetting {
   string appearance = 3;
   // The default visibility of the memo.
   string memo_visibility = 4;
+  // Whether infinite scroll should be enabled
+  ScrollMode scroll_mode = 5;
```
johnnyjoygh, line 253: I think we can directly use infinite scroll loading, so there's no need to add a setting definition.

johnnyjoygh (approved):

PR #4716, opened 2025-05-23
fix(editor): Prevent race condition and ensure correct list continuation on Enter

jonfk:
https://github.com/user-attachments/assets/c029120f-956e-4384-9b0e-5e97e45dfb5a

I am not always able to reproduce it. But this often happens on iOS Safari on my iPhone 13 Pro and iPad Pro M2. 

From what I can tell, it seems there is a race condition between the default behaviour of the event handler and the `insertText` to be inserted. It only seems to happen with the task list continuation, maybe because it is slightly longer than the other list continuations.

This fix removes the race condition and inserts the newline manually before the `insertText`. 
After implementing this fix, I don't see this occurring on my devices anymore.

Thanks for considering the fix.

### DIFF
--- a/web/src/components/MemoEditor/Editor/index.tsx
+++ b/web/src/components/MemoEditor/Editor/index.tsx
@@ -169,6 +169,10 @@ const Editor = forwardRef(function Editor(props: Props, ref: React.ForwardedRef<
       if (event.shiftKey || event.ctrlKey || event.metaKey || event.altKey) {
         return;
       }
+      // Prevent a newline from being inserted, so that we can insert it manually later.
+      // This prevents a race condition that occurs between the newline insertion and
+      // inserting the insertText.
+      event.preventDefault()
 
       const cursorPosition = editorActions.getCursorPosition();
       const prevContent = editorActions.getContent().substring(0, cursorPosition);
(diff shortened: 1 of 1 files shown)


web/src/components/MemoEditor/Editor/index.tsx
```diff
@@ -194,7 +198,7 @@ const Editor = forwardRef(function Editor(props: Props, ref: React.ForwardedRef<
         insertText += `${Number(number) + 1}. `;
       }
       if (insertText) {
-        editorActions.insertText(insertText);
+        editorActions.insertText("\n" + insertText);
```
boojack, line 201: Is `preventDefault` supposed to be here?

```suggestion
        event.preventDefault();
        editorActions.insertText("\n" + insertText);
```

jonfk: Good catch, actually missed this from some of my experiments. The reason we want to have `event.preventDefault` at the top of the function is so that we prevent the default browser behaviour before any async operations otherwise, we will get the aforementioned race condition. 

Pushed a change that simply removes the `if(insertText)` and always add the newline and insertText. That way, a newline will be inserted even if there are no insertText to insert.

Thanks for the review. Also really appreciate the work on memos, really cool program.

boojack (approved):
LGTM

PR #4717, opened 2025-05-23
feat(rss): Use server title and description for RSS feed, if configured

maximilian-krauss:
As much as I like memos, I think it's confusing for users if the title and description is different than the one which shown on the web ui.

This PR adds those in the meta of the RSS feed too:

CleanShot 2025-05-23 at 15 17 02@2x


boojack (approved):
LGTM

PR #4766, opened 2025-06-09
fix: trim queryText and split by regex

Yamico:
**Describe the bug**
After updating from version, an error occurs when searching.

**Steps to reproduce**
When entering text in the Search Bar, if there are leading/trailing spaces or multiple consecutive spaces between words, it causes an abnormal error in "ContentSearch" (as shown in the screenshot).

**The version of Memos you're using**
v0.24.4 (also in develop version)

**Screenshots or additional context**
anime1
anime2
anime3

The submitted code has fixed this bug—please review and merge.


Yamico:
Additionally, I suggest clearing all previous contentSearch results after each new search query. Would this be acceptable?

PR #4769, opened 2025-06-10
feat(security): Adds last usage date to access tokens

maximilian-krauss:
CleanShot 2025-06-10 at 13 37 43@2x

### What does this do?

This PR adds a last usage date to access tokens, because after login from several browsers and devices, it's hard to tell which access tokens are still being used.

It adds one write after each _authenticated_ API operation which could add stress to the database. I tested with sqlite and mysql without any notable performance issues. 

✌️


boojack:
@maximilian-krauss We are refactoring the auth logic, including designing a common session that will include information such as the last used time, etc. I'll let you know once it's done. Thanks for the contribution.

boojack:
Some updates: 

image

maximilian-krauss:
Nice! thanks for adding it 🙂

Issue #3095, opened 2024-03-13
No regex check on SignUp


Otter2576:
### Describe the bug

If you create an user via the SignUp Method, you could use any string for the username and it will be accepted.
On the other hand, creating a user in the Admin->Member menu will throw an error toast, if the regex is not matched.

I'm new to Go, so I could be wrong, but it seems,that the same regex check on the username in the CreateUser func is not used in the SignUp func.  

Maybe this could also be a security vulnerability, because you can put any one-liner into the username via the SignUp Method.
As I said, I am not familiar with Go, so maybe nothing to it.

### Steps to reproduce

1. As an admin, go to Settings&rarr;Member
2. Try to create a new member any one-liner you want (also including special signs, underscore, dots, spaces). 
3. A toast with "invalid username" should pop up.
4. Sign up as a new user
5. For the username, use the same one-liner as in step 2.
6. Your user will be created.

### The version of Memos you're using

0.20.1 on docker

### Screenshots or additional context

Possible username via SignUp:
![Screenshot from 2024-03-13 23-41-34](https://github.com/usememos/memos/assets/152623628/64753105-ae2f-4bcc-8dcc-339c8bdb2b05)

boojack:
Thanks for the reporting. It will be fixed with https://github.com/usememos/memos/commit/606a30640db3ac86e712c86c02d7c4d1e0139f66

PR #4770, opened 2025-06-11
Allow underscores and periods in usernames

EastArctica:
Super simple PR. Just extends the username validation a little by allowing underscores and periods anywhere except the ends of the username field. This stemmed from my desire for underscores and finding that [some people would also like to be able to use dots/periods](https://github.com/usememos/memos/issues/3095#issuecomment-1996040843) in their usernames as well so I tossed that in there too.

Is there any reason that they were excluded and only hyphens allowed? From my very brief testing they seem to work without issue.

### DIFF
--- a/internal/base/resource_name.go
+++ b/internal/base/resource_name.go
@@ -3,5 +3,5 @@ package base
 import "regexp"
 
 var (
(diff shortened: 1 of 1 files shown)


boojack:
> Is there any reason that they were excluded and only hyphens allowed? From my very brief testing they seem to work without issue.

Memos uses a very generic regex for username, and other platforms like GitHub do not allow `.` or `_` in usernames too. However, I think memos can support the characters you mentioned. But I suggest adding a dedicated matcher specifically for username.

Issue #4746, opened 2025-06-02
Unable to change the date and time of memos


hchengting:
### Describe the bug

After updating from version 0.24.3 to 0.24.4, an error occurs when changing the date and time of memos.

### Steps to reproduce

1. Create a memo.
2. Attempt to change the creation time.
3. An error appears: `Invalid datetime format. Use format: 2023-12-31T23:59:59`.

### The version of Memos you're using

v0.24.4

### Screenshots or additional context

![Image](https://github.com/user-attachments/assets/3c4488d8-b6f8-47d4-b385-d195d1131b01)

![Image](https://github.com/user-attachments/assets/eb0c69d3-3ba7-4058-a20e-8035058e2c3d)

liljohn360:
As a user in the UK, I've noticed that the day and month get swapped when editing the time in this new time editor too. Resulting in the Memo burying itself months into the past.

hchengting:
@boojack 
In commit https://github.com/usememos/memos/commit/6d319900d640df6da7e7646a14ea6bfbb2018fd9 the updated time can be modified, but for some reason, the created time cannot be edited.

https://github.com/user-attachments/assets/058736fa-7cf5-446b-8a12-573cb8f3e14b

hchengting:
> [@boojack](https://github.com/boojack) In commit [6d31990](https://github.com/usememos/memos/commit/6d319900d640df6da7e7646a14ea6bfbb2018fd9) the updated time can be modified, but for some reason, the created time cannot be edited.
> 
>  2025-06-03.141159.mp4

It seems that the created time always shows the current time.

boojack:
It should be fixed within https://github.com/usememos/memos/commit/94d1e205709f52a12d81358074ccbb2473d85e28. Thanks for the reporting.

PR #4829, opened 2025-07-08
fix: dayjs - ISO 8601 compatible DATE_TIME_FORMAT

unuunn:
According to dayjs documentation
the `dayjs(inputValue, DATE_TIME_FORMAT, true)` construct does not work without `CustomParseFormat` plugin, [link](https://day.js.org/docs/en/parse/string-format),
so the code contains an error.

Also, according to the documentation, dayjs develloper recommended to use ISO 8601 [link](https://day.js.org/docs/en/parse/string)
is international and better readable for the default format.

It's a continuation of the bug #4746 

Using ISO 8601 without `T` is acceptable in JS. 
I suggest to make it the main one for `DATE_TIME_FORMAT`
It preserves readability when editing the date of records and uses the classic international "big-endian" date format.

It would be ideal to put the preferred date format in the settings, but this requires more resources.

### DIFF
--- a/web/src/components/DateTimeInput.tsx
+++ b/web/src/components/DateTimeInput.tsx
@@ -2,7 +2,7 @@ import dayjs from "dayjs";
 import toast from "react-hot-toast";
 import { cn } from "@/lib/utils";
 
-// ISO 8601 (almost)
+// must be compatible with JS Date.parse(), we use ISO 8601 (almost)
 const DATE_TIME_FORMAT = "YYYY-MM-DD HH:mm:ss";
 
 // convert Date to datetime string.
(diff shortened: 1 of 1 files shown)


web/src/components/DateTimeInput.tsx
```diff
@@ -23,12 +24,25 @@ const DateTimeInput: React.FC<Props> = ({ value, onChange }) => {
...
         const inputValue = e.target.value;
         if (inputValue) {
-          const date = dayjs(inputValue, DATE_TIME_FORMAT, true).toDate();
+          const date = dayjs(inputValue).toDate();
           // Check if the date is valid.
           if (!isNaN(date.getTime())) {
             onChange(date);
           } else {
-            toast.error("Invalid datetime format. Use format: 12/31/2023, 23:59:59");
+            toast.error(
```
boojack: Could we use a simple error message instead? Maybe just an example of wanted datetime.

unuunn: no problem, adapted an old message and added important code notes

boojack: Thanks

boojack (approved):
LGTM

PR #4909, opened 2025-07-23
fix(settings): Enable pointer events in MemoRelatedSettings

maximilian-krauss:
This PR fixes the removal of reactions and NSFW tags which was not possible in Settings/Memo.

### DIFF
--- a/web/src/components/Settings/MemoRelatedSettings.tsx
+++ b/web/src/components/Settings/MemoRelatedSettings.tsx
@@ -126,7 +126,7 @@ const MemoRelatedSettings = observer(() => {
               <Badge key={reactionType} variant="outline" className="flex items-center gap-1">
                 {reactionType}
                 <X
-                  className="w-3 h-3 cursor-pointer hover:text-destructive"
+                  className="w-3 h-3 cursor-pointer hover:text-destructive !pointer-events-auto"
                   onClick={() => updatePartialSetting({ reactions: memoRelatedSetting.reactions.filter((r) => r !== reactionType) })}
                 />
               </Badge>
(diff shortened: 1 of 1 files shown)


web/src/index.css
```diff
@@ -18,3 +18,7 @@
     @apply bg-background text-foreground;
   }
 }
+
+.pointer-events-auto {
```
johnnyjoygh: No need to add this as we're using tailwind. https://tailwindcss.com/docs/pointer-events

maximilian-krauss: That's what I thought too, but adding only the class `pointer-events-auto` had no effect

web/src/components/Settings/MemoRelatedSettings.tsx
```diff
@@ -126,7 +126,7 @@ const MemoRelatedSettings = observer(() => {
               <Badge key={reactionType} variant="outline" className="flex items-center gap-1">
                 {reactionType}
                 <X
-                  className="w-3 h-3 cursor-pointer hover:text-destructive"
+                  className="w-3 h-3 cursor-pointer hover:text-destructive pointer-events-auto"
```
boojack:
```suggestion
                  className="w-3 h-3 cursor-pointer hover:text-destructive !pointer-events-auto"
```

maximilian-krauss: ❤️ thanks, fixed it in both places

johnnyjoygh:
I found that SVG has `pointer-events: none` by default, so we can use a `span` as the container to completely solve this issue. https://github.com/usememos/memos/commit/5074268f4fc8dfbc537db6c3e85fc5dca7c9bc6c

Screenshot 2025-07-29 at 22 11 28

Issue #4928, opened 2025-07-29
Support for an alternate host for the "Copy Link" button?


clnhlzmn:
### Describe the solution you'd like

This is related to [my other question](https://github.com/issues/created?issue=usememos%7Cmemos%7C4923) but more fine tuned.

I'm working on a ["public proxy" server](https://github.com/clnhlzmn/memos-public-proxy) that would work with the memos API like the way [immich-public-proxy](https://github.com/alangrainger/immich-public-proxy) works with the immich API. One nice thing about immich is there is an ["external domain"](https://immich.app/docs/administration/system-settings#external-domain) setting that allows me to create share links with a different host than what is being used to access immich. For example, if I serve immich on photos.private.example.com and I have external domain set to photos.public.example.com then I can serve immich-public-proxy on the public domain and have the main immich server fully isolated on my LAN.

I'd like to do something similar with memos but the problem is when I click on "Copy Link" for a memo I get memos.private.example.com and then to share that link with someone who is not on my LAN I'd have to modify it to instead point to memos.public.example.com. I'd prefer that clicking "Copy Link" instead results in memos.private.example.com directly.

Are you aware of any way that I could use the custom JavaScript functionality to achieve this? I'd even be ok with adding a different button like "Copy Share Link" to do it instead of somehow overriding the existing functionality. 

Alternatively I could make a PR to add this to memos itself perhaps through an environment variable, but honestly I don't know where to start with that or if it would be accepted.

### Type of feature

User Interface (UI)

PR #4930, opened 2025-07-30
Use MEMOS_INSTANCE_URL for "Copy Link"

clnhlzmn:
This relates to my question [here](https://github.com/usememos/memos/issues/4928). Basically I was wondering if memos could support an alternate host for the "Copy Link" button. I noticed a mention of the environment variable `MEMOS_INSTANCE_URL` and as far as I can tell it isn't really used for anything.. So in this PR I have used it for this purpose.

I don't usually work with typescript or go so please forgive me if I am making some egregious error (though this is a pretty simple change).

~I could see an argument that the behavior should be maintained to copy `window.location.origin` in the case that `MEMOS_INSTANCE_URL` is unset so I think I'll see if I can make that work, but as far as I can tell from my quick testing this does exactly what I want as is.~ --------------did this

Now.. if there is some reason that `MEMOS_INSTANCE_URL` should not be used for this I'm happy to try to figure out how to add a different variable for it.

### DIFF
--- a/web/src/components/MemoActionMenu.tsx
+++ b/web/src/components/MemoActionMenu.tsx
@@ -17,6 +17,7 @@ import { markdownServiceClient } from "@/grpcweb";
 import useNavigateTo from "@/hooks/useNavigateTo";
 import { memoStore, userStore } from "@/store";
 import { State } from "@/types/proto/api/v1/common";
+import { workspaceStore } from "@/store";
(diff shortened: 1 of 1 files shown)


johnnyjoygh (approved):
LGTM

clnhlzmn:
@johnnyjoygh do you know when the next release might be made? I can use the canary image until then, just curious.

Issue #4937, opened 2025-07-31
tag filters not working on latest


clnhlzmn:
### Describe the bug

Using the image from commit [6005d0f](https://github.com/usememos/memos/commit/6005d0f3f786bce0a8e11cf8916b91c57e3b897f) when selecting a tag to filter memos in the browser console I see `Uncaught (in promise) ClientError: /memos.api.v1.MemoService/ListMemos INTERNAL: failed to list memos: pq: could not determine data type of parameter $2` and in the database (postgres:16.1) log I see 

```
2025-07-31 04:16:44.329 ERROR:  could not determine data type of parameter $2		
2025-07-31 04:16:44.329 STATEMENT:  SELECT memo.id AS id, memo.uid AS uid, memo.creator_id AS creator_id, memo.created_ts AS created_ts, memo.updated_ts AS updated_ts, memo.row_status AS row_status, memo.visibility AS visibility, memo.pinned AS pinned, memo.payload AS payload, memo_relation.related_memo_id AS parent_id, memo.content AS content
2025-07-31 00:16:44.329 FROM memo
2025-07-31 00:16:44.329 LEFT JOIN memo_relation ON memo.id = memo_relation.memo_id AND memo_relation.type = 'COMMENT'
2025-07-31 00:16:44.329 WHERE 1 = 1 AND ((memo.creator_id = $1 AND memo.payload->'tags' @> jsonb_build_array($2))) AND ((memo.creator_id = $3 OR memo.visibility IN ($4,$5))) AND memo.row_status = $6 AND memo_relation.related_memo_id IS NULL
2025-07-31 00:16:44.329 ORDER BY created_ts DESC LIMIT 17 OFFSET 0
```

### Steps to reproduce

1. use image from commit [6005d0f](https://github.com/usememos/memos/commit/6005d0f3f786bce0a8e11cf8916b91c57e3b897f)
2. use postgres:16.1 db
3. create memo with tag
4. click on tag in sidebar

### The version of Memos you're using

latest

PR #4941, opened 2025-07-31
fix tag filters maybe?

clnhlzmn:
I hate making issues without providing any help fixing them so this is my attempt at fixing [this one](https://github.com/usememos/memos/issues/4937).

Now.. this seems like a bandaid and not the _real_ solution? I don't actually know, but it _does_ seem to make the problem go away in my testing locally.

### DIFF
--- a/plugin/filter/templates.go
+++ b/plugin/filter/templates.go
@@ -35,12 +35,12 @@ var SQLTemplates = map[string]SQLTemplate{
 	"json_contains_element": {
 		SQLite:     "JSON_EXTRACT(`memo`.`payload`, '$.tags') LIKE ?",
 		MySQL:      "JSON_CONTAINS(JSON_EXTRACT(`memo`.`payload`, '$.tags'), ?)",
-		PostgreSQL: "memo.payload->'tags' @> jsonb_build_array(?::text)",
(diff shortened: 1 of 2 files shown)


plugin/filter/templates.go
```diff
@@ -35,12 +35,12 @@ var SQLTemplates = map[string]SQLTemplate{
 	"json_contains_element": {
 		SQLite:     "JSON_EXTRACT(`memo`.`payload`, '$.tags') LIKE ?",
 		MySQL:      "JSON_CONTAINS(JSON_EXTRACT(`memo`.`payload`, '$.tags'), ?)",
-		PostgreSQL: "memo.payload->'tags' @> jsonb_build_array(?)",
+		PostgreSQL: "memo.payload->'tags' @> jsonb_build_array(?::text)",
```
clnhlzmn: Side question: does this need to be changed? It seems like this isn't actually used anywhere.

Can we delete this (and `GetSQL`)?

maximilian-krauss:
Hey @clnhlzmn, the filter was also broken when using mysql. I fixed both in https://github.com/usememos/memos/pull/4944 and borrowed the fix you did 🙂

Issue #4811, opened 2025-07-01
Compile error on Raspberry Pi 4B: unaligned 64-bit atomic operation


Aldenar:
### Describe the bug

Hello,

I've been trying to compile the app on my RaspberryPI 4B, without success. 

Both, the available Docker image and manually launching the backend via the provided development instructions lead to the following crash:
```
aldar@serf:~/memos $ go run ./bin/memos/main.go --mode release --port 8083
panic: unaligned 64-bit atomic operation

goroutine 1 [running]:
internal/runtime/atomic.panicUnaligned()
	/usr/local/go/src/internal/runtime/atomic/unaligned.go:8 +0x24
internal/runtime/atomic.Xadd64(0x313b894, 0x1)
	/usr/local/go/src/internal/runtime/atomic/atomic_arm.s:318 +0x14
github.com/usememos/memos/store/cache.(*Cache).SetWithTTL(0x313b860, {0x1d62574, 0x3125c80}, {0x31c24a0, 0x5}, {0x12d40f8, 0x3298080}, 0x8bb2c97000)
	/home/aldar/memos/store/cache/cache.go:107 +0x158
github.com/usememos/memos/store/cache.(*Cache).Set(...)
	/home/aldar/memos/store/cache/cache.go:94
github.com/usememos/memos/store.(*Store).GetWorkspaceBasicSetting(0x3125ce0, {0x1d62574, 0x3125c80})
	/home/aldar/memos/store/workspace_setting.go:114 +0x1d8
github.com/usememos/memos/store.(*Store).updateCurrentSchemaVersion(0x3125ce0, {0x1d62574, 0x3125c80}, {0x328a0c0, 0x6})
	/home/aldar/memos/store/migrator.go:237 +0x2c
github.com/usememos/memos/store.(*Store).preMigrate(0x3125ce0, {0x1d62574, 0x3125c80})
	/home/aldar/memos/store/migrator.go:136 +0x380
github.com/usememos/memos/store.(*Store).Migrate(0x3125ce0, {0x1d62574, 0x3125c80})
	/home/aldar/memos/store/migrator.go:38 +0x50
main.init.func1(0x27a40f0, {0x3366a60, 0x0, 0x4})
	/home/aldar/memos/bin/memos/main.go:62 +0x448
github.com/spf13/cobra.(*Command).execute(0x27a40f0, {0x30c8008, 0x4, 0x5})
[...]

PR #4945, opened 2025-08-01
fix for issue 4811

einho:
Hello, first time contributor here! 

I was recently having issues using memos on my armbian powered, armv7 tv-box. 
However, I saw this issue was already discovered by #4811:
 
**- Error:** there are some 64bit operations which are performed on the cache struct which cause a *panic: unaligned* error on 32-bit architectures, e.g.
`atomic.AddInt64(&c.itemCount, 1)`

**-Fix description:** The [atomic library docs](https://pkg.go.dev/sync/atomic) already reports that the **Int64** type is automatically aligned, in contrast with the usage of **int64**. Thus, the fix changes *itemCount* into an Int64, applying the equivalent function on this type directly instead of calling the atomic library functions. Example: 
`atomic.AddInt64(&c.itemCount, 1)`
becomes
`(&c.itemCount).Add(1)`

**-Testing:** The application was successfully built with the armv7 target and worked both as a simple executable and inside the pre-built armv7 container, in which I swapped the broken executable for the new one. The x86 version kept working without issues

**-Extra:** Fun fact, as per the atomic documentation, in the bugs section:
> On ARM, 386, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically via the primitive atomic functions (types [Int64](https://pkg.go.dev/sync/atomic#Int64) and [Uint64](https://pkg.go.dev/sync/atomic#Uint64) are automatically aligned). 
**The first word in an allocated struct, array, or slice** [...] **can be relied upon to be 64-bit aligned.**

this technically means that the fix that copilot offered, which placed the *itemCount* variable at the beginning of the struct, is correct as the first word is automatically aligned. However, the documentation still suggests to use Int64 and related functions to avoid issues:
> Consider using the more ergonomic and less error-prone [Int64.Add](https://pkg.go.dev/sync/atomic#Int64.Add) instead (particularly if you target 32-bit platforms; see the bugs section).

### DIFF
--- a/store/cache/cache.go
+++ b/store/cache/cache.go
@@ -65,9 +65,9 @@ func DefaultConfig() Config {
 
 // Cache is a thread-safe in-memory cache with TTL and memory management.
 type Cache struct {
+	itemCount  atomic.Int64 // Use atomic operations to track item count
 	data       sync.Map
 	config     Config
-	itemCount  int64 // Use atomic operations to track item count
 	stopChan   chan struct{}
 	closedChan chan struct{}
 }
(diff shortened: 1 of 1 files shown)


johnnyjoygh (approved):
LGTM

Issue #4946, opened 2025-08-01
ListMemos endpoint worsening slow performance due to N+1 problem


varsnotwars:
### Describe the bug

The ListMemos method in the memo_service suffers from a n+1 problem causing significant degradation in performance once the amount of memos, relations, attachments, and/or comments grows. The response time will increase linearly with the number of memos and their related entities

With just 3 memos, linking each memo to each other, and adding a comment on each memo, I observed about 30 calls to the database. The response time from the api was just over 1 second. 

For each memo returned from the database, the method `convertMemoFromStore` is called passing in the memo resulting in multiple calls to retrieve all the related entities. From testing there are 9 calls to the database per memo. The methods called by 
`convertMemoFromStore` are:
- `ListMemoRelations`
- `ListMemoAttachments`
- `ListMemoReactions`

#### A possible approach to improve performance could be:

- Update the memo sql queries to join all related tables, which would probably require the biggest amount of work refactoring the calling methods and carry the highest risk.

- Batch the calls to the database, which may involve a similar amount of refactoring and risk.

- Introduce an internal service layer that works directly with memo objects exposing IDs/UIDs, which could eliminate some redundant lookups, though the n+1 problem would remain. (see note below for more detail)

#### Important note

There is also another aspect as to why the database calls per memo is so high; because the repo is following a resource-oriented design, if one of the nested calling methods needs the memo's ID, it has to query the database again using the memo's resource name as that is the only identifier in the memo request object. 3 of the 9 calls per memo is due to this.

If there was a second service layer that just dealt with the internal memo object that exposes the ID/UID field then those 3 calls could be removed but there would still be n+1 problem. 

Also it is worth noting that the 3 methods that make all of these database calls per memo aren't even public facing methods, there are no endpoints that expose these methods directly, yet they are restricted to taking the memo request object as an argument resulting in them needing to query the database to retrieve the memo to get the ID/UID field that they all need.

Regardless of which approach is taken the method and the queries in each database driver need to be refactored to avoid performance getting worse as the memos collection in the database grows.

See also https://github.com/usememos/memos/issues/4898

And possibly related to https://github.com/usememos/memos/issues/4899

### Steps to reproduce

1. Create 1 memo and observe the response time of the ListMemos api endpoint
2. Create more memos, add comments, link memos to each other, add attachments, add reactions, and observe the response time
3. Repeat adding more and more entities and observe the linear increase in response time.

### The version of Memos you're using

latest

PR #4969, opened 2025-08-05
feat: filter/method for reactions by content_id

varsnotwars:
Another PR to incrementally address #4946.

Implemented filter logic for `reaction` entities inline with `memo` and `user` filters.

The reactions table is now queryable using a filter. ListMemos queries the reactions table using a collection of memo names, reducing all calls to this table to just a single call. If there are 10 memos in the collection, 1 call is made to the database to retrieve the reactions, instead of 10.

`convertMemoFromStore` has fallback logic to query the DB if no reactions are passed in. The goal is to align `convertMemoFromStore` with a method that purely maps DB entities to response entities.

The `ListMemoReactions` public endpoint remains untouched. It is not publicly exposed by the router but it can be in the future.

Tested with sqlite, mysql, postgres.

### DIFF
--- a/plugin/filter/common_converter.go
+++ b/plugin/filter/common_converter.go
@@ -205,7 +205,7 @@ func (c *CommonSQLConverter) handleInOperator(ctx *ConvertContext, callExpr *exp
 		return err
 	}
 
-	if !slices.Contains([]string{"tag", "visibility"}, identifier) {
+	if !slices.Contains([]string{"tag", "visibility", "content_id"}, identifier) {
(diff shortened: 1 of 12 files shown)


varsnotwars:
@boojack, fixed linting error. Tests should pass now.

johnnyjoygh (approved):
LGTM. @boojack maybe we should do a refactor to the filter plugin. It's hard to review code changes.

boojack (approved):
LGTM

PR #4971, opened 2025-08-05
feat(editor): create text-based autocompleting commands

twaslowski:
Super new to this project, first contribution, please be gentle 😅 This is just a quick draft I hacked together as a proof-of-concept to get the community's opinion on the idea. I could not find anything on this in existing PRs or Issues, so if this is a feature that has been consciously omitted, I apologize.

image

I've tried out a bunch of different note-taking and memo apps and thought this project was super cool! After going through the documentation, I really found myself missing a simple commands feature to perform some repetitive tasks. For instance, I really dislike typing out the syntax for checkboxes, and I thought it would be neat if I could simply type /todo and get the checkbox to appear, much like e.g. Notion does it.

From a UX point-of-view, I imagined this very similarly to the TagSuggestions component, and so I reused a bunch of code, some of which could be extracted in the future. Right now, every command just corresponds to a `() => string` function that inserts into the Editor, but this could be made more powerful if there was demand for it.

Also, due to its sheer simplicity, this is something that could very easily be extended by users on an individual basis.


twaslows:
I additionally took the liberty of adding some more utilities for a more streamlined developer experience:

- Add: Taskfile.yaml for self-documenting lifecycle operations (test, build, run)
- Add: `eslint fix` and `prettier -w` for more automated and consistent formatting. Can be invoked via `pnpm lint:fix` or simply `task lint:frontend`

Note that running `prettier -w` generated a much larger changeset than I initially intended, but I do think it makes sense to run it.

johnnyjoygh:
@twaslowski I suggest reverting the changes in `./web/src/types/proto` which is unrelated to the PR via `buf generate`. 

> Note that running prettier -w generated a much larger changeset than I initially intended, but I do think it makes sense to run it.

As for `prettier -w`, we should ignore the generated files in `./web/src/types/proto`.

twaslows:
Thanks for the feedback! I reverted the previous changed and added a `.prettierignore` to ignore `types/proto` as well as the `locales` directory.

web/pnpm-lock.yaml
boojack, line 1: maybe unrelated change

Taskfile.yaml
boojack, line 1: Looks like it's your personal preference. could we move it into `./scripts/` or just remove it

twaslowski:
Reverted most changes back down to a minimum, to just implement the features and keep my personal preferences out of the changeset entirely. Sorry for making this more complicated than it needed to be.

PR #4973, opened 2025-08-07
Add a new LOG_STACKTRACES option

bcspragu:
Adds a new command line configuration parameter for logging stacktraces on panics

I found this useful for development, where I kept hitting nil panics when using the API and had trouble figuring out where they were happening with the default logging

### DIFF
--- a/bin/memos/main.go
+++ b/bin/memos/main.go
@@ -36,15 +36,16 @@ var (
 		Short: `An open source, lightweight note-taking service. Easily capture and share your great thoughts.`,
 		Run: func(_ *cobra.Command, _ []string) {
 			instanceProfile := &profile.Profile{
-				Mode:        viper.GetString("mode"),
-				Addr:        viper.GetString("addr"),
-				Port:        viper.GetInt("port"),
-				UNIXSock:    viper.GetString("unix-sock"),
-				Data:        viper.GetString("data"),
-				Driver:      viper.GetString("driver"),
-				DSN:         viper.GetString("dsn"),
-				InstanceURL: viper.GetString("instance-url"),
-				Version:     version.GetCurrentVersion(viper.GetString("mode")),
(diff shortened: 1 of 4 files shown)


bin/memos/main.go
```diff
@@ -137,12 +139,18 @@ func init() {
...
+	if err := viper.BindPFlag("log-stacktraces", rootCmd.PersistentFlags().Lookup("log-stacktraces")); err != nil {
+		panic(err)
+	}
 
 	viper.SetEnvPrefix("memos")
 	viper.AutomaticEnv()
 	if err := viper.BindEnv("instance-url", "MEMOS_INSTANCE_URL"); err != nil {
 		panic(err)
 	}
+	if err := viper.BindEnv("log-stacktraces", "MEMOS_LOG_STACKTRACES"); err != nil {
```
bcspragu: I'm less familiar with viper, but I think this is needed because it won't automatically turn hyphens into underscores. The `unix-sock` flag will also suffer from this I think. Might be worth just applying the string replacer so that it works for all future hyphenated flags.

server/router/api/v1/logger_interceptor.go
```diff
@@ -43,6 +45,9 @@ func (*LoggerInterceptor) loggerInterceptorDo(ctx context.Context, fullMethod st
 	logAttrs := []slog.Attr{slog.String("method", fullMethod)}
 	if err != nil {
 		logAttrs = append(logAttrs, slog.String("error", err.Error()))
+		if in.logStacktrace {
```
johnnyjoygh, line 48: Instead of adding a new flag, can we reuse the `mode` field in profile? We can always log stack trace in `dev` mode.

bcspragu: For sure, done

boojack (approved):

PR #4974, opened 2025-08-07
Update nl.json

Stephan-P:
Checking de Demo, I stumbled upon "Leeg je gedachten..." ("Empty your thoughts..." as translation for "Any thoughts...") which cam across rather odd. I decided to suggest a correction for that... and then made some more improvements. Please have these suggestions reviewed by the original Dutch translator who did a truly fine job. I wouldn't just want to impose. Kind regards, Stephan


johnnyjoygh (approved):
LGTM, thanks a lot

Issue #4978, opened 2025-08-07
Preferences are not being saved


maximilian-krauss:
### Describe the bug

I'm using the latest memos version from `main` branch and it looks like that the Preference section inside the settings is broken:

https://github.com/user-attachments/assets/ee7bf426-d6f2-4cf6-ab26-14362cdce631

On any change in the preferences all settings are lost and are not being saved.

### Steps to reproduce

Go to settings -> preferences and update one of the values (theme for example)

### The version of Memos you're using

latest - main

PR #4980, opened 2025-08-08
fix: preference not being saved

varsnotwars:
`initialUserStore` has temporal coupling which led to `fetchUserSettings` unintentionally being called before a user was set in the store.

Fixed by moving the call to occur after the user has been set to the store.

Added comments on the order, a todo to try avoid similar bugs, and a console.error call to help debug if it occurs before/if a refactor is done.

fixes: #4978

### DIFF
--- a/web/src/store/user.ts
+++ b/web/src/store/user.ts
@@ -177,8 +177,6 @@ const userStore = (() => {
 
   const fetchUserSettings = async () => {
     if (!state.currentUser) {
-      console.error("No current user");
-
       return;
     }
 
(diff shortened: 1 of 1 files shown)


web/src/store/user.ts
```diff
@@ -177,6 +177,8 @@ const userStore = (() => {
 
   const fetchUserSettings = async () => {
     if (!state.currentUser) {
+      console.error("No current user");
+
```
johnnyjoygh:
```suggestion
```

johnnyjoygh (approved):

PR #4981, opened 2025-08-08
fix: Add mime type check for file types

SaraVieira:
👋 

First of all thank you so much for your project!

When using I tried to upload a `.3mf` and got an error that said: type is required, turns out chrome doesn't understand the mime type of that file so in the PR I added a library that checks for the mime type if the browser cannot allowing way more file types to be uploaded

Let me know your thoughts


SaraVieira:
Hell yeah! Danke!

PR #4987, opened 2025-08-08
fix: pinned shortcut comparison operators

### LINKED ISSUE #4986: Shortcut filter pinned == true fails with "invalid identifier for ==" error
### Describe the bug

According to [document](https://www.usememos.com/docs/getting-started/shortcuts#2-pinned-memos-only) we can use comparison operators with pinned fillter.
When creating shortcuts with boolean comparison filters like pinned == true or pinned == false, the system fails with an "invalid identifier" error, preventing filter memos by pinned status using comparison operators.
```
ERROR unknown error method=/memos.api.v1.ShortcutService/UpdateShortcut error="rpc error: code = InvalidArgument desc = invalid filter: failed to convert
   filter to SQL: invalid identifier for _==_"
```

### Steps to reproduce

1. Navigate to shortcuts creation/editing
2. Try to create a shortcut with filter: pinned == true
3. Observe the "invalid identifier for _==_" error
4. Note that standalone pinned filter works, but pinned == true or pinned != true does not

main

_No response_

chojs23:
Pinned wasn't included in the allowed identifiers for comparison operations (_==_)
Currently only allowed standalone pinned (line 375), but not pinned == true comparisons.

Added "pinned" to allowed identifiers for comparison operations and handling case to route pinned comparisons to a new handler.

closes #4986

### DIFF
--- a/plugin/filter/common_converter.go
+++ b/plugin/filter/common_converter.go
@@ -122,7 +122,7 @@ func (c *CommonSQLConverter) handleComparisonOperator(ctx *ConvertContext, callE
 		return err
 	}
 
-	if !slices.Contains([]string{"creator_id", "created_ts", "updated_ts", "visibility", "content", "has_task_list", "has_link", "has_code", "has_incomplete_tasks"}, identifier) {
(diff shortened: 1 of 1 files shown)


johnnyjoygh (approved):
LGTM

Issue #4989, opened 2025-08-09
Unable to save more than one preference in the Basic section


varsnotwars:
### Describe the bug

If you try to change the Language, Appearance, Theme, or Default memo visibility in the preference section only the preference you selected is saved. If you try to change another preference the previous change is wiped and the last change is saved.

### Steps to reproduce

1. Change any preference in the basic section to a value other than the one you currently have
2. Refresh the page to confirm the change has been saved
3. Change a different preference
4. Observe that the latest change was saved but the change you made in step 1 was removed

### The version of Memos you're using

latest

PR #4990, opened 2025-08-09
fix: preferences being overwritten

varsnotwars:
The current implementation of `UpdateUserSetting` in the user service replaces the entire `value` column if the record already exists in the table. The front end sends an update request as soon as a single preference is changed resulting in the backend only receiving a single preference at a time. This means the record is always overwritten with only a single preference.

Also, `UpdateUserSetting` is only used to update the general settings even though method is setup to handle other user settings types. I have added comments explaining how these other types are updated and where we should add logic if we later want to use `UpdateUserSetting` to handle other user setting types.

Also worth considering doing a migration on the `user_setting` table to change the type of the `value` column from a text type to a json type, which _may_ make it easier to simplify some of the logic in the service layer.

fixes: #4989


johnnyjoygh (approved):
LGTM

Issue #4924, opened 2025-07-28
.ps1 attachments are treated as images


thequickbrownfox1234:
### Describe the bug

Hi,
When creating a new memo and attaching some files, all files are listed correctly and I'm able to download those, except for .ps1 files.

.ps1 files attached to a memo are not listed alongside the other files. Instead, they are treated as images, they appear with a broken image icon and cannot be previewed or downloaded.

Tested on Memos v0.24.0 and Chromium 130.0.6723.122, Firefox 128.4.0esr, Firefox 141 (Android).

### Steps to reproduce

1. Create a new memo
2. Attach a .ps1 file
3. Try to preview or download it

### The version of Memos you're using

v0.25.0

### Screenshots or additional context

<img width="681" height="209" alt="Image" src="https://github.com/user-attachments/assets/ce6bdca2-99fe-4830-ae97-a6ab3cf7a687" />

PR #4993, opened 2025-08-10
When the type is not found default to text/plain

SaraVieira:
After trying to add some `.gb` files I also encountered the `type is required` error.

This fixes it and just makes any file not in the `mime-db` a text/plain type and that makes the API happy and the browser happy because it will open it in a new tab when it is text like a `.ps1` file and download it when it's not like a `gb` file

Also closes https://github.com/usememos/memos/issues/4924 as they are now treated as text


web/pnpm-lock.yaml
boojack, line 1: maybe unrelated change

web/src/components/MemoEditor/ActionButton/UploadAttachmentButton.tsx
```diff
@@ -10,6 +10,7 @@ import { attachmentStore } from "@/store";
 import { Attachment } from "@/types/proto/api/v1/attachment_service";
 import { MemoEditorContext } from "../types";
 
+
```
boojack: extra empty line
```suggestion
```

boojack (changes requested):

web/pnpm-lock.yaml
SaraVieira: Change removed

PR #5001, opened 2025-08-13
chore: simplify convert reaction

varsnotwars:
This PR simplifies `convertReactionFromStore` to a pure function that just maps the db model to response model.

A redundant call to the db for the creator id has been removed as the reaction already has the creator id. This also partially addresses #4946 as each reaction on a memo called the database to retrieve the creator. So if a memo had 10 reactions, 10 calls were made to the database, now it is 0 calls.

`convertMemoFromStore`  has also been simplified as it no longer makes a call to retrieve reactions as a fallback as this logic has been moved up a layer to the calling method, this is inline with other calls to the database that occur at this level, e.g.,  `UpdateMemo` calls to get the current user, and the memo that is being updated, and now also retrieves the reactions for the memo. This change also moves `convertMemoFromStore` closer to being a pure mapping function and fixing the n+1 problem in `ListMemos`

Also addressed a n+1 problem in `ListMemoComments` but this endpoint is not called from the front end at all. Might be worth removing this endpoint.

### DIFF
--- a/server/router/api/v1/memo_service.go
+++ b/server/router/api/v1/memo_service.go
@@ -571,42 +571,58 @@ func (s *APIV1Service) ListMemoComments(ctx context.Context, request *v1pb.ListM
 		return nil, status.Errorf(codes.Internal, "failed to list memo relations")
 	}
 
-	reactionMap := make(map[string][]*store.Reaction)
-	memoNames := make([]string, 0, len(memoRelations))
-	for _, m := range memoRelations {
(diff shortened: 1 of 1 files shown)


varsnotwars:
@boojack ready for review now.

Latest commit addresses a n+1 problem in `ListMemoComments` but this endpoint is not called from the front end at all. Might be worth removing this endpoint.

boojack (approved):
LGTM

PR #5006, opened 2025-08-14
fix: Update SSO login flow to encode redirect_uri

Amazingca:
This fix is intended to fix some authorization flow errors that I have encountered with my identity provider. It seems to be standard to encode the `redirect_uri` so that the server can properly parse the request parameters. Without encoding, my server (running pocket-id) is unable to handle the request.

Please let me know if you have any thoughts or suggestions.


boojack (approved):
LGTM

PR #5008, opened 2025-08-15
feat: attachments by id

varsnotwars:
This PR addresses attachments in regards to the n+1 problem listed in #4946 .

`convertAttachmentFromStore` has been converted to a pure function that only maps the database model to a response model.

The store.attachment model now has a composed field - memo_uid - that is used to avoid querying the db to retrieve the related memo, which was done for every memo in the list of memos. This change is inline with the parent_uid composed field in the store.memo model.

The ListAttachments method can now also filter by memo_id.

There is now a single call to retrieve the attachments in ListMemos instead of 2 calls per memo. `convertMemoFromStore`  is closer to being a pure mapping function and completely removing the n+1 problem. The last remaining entities are the memo relations which I will address in the next PR.

### DIFF
--- a/plugin/filter/common_converter.go
+++ b/plugin/filter/common_converter.go
@@ -207,7 +207,7 @@ func (c *CommonSQLConverter) handleInOperator(ctx *ConvertContext, callExpr *exp
 		return err
 	}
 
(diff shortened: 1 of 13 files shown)


johnnyjoygh (approved):
LGTM

PR #5079, opened 2025-09-08
fix: avoid hiding reaction selector and keep it always shown for mobile

### LINKED ISSUE #5045: Reaction to Notes in mobile interface

It would be great if we could have emoji reactions to posts and comments.

-> Discovered this already exists, but it is somehow hidden on the mobile interface.

### Type of feature

User Interface (UI)

### Additional context

Sometimes showing support to a comment or highlighting it with something visual could be very useful. Especially in a multi user deployment.

giacomocerquone:
closes #5045

### TLDR of the problem

Currently there is no proper way of adding a reaction to a note on mobile because the reaction selector is only shown on hover (`hover` and `focus` are applied altogether when you tap on mobile devices but it's much more of a hack than a real thought user experience).
This PR makes it always visible.

Considering the very small size of the button, I think it's perfectly safe to keep it as visible as it doesn't obscure or shadows any important content.

## Screenshot

image

### DIFF
--- a/web/src/components/MemoView.tsx
+++ b/web/src/components/MemoView.tsx
@@ -136,7 +136,7 @@ const MemoView: React.FC<Props> = observer((props: Props) => {
   ) : (
     <div
       className={cn(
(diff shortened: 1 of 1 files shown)


johnnyjoygh (approved):
LGTM

Issue #5097, opened 2025-09-15
Pinning doesn't work


gnammix:
### Describe the bug

I'm on version: v0.25.1, pinned memos seems to not be put on top.
Sorting seems to override pinning.

### Steps to reproduce

Not sure how to reproduce, was working ok on previous version, so I suspect a regression.

### The version of Memos you're using

v0.25.1

### Screenshots or additional context

<img width="681" height="682" alt="Image" src="https://github.com/user-attachments/assets/6d51aaa1-a47a-48a3-b5d4-14ea4e70f369" />

Novapixel1010:
This is the same issue reported #5109

gnammix:
Yes looks so

PR #5099, opened 2025-09-16
fix:update sort to include pinned

varsnotwars:
Update sort functions to include a check if memo is pinned.

fixes: #5097


boojack:
Thanks, it should be fixed in https://github.com/usememos/memos/commit/b4ea7d843f872b6a5f0ab43368e6673c6bac59d2

PR #5108, opened 2025-09-19
fix: fallback mime type for drag-n-drop attachments

xtrafrancyz:
After fixes #4993 and #4981 for attachment button, drag-and-drop and paste actions were left broken. This fixes it.


johnnyjoygh (approved):
LGTM

Issue #5109, opened 2025-09-19
PInned but not show on the top


railzen:
### Describe the bug

After upgrading to V0.25.1, if you modify any other articles, the top article will be demoted to the second priority.

### Steps to reproduce

After upgrading to V0.25.1, if you modify any other articles, the top article will be demoted to the second priority.

### The version of Memos you're using

V0.25.1

wjup:
+1

amamoh:
I can confirm, now pinned are on the bottom of the list  :/

Novapixel1010:
Can confirm this is not working. I also confirmed it via the demo. I think it works in `v0.25.0`

<img width="1465" height="934" alt="Image" src="https://github.com/user-attachments/assets/c4f35d9a-ad07-4f05-b156-96e27d8e9a40" />

boojack:
Should be fixed in https://github.com/usememos/memos/commit/b4ea7d843f872b6a5f0ab43368e6673c6bac59d2

PR #5150, opened 2025-10-04
Fixed the pinned message issue

rashelshah:
#5109 fixed this issue. 
Now the messages that are pinned will show at the top. 

Before:

https://github.com/user-attachments/assets/4bb16158-f597-4d59-8b99-ac6677ea08b1

After:

https://github.com/user-attachments/assets/69a32b04-cd2c-4266-9ab0-4f42762e2b56

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1,0 +1,10019 @@
+{
+  "name": "memos",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "memos",
+      "dependencies": {
+        "@dnd-kit/core": "^6.3.1",
+        "@dnd-kit/sortable": "^10.0.0",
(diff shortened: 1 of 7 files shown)


varsnotwars:
This PR does not fix the issue because the GetMemos query can order by asc or desc and there is a limit to how many memos are returned depending on the pagination size. Only sorting on the front end after the fact means pinned posts aren't even retrieved if the memos are outside the pagination size limit once the order by filter has been applied.

Importantly, there is already a [PR](https://github.com/usememos/memos/pull/5099) from 3 weeks ago that fixes this.

varsnotwars:
Also you have committed 10k line change to the package-lock.json file which isn't even needed for your changes.

boojack:
refer to https://github.com/usememos/memos/commit/b4ea7d843f872b6a5f0ab43368e6673c6bac59d2

PR #5185, opened 2025-10-23
fix: make thumbnail generation for images in S3 configurable

### LINKED ISSUE #5183: High memory usage when opening resource tab
### Describe the bug

Before commit https://github.com/usememos/memos/commit/e4f6345342705a6b834a4aa165c6b82cafb40cf9, this issue only occurs when storage is set to local or database.
Using S3 as storage avoids high memory usage because when S3 is used, the frontend loads images via external links (S3 presigned URLs), eliminating the need for additional processing by the memos instance itself.

However, https://github.com/usememos/memos/commit/e4f6345342705a6b834a4aa165c6b82cafb40cf9 also points the image URL to the memos instance when using S3 as storage, causing high memory usage and contradicting the original intent of using S3 to reduce workload.

In my case, I deployed memos using docker with memory limit and using S3 as storage.
Update to this commit caused the program to crash frequently.

### Steps to reproduce

1. Update memos to commit https://github.com/usememos/memos/commit/e4f6345342705a6b834a4aa165c6b82cafb40cf9.
2. Open resource tab.
3. Program crashes due to memory limit.

v0.25.1

_No response_

FloDwld:
This pull request adds a toggle to disable thumbnails for images stored in S3. The default value is to not generate thumbnails.

Fixes #5183

### DIFF
--- a/proto/gen/openapi.yaml
+++ b/proto/gen/openapi.yaml
@@ -78,23 +78,35 @@ paths:
             parameters:
                 - name: pageSize
                   in: query
-                  description: "Optional. The maximum number of attachments to return.\r\n The service may return fewer than this value.\r\n If unspecified, at most 50 attachments will be returned.\r\n The maximum value is 1000; values above 1000 will be coerced to 1000."
+                  description: |-
+                    Optional. The maximum number of attachments to return.
+                     The service may return fewer than this value.
+                     If unspecified, at most 50 attachments will be returned.
+                     The maximum value is 1000; values above 1000 will be coerced to 1000.
                   schema:
                     type: integer
                     format: int32
                 - name: pageToken
                   in: query
-                  description: "Optional. A page token, received from a previous `ListAttachments` call.\r\n Provide this to retrieve the subsequent page."
+                  description: |-
+                    Optional. A page token, received from a previous `ListAttachments` call.
+                     Provide this to retrieve the subsequent page.
                   schema:
                     type: string
                 - name: filter
                   in: query
-                  description: "Optional. Filter to apply to the list results.\r\n Example: \"type=image/png\" or \"filename:*.jpg\"\r\n Supported operators: =, !=, <, <=, >, >=, :\r\n Supported fields: filename, type, size, create_time, memo"
+                  description: |-
+                    Optional. Filter to apply to the list results.
+                     Example: "type=image/png" or "filename:*.jpg"
+                     Supported operators: =, !=, <, <=, >, >=, :
+                     Supported fields: filename, type, size, create_time, memo
                   schema:
                     type: string
                 - name: orderBy
                   in: query
-                  description: "Optional. The order to sort results by.\r\n Example: \"create_time desc\" or \"filename asc\""
+                  description: |-
+                    Optional. The order to sort results by.
+                     Example: "create_time desc" or "filename asc"
                   schema:
                     type: string
             responses:
@@ -118,7 +130,9 @@ paths:
             parameters:
                 - name: attachmentId
                   in: query
-                  description: "Optional. The attachment ID to use for this attachment.\r\n If empty, a unique ID will be generated."
+                  description: |-
+                    Optional. The attachment ID to use for this attachment.
+                     If empty, a unique ID will be generated.
                   schema:
                     type: string
             requestBody:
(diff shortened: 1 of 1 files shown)


proto/api/v1/workspace_service.proto
```diff
@@ -169,6 +169,9 @@ message WorkspaceSetting {
     bool enable_blur_nsfw_content = 9;
     // nsfw_tags is the list of tags that mark content as NSFW for blurring.
     repeated string nsfw_tags = 10;
+    // use_thumbnails_for_s3_images enables thumbnail generation for images stored in S3.
+    // When false, images stored in S3 will not have thumbnails generated.
+    bool use_thumbnails_for_s3_images = 11;
```
boojack: Maybe it's better to put it into each s3 storage config.

FloDwld: The current implementation requires the frontend to have access to that value. The S3 storage config is only exposed to admin users, making that not possible.

If this does not look right to you, I've an idea for an alternative implementation. See below's top-level comment.

FloDwld:
An alternative implementation would be to introduce a 3-way or 4-way setting (no thumbnails, on-the-fly, stored in S3) stored in the S3 configuration with the following semantics:

- no thumbnails: nothing special, would just require exposing that to users
- on-the-fly: current behavior to generate and cache a thumbnail locally when it is first requested
- stored in S3 only: generate a thumbnail whenever a new attachment is uploaded and store that in S3. The resource model (and the database) would need to be extended with a separate reference_thumb column that stores a pre-signed URL to the thumbnail image.
- (optional) stored in S3 for new uploads and on-the-fly for existing resources

This would be the cleanest change in my opinion, but is a larger rework and requires a database migration. What do you think?

boojack:
> An alternative implementation would be to introduce a 3-way or 4-way setting (no thumbnails, on-the-fly, stored in S3) stored in the S3 configuration with the following semantics:
> 
> * no thumbnails: nothing special, would just require exposing that to users
> * on-the-fly: current behavior to generate and cache a thumbnail locally when it is first requested
> * stored in S3 only: generate a thumbnail whenever a new attachment is uploaded and store that in S3. The resource model (and the database) would need to be extended with a separate reference_thumb column that stores a pre-signed URL to the thumbnail image.
> * (optional) stored in S3 for new uploads and on-the-fly for existing resources
> 
> This would be the cleanest change in my opinion, but is a larger rework and requires a database migration. What do you think?

If just solving the issue of excessive memory usage, we can first just add a boolean field in `WorkspaceStorageSetting`, similar to the “enable thumbnail” option. When it’s enabled, we keep the current behavior (storing files locally).

FloDwld:
I've refactored the pull request such that the setting is now located in the storage configuration. Note that the current value is reflected in the frontend attachment object, because no access to the storage configuration is possible for non-admin users.

server/router/api/v1/attachment_service.go
```diff
@@ -368,7 +381,7 @@ func (s *APIV1Service) DeleteAttachment(ctx context.Context, request *v1pb.Delet
 	return &emptypb.Empty{}, nil
 }
 
-func convertAttachmentFromStore(attachment *store.Attachment) *v1pb.Attachment {
+func (s *APIV1Service) convertAttachmentFromStore(ctx context.Context, attachment *store.Attachment) *v1pb.Attachment {
```
varsnotwars: This is a mapping function, there should be no calls to the database done at this level. Move calls to the database a level higher to the calling method. This change introduces tight coupling between pure mapping logic and database reads.

FloDwld: Thanks for the feedback. I've introduced the storage settings as an additional paramter to the function to avoid the coupling.

FloDwld:
I've rebased the pull request to the current main to fix merge conflicts.

proto/api/v1/attachment_service.proto
```diff
@@ -84,6 +84,9 @@ message Attachment {
   // Optional. The related memo. Refer to `Memo.name`.
   // Format: memos/{memo}
   optional string memo = 8 [(google.api.field_behavior) = OPTIONAL];
+
+  // Optional. Output only. Whether to use thumbnails for this attachment when stored in S3.
+  optional bool use_thumbnail_for_s3_image = 9 [(google.api.field_behavior) = OUTPUT_ONLY];
```
boojack: We should fetch the value from the setting instead of passing it into attachment.

proto/api/v1/workspace_service.proto
```diff
@@ -147,6 +147,9 @@ message WorkspaceSetting {
     }
     // The S3 config.
     S3Config s3_config = 4;
+    // use_thumbnails_for_s3_images enables thumbnail generation for images stored in S3.
+    // When false, images stored in S3 will not have thumbnails generated.
+    bool use_thumbnails_for_s3_images = 5;
```
boojack: how about `enable_s3_image_thumbnails`

proto/api/v1/attachment_service.proto
boojack: If normal users cannot get this setting, we can also update the code to allow them to access non-sensitive settings.

FloDwld: I just gave that a try, but still see one drawback in my current draft. My current plan is to add a fetch for the storage settings to the workspace initialization code in the frontend. However, this fetch would be executed independently of whether the setting value is actually needed to render the page or not.

FloDwld: I've moved the option out of the attachment service into the storage settings. Low-priv users are now allowed to read the thumbnail setting. All other values are replaced with non-sensitive default values.

PR #5256, opened 2025-11-17
Fix max-w on tags in the sidebar

TomK32:
In Firefox on Linux the tags on the sidebar are truncated even shorter tags. It might be my larger than usual min fonts or something because Chromium on the same machine doesn't have this problem. Anways, changing the 16px to 1em works fine with short and long tags.
Screenshot 2025-11-17 at 13-04-15 Memos


TomK32:
@boojack I'm not a fan of co-pilot and it isn't even consistent in its suggestions, so I'll just ignore that.

PR #5271, opened 2025-11-21
fix(markdown): use rune-based truncation to prevent invalid UTF-8

gitkeniwo:
Fixes invalid UTF-8 error caused by https://github.com/usememos/memos/commit/739fd2cde6d7766b6663fa3fc8ea7727667b9484

```
ERROR: [core] [Server #1]grpc: server failed to encode response: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8
```

`truncateAtWord` used byte slicing (`s[:maxLength]`), which might split CJK or emojis in half and cause invalid utf-8 errors in serialization

this change 1) converts the string to `[]rune` before truncation and 2) adds more test cases covering CJK and emojis

### DIFF
--- a/plugin/markdown/markdown.go
+++ b/plugin/markdown/markdown.go
@@ -389,15 +389,18 @@ func uniqueLowercase(strs []string) []string {
 }
 
 // truncateAtWord truncates a string at the last word boundary before maxLength.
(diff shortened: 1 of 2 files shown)


johnnyjoygh:
Close by https://github.com/usememos/memos/commit/b78d4c256848447fa16a186347a61f2b85a21f4f

Issue #5319, opened 2025-12-06
Wrong checkbox checked when it's preceded by checkbox in a codeblock


spaghetti-coder:
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.25.3

### Bug Description

Wrong checkbox checked when it's preceded by checkbox in a codeblock

### Steps to Reproduce

1. Create a note:
    ````
    ```
    -[ ] foo
    ```
    - [ ] bar
    ````
    I.e. the first checkbox is in a codeblock.
2. In view mode click `bar` checkbox

### Expected

```
-[ ] foo
```
- [x] bar

### Actual

```
-[x] foo
```
- [ ] bar

### Screenshots & Additional Context

**Client env**:
- Linux Mint 22.2, Chrome 142.0.7444.175 / Brave
- Android 15, Chrome 142.0.7444.171

**Deployment**:
- Container from official docker image
- Official demo site

maishivamhoo123:
@team can you please assign this issue to me ?

PR #5321, opened 2025-12-06
Fix/checkbox indexing codeblock

maishivamhoo123:
### Fix: Incorrect checkbox toggling when preceded by a codeblock checkbox

This PR fixes **Issue #5319**, where toggling a checkbox in view mode updates the **wrong task** if a checkbox inside a fenced code block appears earlier in the note.

#### What’s fixed
- Checkboxes inside code blocks (``` or ~~~) are no longer counted as task items.
- Task indexing now correctly ignores codeblock content.
- The correct checkbox is always toggled.

A demo video is attached showing the bug and the expected behavior.

https://github.com/user-attachments/assets/7039f249-8b26-4081-b635-3ed05aafbd7a

### DIFF
--- a/web/src/utils/markdown-manipulation.ts
+++ b/web/src/utils/markdown-manipulation.ts
@@ -24,6 +24,39 @@ export function toggleTaskAtLine(markdown: string, lineNumber: number, checked: 
 
   return lines.join("\n");
 }
+export function isInsideCodeBlock(lines :string [] , index : number):boolean{
+  let inside = false;
+  for(let i = 0 ; i <= index ; i++){
+    if(lines[i].startsWith("```")){
+      inside = !inside;
+    }
+  }
+   return inside;
+}
+export function toggleTaskAtIndex(markdown: string, taskIndex: number, checked: boolean): string {
+  const lines = markdown.split("\n");
+  const taskPattern = /^(\s*[-*+]\s+)\[([ xX])\](\s+.*)$/;
+
+  let currentTaskIndex = 0;
+
+  for (let i = 0; i < lines.length; i++) {
+    if (isInsideCodeBlock(lines, i)) continue;
+    const line = lines[i];
+    const match = line.match(taskPattern);
+
+    if (match) {
+      if (currentTaskIndex === taskIndex) {
+        const [, prefix, , suffix] = match;
+        const newCheckmark = checked ? "x" : " ";
+        lines[i] = `${prefix}[${newCheckmark}]${suffix}`;
+        break;
+      }
+      currentTaskIndex++;
+    }
+  }
+
+  return lines.join("\n");
+}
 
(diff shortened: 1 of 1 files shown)


web/src/utils/markdown-manipulation.ts
```diff
@@ -24,14 +24,23 @@ export function toggleTaskAtLine(markdown: string, lineNumber: number, checked:
 
   return lines.join("\n");
 }
-
+export function isInsideCodeBlock(lines :string [] , index : number):boolean{
+  let inside = false;
+  for(let i = 0 ; i <= index ; i++){
+    if(lines[i].startsWith("```")){
```
spaghetti-coder, line 30: It doesn't trap the following cases:
1. When codeblock is embraced with `~~~` instead of ` ``` `
2. When the codeblock has let's say 2 spaces offset:
~~~
The codeblock has 2 spaces offset
  ```
  - [ ] foo
  ```
- [ ] bar
~~~
3. The following one will still make `foo` checked:
~~~
````
``` <- This will be rendered as a triple back quote inside the code block
- [ ] foo
````
- [ ] bar
~~~

File: web/src/utils/markdown-manipulation.ts [hunk 2]
```diff
@@ -24,14 +24,23 @@ export function toggleTaskAtLine(markdown: string, lineNumber: number, checked:
...
+   return inside;
+}
 export function toggleTaskAtIndex(markdown: string, taskIndex: number, checked: boolean): string {
   const lines = markdown.split("\n");
   const taskPattern = /^(\s*[-*+]\s+)\[([ xX])\](\s+.*)$/;
 
   let currentTaskIndex = 0;
 
   for (let i = 0; i < lines.length; i++) {
+    if (isInsideCodeBlock(lines, i)) continue;
```
spaghetti-coder, line 43: It will unnecessarily slow down the execution. Better put the check inside the match block

spaghetti-coder (changes requested):
Thanks for the quick reaction, but found some issues with the solution.

web/src/utils/markdown-manipulation.ts
```diff
@@ -24,14 +24,23 @@ export function toggleTaskAtLine(markdown: string, lineNumber: number, checked:
 
   return lines.join("\n");
 }
-
+export function isInsideCodeBlock(lines :string [] , index : number):boolean{
```
spaghetti-coder, line 27: The function only covers very straight scenarios. A sample when it fails

~~~~
<!--
Some comments
```
More comments
-->
```
The function thinks this line is not in a codeblock
```
~~~~

maishivamhoo123: sorry sir for inconvenience i will fix it please give me some time.

PR #5367, opened 2025-12-20
feat(web): render index.html as template to inject HTML tags

### LINKED ISSUE #4276: Autodiscoverable RSS feeds: <link>s in statically served HTML

It'd be excellent if the RSS URLs were added as `<link>` tags in the `<head>` of the HTML pages, so they can be auto-discovered! Eg:

```html
<link rel="alternate" type="application/rss+xml" href="https://memos.example.com/explore.xml" title="Explore">
```

Note that these tags need to be present in the _base HTML_, not added with JS/React (otherwise they won't be auto-discoverable). This is also needed for the various Social Share link/meta tags that would also be worth adding (so resolving this wouldn't be just for this feature).

### Type of feature

User Interface (UI)

### Additional context

_No response_

UiP9AV6Y:
render the SPA differently based on the route. if a page is available as RSS feed, inject a link element into the head element of the HTML variant. this ensures feeds are discoverable.

this change replaces the middleware with custom router handlers to serve either assets from the embedded filesystem or the SPA document.

closes #4276

### DIFF
--- a/server/router/frontend/frontend.go
+++ b/server/router/frontend/frontend.go
@@ -3,14 +3,12 @@ package frontend
 import (
 	"context"
 	"embed"
+	"errors"
 	"io/fs"
-	"net/http"
 
 	"github.com/labstack/echo/v4"
-	"github.com/labstack/echo/v4/middleware"
 
 	"github.com/usememos/memos/internal/profile"
-	"github.com/usememos/memos/internal/util"
 	"github.com/usememos/memos/store"
 )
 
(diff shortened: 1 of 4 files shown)


UiP9AV6Y:
the `<!-- {{ printf "memos.metadata.head %s>" "--" }}` in the index.html ensures that the file remains valid HTML in case of debugging the site with Vite alone without a backend service. when rendered as template the resulting HTML will end up with two commends with injected code in between

```html
<!-- memos.metadata.head -->
<!-- -->
```

i also use *text/template* instead of *html/template* as the latter removes comments from templates and i could not find any information if the `memos.metadata.` comments serve any logic purpose which would break where the comments removed.

Issue #5191, opened 2025-10-27
Custom S3 Host/CDN Domain


RekinChiang:
Now exposes the raw endpoint.

ikelvingo:
I agree. I searched the closed issues, and I found that so many users wanted the URL_PREFIX back.
I don't know why memos doesn't support URL prefix, and it once supported the function yet.

<img width="1339" height="970" alt="Image" src="https://github.com/user-attachments/assets/8f2470cd-ea14-44d0-9a18-910bb48a3c0a" />

Issue #4852, opened 2025-07-13
Allow users to choose between presigned URLs and custom domain access for S3 storage


zouzonghao:
### Hi team,

Thank you for the great work on Memos! I’ve been using it with S3-compatible storage (specifically Cloudflare R2), and I would like to propose a feature enhancement regarding image resource URLs.

### Current Behavior:
When using S3 storage, the system **forces the use of presigned URLs** to serve images and other assets.

### Problems with the current approach:
1. **No CDN caching for presigned URLs:**  
   Cloudflare R2 does **not** serve presigned URLs through the CDN cache layer. I have verified this with a custom script and confirmed that presigned URLs bypass the cache, leading to higher egress costs and slower access.
   
2. **No support for custom domain access:**  
   Currently, it’s **not possible to configure Memos to use a custom domain** for R2 or other S3-compatible resources. This makes it impossible to leverage CDN benefits like caching, security, performance, or image optimization.
   
3. **Inflexible expiration time:**  
   The default expiration time for presigned URLs is **5 days**, which is quite long and could pose security risks for sensitive content. However, shortening it would break links more often.
   
4. **Limited sharing and external usage:**  
   If custom domain access were supported, users could **share links more easily** and even use Memos as a lightweight image hosting solution (e.g., for blogs or social media).

---

### Feature Request:
Please consider **adding a configuration option** that allows users to choose between:

- **Presigned URLs** (default, secure for private S3 access), or  
- **Custom domain access** (static base URL pointing to S3 resources via a public endpoint or CDN)

This flexibility would allow users to optimize for performance, cost, and ease of sharing, depending on their use case.

### Type of feature

User Experience (UX)

zouzonghao:
I designed a script to test whether Presigned URLs can be cached by Cloudflare CDN. Unfortunately, while Presigned URLs do reach Cloudflare's edge nodes, they are not being cached by the CDN.
[Presigned URLs Cloudflare CDN Cache Test Log](https://gist.github.com/zouzonghao/4c36f056d3a1a89526bc0fe2be19330b)

Issue #3819, opened 2024-08-19
Enable CDN domain name access for S3


RunDouble:
### Describe the solution you'd like

Enable CDN domain name access for S3.

### Type of feature

API

PR #5411, opened 2026-01-03
Add the function to customize the S3 domain

Euruson:
This PR adds support for configuring a custom domain (CDN) for S3 storage. When configured, attachment URLs will be generated using the custom domain instead of the default S3 presigned URLs. This allows for serving assets via a CDN or a custom domain name. Existing attachments will **automatically migrate** to the new URL format via the background runner **immediately** upon saving settings.

A lot of people asked for this feature to accelerate accessing and reduce network traffic costs:
#3575
#3717 
#3819
#4852
#5191


Euruson:
Also, I use `filePathTemplate` for attachment naming to avoid unpredictable filenames returned by S3.

Some S3-compatible providers return the bucket name as part of the filename for large files but not for small ones. This inconsistency stems from the multipart upload mechanism used for larger files, which can return a different key format than the single PUT mechanism used for smaller files.

PR #5434, opened 2026-01-05
Hardening: Switch to non-root user in Docker

RinZ27:
Dropped root privileges in the production Docker image. Running the app process as a dedicated `memos` user instead of root to minimize the potential impact if the process is compromised. Standard hardening practice.

Also updated the directory creation and ownership to ensure the `memos` user has proper access to its data volumes.


RinZ27:
Good catch. Updated the Dockerfile to use explicit UID/GID 10001 for better predictability with volume mounts.

PR #5476, opened 2026-01-11
fix: inbox notifications fail to be fetched when using postgres driver

C1C0:
### Error

When visiting `/inbox`, server was responding with:

`ERROR server error method=/memos.api.v1.UserService/ListUserNotifications error="internal: rpc error: code = Internal desc = failed to list inboxes: pq: operator does not exist: text ->> unknown"`

### Fix

Updated the message type filter in `ListInboxes` to cast the `message` column from TEXT to JSONB before extracting the `type` field, ensuring compatibility with the current database schema where `inbox.message` is stored as TEXT.


johnnyjoygh:
Thanks for the reporting, it should already be fixed in https://github.com/usememos/memos/commit/31f634b71ab26eda260c58a6cd61d02e515268e7

PR #5491, opened 2026-01-16
fix: detect autolinks for HasLink property

thefatcode:
Fixes #5407

Auto-links (raw URLs like `https://example.com`) were not being detected when setting the `HasLink` property in memo payloads.

### Changes
- Added `gast.KindAutoLink` to the switch case in `ExtractProperties()`
- Added `gast.KindAutoLink` to the switch case in `ExtractAll()`

This ensures that both standard markdown links `[text](url)` and auto-links are properly detected.

### DIFF
--- a/plugin/markdown/markdown.go
+++ b/plugin/markdown/markdown.go
@@ -153,7 +153,7 @@ func (s *service) ExtractProperties(content []byte) (*storepb.MemoPayload_Proper
 		}
 
 		switch n.Kind() {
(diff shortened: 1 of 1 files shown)


server/router/fileserver/fileserver.go
johnnyjoygh, line 1: Looks like an unexpected change

johnnyjoygh (approved):
LGTM, otherwise

PR #5602, opened 2026-02-08
fix: handle underscores in environment variable names

### LINKED ISSUE #5597: underscores in env vars are not correctly handled
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.26.0

### Bug Description

I noticed that the environment variable `MEMOS_INSTANCE_URL` is not picked up. However, the (invalid) `MEMOS_INSTANCE-URL`, with a dash, works.

This is probably related to how you are using the [viper](https://github.com/spf13/viper?tab=readme-ov-file#working-with-environment-variables) library:

https://github.com/usememos/memos/blob/81ef53b398dd12e934eb4f0e59b49f3d7ef45998/cmd/memos/main.go#L129-L135

the flag is called `instance-url` (with a dash), which then expects `MEMOS_INSTANCE-URL`. You probably need to add a `viper.SetEnvKeyReplacer(...)` to correctly map underscores in env vars to dashes in the flag.

This is only relevant for the two variables `MEMOS_INSTANCE_URL` and `MEMOS_UNIX_SOCK`. The first underscore after `MEMOS` is handled by `SetEnvPrefix` already.

### Steps to Reproduce

- set the environment variable `MEMOS_INSTANCE_URL` in docker compose
- verify if it is picked up with `curl "https://demo.usememos.com/api/v1/instance/profile"`
- the returned `"instanceUrl"` field is an empty string.

when setting `MEMOS_INSTANCE-URL`, the `"instanceUrl"` field is populated.

### Expected Behavior

_No response_

### Screenshots & Additional Context

_No response_

x34-dzt:
Add SetEnvKeyReplacer to map dashes in flag names to underscores in env vars, so MEMOS_INSTANCE_URL and MEMOS_UNIX_SOCK work correctly.

Fixes #5597

### DIFF
--- a/cmd/memos/main.go
+++ b/cmd/memos/main.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	"github.com/spf13/cobra"
@@ -131,6 +132,7 @@ func init() {
 	}
 
 	viper.SetEnvPrefix("memos")
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	viper.AutomaticEnv()
 }
 


x34-dzt:
Closing — this was resolved directly in a69056a1

PR #5666, opened 2026-02-26
Fix migration failure when `BASIC.schemaVersion` is missing but `migration_history` exists

chazeon:
This PR fixes a migration regression for legacy installations (commonly from older Docker upgrades) where:
- `system_setting` `BASIC.schemaVersion` is empty, but
- `migration_history` still contains the applied schema versions.

Previously, Memos treated the DB as version `0.0.0` and attempted to reapply early migrations, which failed with errors like `table activity already exists`.

### DIFF
--- a/store/migrator.go
+++ b/store/migrator.go
@@ -249,6 +249,10 @@ func (s *Store) preMigrate(ctx context.Context) error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


johnnyjoygh:
should be fixed with in https://github.com/usememos/memos/commit/9628d3de21289e887ffeb6cc4b6710913a0de44e

Issue #5607, opened 2026-02-09
Auto-refresh error still not fixed


thecoffeebin:
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.26.1

### Bug Description

Similar to these issues created for v0.26.0 below, the auto-refresh bug is not fixed and all posts have disappear after letting memos idle for more than 10min. A refresh is enough to bring all posts back.

https://github.com/usememos/memos/issues/5589
https://github.com/usememos/memos/issues/5565

### Steps to Reproduce

1. Login to memos and all private posts are shown (not sure about public as it's a private instance)
2. Switch to other tab and let memos idle for more than 10min
3. Now switch back to memos tab and witness all posts disappear. Capture docker log

### Expected Behavior

All memos posts should not disappear after idling.

### Screenshots & Additional Context

```
memos  | 2026/02/09 00:47:38 INFO background runners started goroutines=10
memos  | Memos 0.26.1 started successfully!
memos  | Data directory: /var/opt/memos
memos  | Database driver: sqlite
memos  | Server running on port 5230
memos  | Access your memos at: http://localhost:5230
memos  |
memos  | Documentation: https://usememos.com
memos  | Source code: https://github.com/usememos/memos
memos  |
memos  | Happy note-taking!
memos  | 2026/02/09 00:48:09 INFO OK method=/memos.api.v1.UserService/GetUserStats
memos  | 2026/02/09 00:48:09 INFO OK method=/memos.api.v1.UserService/ListUserNotifications
memos  | 2026/02/09 00:48:09 INFO OK method=/memos.api.v1.UserService/GetUserStats
memos  | 2026/02/09 00:48:09 INFO OK method=/memos.api.v1.MemoService/ListMemos
memos  | 2026/02/09 00:48:09 INFO OK method=/memos.api.v1.MemoService/ListMemos
memos  | 2026/02/09 00:48:11 INFO OK method=/memos.api.v1.AuthService/SignOut
memos  | 2026/02/09 00:48:11 INFO client error method=/memos.api.v1.AuthService/GetCurrentUser error="unauthenticated: authentication required"
memos  | 2026/02/09 00:48:11 INFO OK method=/memos.api.v1.MemoService/ListMemos
[...]

Issue #5589, opened 2026-02-04
Auto-refresh error when switching from background to foreground


WangLei1993:
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.26.0

### Bug Description

See this video:

https://github.com/user-attachments/assets/de578958-a623-4d7e-91bc-9b7e5df967d7

When the page switches from background to foreground, it automatically refreshes the content. However, the content disappears after the refresh. Checking the Network tab, there are two 401 (Unauthorized) errors. If I manually press F5 to refresh, everything returns to normal.

### Steps to Reproduce

1. Open the page and switch it to the background.
2. Operate other applications in the foreground for a while.
3. Switch the page back to the foreground, which triggers the bug.
4. Press F5 to refresh the page to restore it, or switch to the background again and repeat steps 2 and 3 to see the content return to normal.

### Expected Behavior

Ensure content refreshes as expected

jpoggi:
Hi,

Same here with the version `neosmemo/memos:0.26.0`.

Regards.

thecoffeebin:
Not sure if it's the right way but I've pulled the source (git clone) and start it in a docker.

While the auto-refresh bug is not as frequent as before (letting it idle and switching from background to foreground), it still happens once in a while, so it is not a full fix. I know it would be better to wait for next release but this bug is quite annoying and I couldn't wait any longer.

Just drop a note here in case it's slipped again for next release.

semiligneous:
I continue to get the same issue in v0.26.1, despite the fix being released, there are a couple of related issues (#5607 and #5565). Not sure whether this will be reopened, as the fix didn't solve the issue, or it one of the other tasks will become the active one -apologies for duplicate comments therefore!
o

Issue #5565, opened 2026-02-02
v0.26.0 the memos will disappear after a while


cheetah747:
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.26.0

### Bug Description

when open the memos homapage, login ,and leave for a while,when return to hompage,and all the private memos will disappear until a refresh by F5

### Steps to Reproduce

1. login memos
2. wait for a while to do something else
3. return to memos page 
4. all of the private memos will disappear ,but public memos are visible
5. refresh by F5
6. all memos appear again

### Expected Behavior

I expected when I wait for a long time and then return to the memos page, and all the private memos will be visible.(the versions before v0.26.0 didn't have this problem)

### Screenshots & Additional Context

- Browser: Chrome 141.0.7390.108
- OS：windows 11 25H2
- Deployment: Docker
- Database: SQLite

thecoffeebin:
Came here just for this. Thanks for the fix in next release.

gsabater:
Warn to anyone who might be reading this: If you write a text and not save it, you **can** lose the text after a while (even if is just the computer entering suspension)

nwgat:
am running into this exact issue Version: v0.26.2

Issue #5643, opened 2026-02-20
No cache when edditing a memo when using PWA


iioel:
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.26.1

### Bug Description

[on mobile, PWA with chrome]
It has been multiple times that I lost the content that I add to my memos by editing it and then multitasking to another app to copy something and then coming back to memos and I find a blank page (no memos). I need to refresh to see my memos and then while editing back the same one, my unsaved changes have been lost.

This is really annoying. I don't want to same my memo each time I multitask between different apps.

I don't know if my explanations are clear. I can provide a video clip to show you what is going wrong. 

### Steps to Reproduce

1. Install memos as a PWA on android with chrome
2. Login and create a new memo. Save it.
3. Edit this newly created memo by clicking on the `3 dots` then `Edit`
4. Add new things to the memo
5. Multitask to another application without saving the modifications then come back to memos
6. You should see an empty screen (no memos) (it does not always happen, maybe after few minutes editing the same memo)
7. Refresh the page to have the memos back
8. Edit the previously edited memos
9. Your unsaved changes should not be present...

### Expected Behavior

I expected either a non-black screen were I can continue to edit my memo OR the unsaved content of my edit to by cached to be able to persist after a refresh.

### Screenshots & Additional Context

- OS: Android 16
- Phone: Pixel 8
- Browser: Chrome (memos installed as PWA)
- Authentication: SSO - PocketID
- Deployment: docker-compose
- Network: is behind a nginx reverse-proxy

docker-compose.yml
```
services:
  memos:
    image: docker.io/neosmemo/memos:0.26.1
    container_name: memos
    restart: always
    depends_on:
      - db
    ports:
      - 5230:5230
    environment:
      - MEMOS_DRIVER=postgres
      - MEMOS_DSN=user=xxx password=xxx dbname=memosdb host=db sslmode=disable

  db:
    image: docker.io/postgres:18.2
    container_name: memos_db
    restart: unless-stopped
    networks:
      - memos
    volumes:
      - memos-db-data:/var/lib/postgresql/
    environment:
      POSTGRES_USER: xxx
      POSTGRES_PASSWORD: xxx
      POSTGRES_DB: memosdb

volumes:
  memos-db-data:
```

[...]

PR #5681, opened 2026-03-04
fix: access token refresh on web app

### LINKED ISSUE #5658: Notes disappearing when returning to homepage
### Pre-submission Checklist

- [x] I have searched the existing issues and this bug has not been reported yet
- [x] I have tested this issue on the [demo site](https://demo.usememos.com) or the latest version

### Where did you encounter this bug?

Latest stable version (self-hosted)

### Memos Version

v0.26.2

### Bug Description

When leaving the page (browser tab) and returning after a while, the notes disappear. Checking the console shows that a request is sent to `/memos.api.v1.UserService/ListUserNotifications` which is failing with status 401.
Notes return when refreshing the page.

### Steps to Reproduce

1. Open memos and login
2. Go to the homepage where your notes are listed
3. Leave the page by switching tabs and wait for a while, I would say more than 15 minutes at least
4. Return to the page

### Expected Behavior

I expect the notes not to disappear and the user to stay authorized.

### Screenshots & Additional Context

- Google Chrome and Brave (happens on both)
- Docker Image v0.26.2
- Ubuntu
- Proxied by Cloudlare Tunnel (cloudflared)

peteretelej:
A bug exists within the token refresh flow that invalidates active browser sessions. Retrieving the active access token returns `null` if it's expired. Seem like a small oversight when token storage was migrated to the jwt tokens.

This means expired tokens don't refresh, and if leave the browser tab idle for 15min, the tab drops auth context and shows logged out view (you're still logged in though - reload fixes it). A few issues are open for this.

Resolves #5658, #5607, #5589, #5565
(last 3 are closed but underlying issue persisted)

May also fix some of the other issues where drafts keep getting lost (but I'm not sure, may be that the reauth is what makes the draft get lost)

### DIFF
--- a/web/src/hooks/useTokenRefreshOnFocus.ts
+++ b/web/src/hooks/useTokenRefreshOnFocus.ts
@@ -1,5 +1,5 @@
 import { useEffect } from "react";
-import { FOCUS_TOKEN_EXPIRY_BUFFER_MS, getAccessToken, isTokenExpired } from "@/auth-state";
+import { FOCUS_TOKEN_EXPIRY_BUFFER_MS, hasStoredToken, isTokenExpired } from "@/auth-state";
 
 /**
  * Hook that proactively refreshes the access token when the tab becomes visible
@@ -20,9 +20,8 @@ export function useTokenRefreshOnFocus(refreshFn: () => Promise<void>, enabled: 
         return;
       }
 
-      // Only refresh if we have a token
-      const token = getAccessToken();
-      if (!token) {
+      // Only refresh if the user has logged in before (token in localStorage)
+      if (!hasStoredToken()) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


web/src/hooks/useTokenRefreshOnFocus.ts
```diff
@@ -20,9 +20,8 @@ export function useTokenRefreshOnFocus(refreshFn: () => Promise<void>, enabled:
         return;
       }
 
-      // Only refresh if we have a token
-      const token = getAccessToken();
```
peteretelej, line 24: this was the bug: `getAccessToken()` returns null if the token we have is expired, meaning we don't get a chance to refresh it

boojack (approved):
LGTM, good catch

theacodes:
Started running into this today and noticed it was fixed but not yet released, would it be possible to get a new release out soon? Thank you

boojack:
@theacodes This has already been included in our latest RC release: https://github.com/usememos/memos/releases/tag/v0.27.0-rc.2

theacodes:
Ah I see, my docker image is pulling 0.26.2 since that's the latest
non-prerelease. I'll try out the pre releases, thank you!

iioel:
Good catch ! This actually fixed the issue I created: https://github.com/usememos/memos/issues/5643

PR #5738, opened 2026-03-19
Redirect unauthenticated protected memo access to sign in

### LINKED ISSUE #5644: Redirect to login screen when accessing protected or private memo
### Pre-submission Checklist

- [x] I have searched the existing issues and this feature has not been requested yet

### Type of Feature

User Experience (UX)

### Problem or Use Case

When I access a link to a memo that is private or protected, and I am not logged in to Memos, I get a 403 permission denied error. This leads the user to think that something is wrong with the application instead of giving them next steps (i.e. to log in).

### Proposed Solution

Ideally, when accessing a non-public memo while browsing anonymously, the user should be redirected to the login screen to authenticate. A banner notification could appear indicating "This memo is not public. Try logging in first." or similar. 

This would help users who are accessing Memos for the first time or were provided a non-public link have a more seamless experience.

### Alternatives Considered

_No response_

### Additional Context

_No response_

darkestni:
Resolves #5644

## Summary

When an anonymous user opens a protected or private memo link, Memos currently sends them to a 403 page. This is confusing because the next useful action is usually to sign in.

This change redirects unauthenticated users to the sign-in page instead, shows a short explanation, preserves the original memo URL, and returns them to that memo after sign-in or sign-up.

## Changes

- Memo detail page now distinguishes between unauthenticated, permission denied, and not found errors.
- Added auth redirect helpers for safe redirect targets and auth reasons.
- Sign-in page now reads and preserves redirect targets.
- Password sign-in now returns users to the original page after authentication.
- Sign-up flow also preserves the original target and returns there after registration.

## Behavior

- Unauthenticated access to protected/private memo:
  - Redirect to sign-in
  - Show a short message explaining the memo is not public
  - Return to the memo after successful sign-in/sign-up
- Authenticated but unauthorized access:
  - Keep existing 403 behavior
- Missing memo:
  - Go to 404

## Validation

- `corepack pnpm lint`
- `corepack pnpm build`

### DIFF
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -10,6 +10,7 @@
     "create-your-account": "Create your account",
     "host-tip": "You are registering as the Site Host.",
     "new-password": "New password",
+    "protected-memo-notice": "This memo is not public. Sign in to continue.",
     "repeat-new-password": "Repeat the new password",
     "sign-in-tip": "Already have an account?",
     "sign-up-tip": "Don't have an account yet?"
@@ -160,6 +161,7 @@
     "direction-asc": "Ascending",
     "direction-desc": "Descending",
     "display-time": "Display Time",
+    "failed-to-load": "Failed to load memo.",
     "filters": {
       "has-code": "hasCode",
       "has-link": "hasLink",
--- a/web/src/locales/zh-Hans.json
+++ b/web/src/locales/zh-Hans.json
@@ -10,6 +10,7 @@
     "create-your-account": "创建您的账户",
     "host-tip": "您正在注册为站点管理员。",
     "new-password": "新密码",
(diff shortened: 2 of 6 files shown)


boojack:
@codex review

darkestni:
Thanks for the review — I addressed the requested changes in `4c67e482`:

- moved the protected memo notice into i18n resources and used `t()` in both auth pages
- stopped routing unexpected `MemoDetail` errors to `/404` and kept them as toast-only errors
- replaced the hardcoded sign-up path in `SignIn.tsx` with `${ROUTES.AUTH}/signup`

I also localized the fallback memo load error message while touching that code path.

Validation run:
- `corepack pnpm lint`
- `corepack pnpm build`

PR #5741, opened 2026-03-19
refactor(inbox): store memo comment payloads without activity records

memoclaw:
## Summary
- remove the separate activity store model and persist memo comment notification data directly in inbox messages
- migrate existing inbox rows by backfilling memo comment payloads from activity data before dropping the activity table
- keep the user notification API behavior unchanged while simplifying notification writes and reads

## Testing
- `golangci-lint run`
- `go test ./...`


boojack:
@codex review

PR #5748, opened 2026-03-20
fix(auth): recover session via refresh cookie when localStorage is empty

### LINKED ISSUE #5728: PWA always opens on "Explore"
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this bug has not been reported
- [x] I can reproduce this bug on the latest version or the [demo site](https://demo.usememos.com)
- [x] This is a bug, not a question (use [Discussions](https://github.com/usememos/memos/discussions) for questions)

### Memos Version

v0.26.2

### Deployment Method

Docker

### Database

SQLite

### Browser & OS

Android 16 + Chrome

### Bug Description

After installing the PWA on Android, a cold start (= the app is not opened yet) always lands on the "Explore" tab. This happens only with the installed PWA, not when going to the web endpoint from the browser.

### Steps to Reproduce

Install PWA, open it.

Or for an installed PWA: close it (in the tasks manager), then open it

### Expected Behavior

The main tab being accessed, like for any other generic access to the URL

### Screenshots, Logs & Additional Context

I checked the details of the installation, the referenced URL is the main one (not a tab)

fiatcode-gh:
Fixes #5728

## Summary

PWA always opens on the Explore page instead of the home page.

On platforms with isolated storage (e.g., iOS Safari PWA), localStorage may be empty even when a valid HTTP-only refresh token cookie exists. The previous code used `hasStoredToken()` which only checked localStorage, skipping session recovery entirely for PWAs.

## Changes

- Replace `hasStoredToken()` early return in `AuthContext.initialize()` with `getAccessToken()` check
- Attempt to refresh the access token via HTTP-only cookie when no token is available
- Add optimization to skip `getCurrentUser()` call if refresh fails (avoid unnecessary network request)

## Flow Comparison

| Scenario | Before | After |
|----------|--------|-------|
| Browser, valid token | getToken from memory/localStorage | Same |
| Browser, expired token | getToken → null → fail early | getToken → null → refresh → succeed |
| PWA isolated storage | fail early, no recovery | refresh via cookie → succeed |
| New user (unauthenticated) | immediate return, 0 calls | refresh fails → return, 1 call |

## Testing

- Verified lint and typecheck pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication initialization to attempt session recovery by refreshing tokens when cached credentials are unavailable, providing better session persistence and reducing unexpected sign-outs.
  * Maintained existing public behavior and interfaces while enhancing resilience of user session restoration.

### DIFF
--- a/web/src/auth-state.ts
+++ b/web/src/auth-state.ts
@@ -61,11 +61,10 @@ export const getAccessToken = (): string | null => {
           accessToken = storedToken;
           tokenExpiresAt = expiresAt;
         }
-        // Do NOT remove expired tokens here. Callers such as InstanceContext.initialize()
-        // run concurrently with AuthContext.initialize() via Promise.all. If we eagerly
-        // delete the expired token from localStorage, hasStoredToken() (called synchronously
-        // inside AuthContext.initialize()) finds nothing and skips the refresh attempt,
-        // logging the user out even when the refresh-token cookie is still valid.
+        // Do NOT remove expired tokens here. getRequestToken() in connect.ts calls
+        // hasStoredToken() to decide whether to attempt a refresh — if we eagerly delete
+        // the expired token, it returns null immediately, skipping the refresh and sending
+        // the request without credentials.
         // clearAccessToken() handles proper cleanup after a confirmed auth failure or logout.
       }
     } catch (e) {


bluedbird:
The fix is correct and well-reasoned. One stale comment should be updated before merging.

In `web/src/auth-state.ts` lines 64–69, the comment inside `getAccessToken()` explicitly references `hasStoredToken()` being called in `AuthContext.initialize()`, which this PR removes:

```ts
// Do NOT remove expired tokens here. Callers such as InstanceContext.initialize()
// run concurrently with AuthContext.initialize() via Promise.all. If we eagerly
// delete the expired token from localStorage, hasStoredToken() (called synchronously
// inside AuthContext.initialize()) finds nothing and skips the refresh attempt,
// logging the user out even when the refresh-token cookie is still valid.
```

After this PR, `AuthContext.initialize()` no longer calls `hasStoredToken()`, so the rationale is misleading. The real reason to keep expired tokens in localStorage is still valid though — `getRequestToken()` in `connect.ts` still calls `hasStoredToken()` and relies on expired tokens being present to know it should attempt a refresh rather than returning `null` immediately.

Suggested update:
```ts
// Do NOT remove expired tokens here. getRequestToken() in connect.ts calls
// hasStoredToken() to decide whether to attempt a refresh — if we eagerly delete
// the expired token, it returns null immediately, skipping the refresh and sending
// the request without credentials.
// clearAccessToken() handles proper cleanup after a confirmed auth failure or logout.
```

fiatcode-gh:
Good catch! I've updated the comment in `web/src/auth-state.ts` as suggested. 

The comment now accurately reflects that `getRequestToken()` in `connect.ts` is the caller of `hasStoredToken()`, not `AuthContext.initialize()`.

Thanks for the review!

PR #5856, opened 2026-04-19
fix: user resource names can be uuidv4 from idp sub claim

gbw:
usernames that come from identity providers might be UUIDv4 which would fail the UIDMatcher regex causing an invalid username error when trying to log in.

## Summary by CodeRabbit

* **Enhancements**
  * Increased allowed UID length from 32 to 36 characters — validations and acceptance now permit longer identifiers.

* **Tests**
  * Added test cases to cover the updated UID length and expected validation outcomes.

### DIFF
--- a/internal/base/resource_name.go
+++ b/internal/base/resource_name.go
@@ -3,5 +3,5 @@ package base
 import "regexp"
 
 var (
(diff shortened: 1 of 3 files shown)


boojack (approved):
LGTM

PR #5864, opened 2026-04-20
add passkey support

LxnChan:
Add passkey support, users can login with passkey now.

image

image

## Release Notes

* **New Features**
  * Passkey authentication: securely register and use passkeys to sign in without passwords
  * Passkey sign-in option now available on the sign-in page
  * Passkey management in account settings: create new passkeys, view usage history with timestamps, and delete passkeys
  * Instance administrators can disable password authentication to require passkeys


LxnChan:
fixing

PR #5879, opened 2026-04-22
docs: refine SSO identity linkage design

boojack:
## Summary
- expand the SSO identity-linkage research with verified upstream schema references
- clarify the issue definition around provider-scoped identity linkage patterns
- keep the proposed memos table name as `user_identity` while retaining `provider` and `extern_uid`

## Testing
- not run (docs only)


boojack:
@codex review

PR #5880, opened 2026-04-22
fix: reduce list memo query overhead

boojack:
## Summary

This PR reduces `ListMemos` overhead without changing schema or adding indexes.

It makes three targeted changes:
- split memo relation batch lookups into directional queries and avoid eager relation snippet hydration on list responses
- reuse creator and display-time data across list conversion work in the list handlers
- cap memo comment preview fetches to 3 items and lazily resolve missing relation snippets on the client

## Why

`ListMemos` became noticeably slower on larger SQLite datasets after `0.27.x` because the list path started doing more relation and comment work per page. The goal here is to make the hot path cheaper while preserving existing behavior and avoiding schema/index changes.

## Impact

- cheaper authenticated `ListMemos` and `ListMemoComments` request paths
- smaller comment preview fetches from the web client
- reproducible API benchmarks for future regression tracking on seeded SQLite data

## Validation

- `go test ./server/router/api/v1 ./store ./store/db/sqlite ./store/db/mysql ./store/db/postgres`
- `pnpm -C web lint`
- `go test -run '^$' -bench BenchmarkListMemos -benchtime=1x ./server/router/api/v1/test`
- `go test -run '^$' -bench BenchmarkListMemoCommentsPreview -benchtime=1x ./server/router/api/v1/test`

### DIFF
--- a/server/router/api/v1/memo_service.go
+++ b/server/router/api/v1/memo_service.go
@@ -312,10 +312,6 @@ func (s *APIV1Service) ListMemos(ctx context.Context, request *v1pb.ListMemosReq
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to list memo creators: %v", err)
 	}
-	instanceMemoRelatedSetting, err = s.Store.GetInstanceMemoRelatedSetting(ctx)
-	if err != nil {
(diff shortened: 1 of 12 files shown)


boojack:
@codex review

PR #5887, opened 2026-04-24
Color change branch

Cammy276:
Modify the color of the interfaces (button & background color).


Cammy276:
Accidentally send wrong pull request. Apologise for that

Issue #5913, opened 2026-04-29
ListUserSettings returns unsupported user_setting rows as generalSetting


mfjt:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this has not already been reported
- [x] This is a bug report, not a question or configuration request

### Memos Version

v0.26.x and later

### Environment Details

- Deployment: Docker
- Database: PostgreSQL

### Steps to Reproduce

1. Run Memos v0.26 or later with PostgreSQL.
2. Sign in as a user.
3. Open Settings.
4. Change the language, for example from English to Japanese.
5. Reload the page.

### Current Behavior

After reloading the page, the UI may still be displayed in English instead of the newly selected language. The selected locale appears to be saved correctly in the database, but the UI does not consistently use it after reload.

Further investigation shows that `ListUserSettings` can return an internal `user_setting` row as a default `generalSetting` before the real `GENERAL` row:

```text
users/example/settings/unknown
  generalSetting:
    locale: "en"
    memoVisibility: "PRIVATE"

users/example/settings/GENERAL
  generalSetting:
    locale: "ja"
```

The frontend currently uses the first `generalSetting` in the response. With PostgreSQL, row order is not guaranteed without `ORDER BY`, so the fake `generalSetting` may be selected instead of the real `GENERAL` setting.

### Expected Behavior

After changing the language in Settings and reloading the page, the UI should be displayed in the newly selected language.

`ListUserSettings` should not return unsupported/internal `user_setting` keys as `generalSetting`.

Keys that are not exposed as API `UserSetting` values, such as `REFRESH_TOKENS`, `PERSONAL_ACCESS_TOKENS`, and possibly `SHORTCUTS`, should either be omitted from the response or handled explicitly with the correct type.

### Logs

```shell

```

### Additional Context

Example database rows:

```text
user_id | key            | value
1       | REFRESH_TOKENS | {...}
1       | GENERAL        | {"locale":"ja"}
```

`REFRESH_TOKENS` appears to be converted to `settings/unknown`, then returned as a default `generalSetting`.

SQLite can also return `settings/unknown` or `settings/SHORTCUTS` as `generalSetting`, but the issue may not be visible if the real `GENERAL` row happens to be returned first.

This makes locale/theme settings appear not to persist even though the correct `GENERAL` row is stored in the database.

PR #5917, opened 2026-04-30
fix(user): omit internal settings from list responses

boojack:
## Summary
- Stop converting internal store-only user setting rows into public `generalSetting` responses.
- Keep the real `GENERAL` setting as the source of truth when listing user settings.
- Add a regression test covering mixed internal and public settings.

## Testing
- `go test ./server/router/api/v1/test -count=1`
- `go test ./server/router/api/v1/... -count=1`

### DIFF
--- a/server/router/api/v1/test/user_setting_test.go
+++ b/server/router/api/v1/test/user_setting_test.go
@@ -1,0 +1,58 @@
+package test
+
+import (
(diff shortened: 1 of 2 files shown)


boojack:
Close https://github.com/usememos/memos/issues/5913

PR #5921, opened 2026-05-01
fix(frontend): use correct url path for memos in sitemap.xml

tokenicrat:
Newer Memos versions use `/memos/<UID>` instead of `/m/<UID>`. Sitemap generation doesn't reflect this change, and URLs result in 404.

This fix replaces `/m/` with `/memos/`.


boojack (approved):
LGTM, good catch!

PR #5925, opened 2026-05-02
feat(memo): create memos on the selected calendar date

boojack:
When a date is active in the activity calendar, the home memo editor now seeds the new memo's createTime/updateTime to that date and reveals the existing TimestampPopover so the user can adjust before saving. Empty in-month calendar cells also become clickable so users can start the first memo for an empty day.

- Add `deriveDefaultCreateTimeFromFilters` helper that turns the active `displayTime` filter into a Date (selected local date + current local hh:mm:ss), with defensive parsing.
- Drop the `count > 0` gate in CalendarCell so empty in-month cells render as buttons; out-of-month cells stay non-interactive.
- Add `defaultCreateTime?: Date` prop on MemoEditor; in create mode, useMemoInit seeds timestamps from the prop and a useEffect re-syncs whenever the prop changes (live-derived from the calendar filter). Edit mode and the comment editor are unaffected by construction.
- PagedMemoList reads MemoFilterContext at the home-editor render site and passes the derived prop.

Frontend-only: the API and store layer already accept custom createTime/updateTime on CreateMemo.


boojack:
@codex review

PR #5951, opened 2026-05-13
feat(preferences): add UI + code font selection via Google Fonts

yazilimalanlari:
Add `ui_font` and `code_font` fields to user general settings so each user can pick a UI sans-serif and a monospaced code font independently. Fonts are loaded directly from Google Fonts' public CSS endpoint — no API key required.

The Preferences > Appearance section gets two new rows: each offers a curated list of popular families (Inter, Bricolage Grotesque, JetBrains Mono, Cascadia Code, Recursive, …) plus a "Custom…" option that accepts any Google Fonts family name. Selections are applied immediately, persisted via `UpdateUserSetting`, and re-applied early on page load (localStorage cache) to avoid a flash of the wrong typeface.

Custom family inputs are validated against an ASCII allowlist so they can't inject CSS or URL syntax into the generated <link>/<style> tags.

---

Screenshot 2026-05-13 at 23 26 09

https://github.com/user-attachments/assets/1048959b-4ad7-4478-81fe-d4e13e941e4e


boojack:
@codex review

PR #5959, opened 2026-05-16
feat: add user setting to always expand long memos

rkilburn:
Add a user scoped preference to always expand the full memo comment in the list of memos

Settings
image

Disabled (default)
image

Enabled
image


boojack:
Refer to https://github.com/usememos/memos/commit/cafa56f1a8c9091062a4b220284e8ca697b050e8

PR #5998, opened 2026-05-30
Add markdown exporter

clnhlzmn:
This adds a simple markdown export background service. See #5988

Adds an optional background runner that mirrors all memos to a directory of markdown files on disk. One-way sync (store → filesystem): new memos become files, edits rewrite in place, deleted/archived memos are pruned.

- Set MEMOS_MARKDOWN_EXPORT_DIR to enable. Runs once at startup, then every 24h.
- Layout: `<exportDir>/<username>/<YYYY-MM-DD>/<uid>.md`
- Each file is YAML frontmatter (uid, created, updated, visibility, pinned, tags, attachments) followed by the memo's raw markdown. Attachments list each file's filename and storage path (relative to the data dir for local storage); attachment bytes themselves aren't copied.

See server/runner/markdownexport/README.md for details.


clnhlzmn:
@johnnyjoygh what do you think about this? Is this a feature that could be merged? Should I go in another direction with it?

boojack:
@clnhlzmn Thanks for this — really nice work. The tests, the determinism, and details like recording the S3 object key instead of the presigned URL all show a lot of care.

That said, I don't think a continuous filesystem-mirror runner is the right fit for memos core:

- **Maintenance surface** — a permanent background runner plus an on-disk format contract the project has to support long-term, for a feature most instances won't enable.
- **Hard to keep neutral** — layout, frontmatter, cadence, and attachment handling all have many reasonable answers; baking one into core invites the rest as feature requests.
- **Prune footgun** — deleting any `.md` it didn't write (including a user's own files if the dir's misconfigured) is a data-loss risk I'd rather not carry in core.
- **Great as an external tool** — this is an ideal standalone CLI/cron sidecar against the existing API.

So I'd rather not merge this as-is, but I'd happily look at a one-shot export built on this rendering logic, or link it from the community tools if you'd prefer to keep the mirror behavior.

PR #6002, opened 2026-06-01
fix(editor): Add Tab/Shift+Tab handling in markdown list editing

An-n-ya:
1. Refactored the list detection logic, changing indent to numbers.
2. Added a utility function for automatically renumbering ordered lists.
3. Extended the Tab/Shift+Tab shortcut to support list indentation/unindentation.
4. Added test cases related to list indentation.

presentation

### DIFF
--- a/web/src/utils/markdown-list-detection.ts
+++ b/web/src/utils/markdown-list-detection.ts
@@ -72,6 +72,7 @@ export function renumberOrderedLists(content: string): string {
     const line = lines[i];
 
     if (line.trim() === "") {
+      counters.clear();
       continue;
     }
 
--- a/web/src/utils/markdown-list-detection.ts
+++ b/web/src/utils/markdown-list-detection.ts
@@ -2,17 +2,14 @@ export interface ListItemInfo {
   type: "task" | "unordered" | "ordered" | null;
   symbol?: string; // For task/unordered lists: "- ", "* ", "+ "
   number?: number; // For ordered lists: 1, 2, 3, etc.
-  indent?: string; // Leading whitespace
+  indent: number; // Whitespace count
 }
 
 // Detect the list item type of the last line before cursor
-export function detectLastListItem(contentBeforeCursor: string): ListItemInfo {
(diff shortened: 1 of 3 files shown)


web/src/utils/markdown-list-detection.ts
```diff
@@ -65,3 +62,37 @@ export function generateListContinuation(listInfo: ListItemInfo): string {
...
+export function renumberOrderedLists(content: string): string {
+  const lines = content.split("\n");
+  const orderedListRegex = /^(\s*)(\d+)([.)])\s+/;
+  const counters: Map<number, number> = new Map();
+
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+
+    if (line.trim() === "") {
+      continue;
```
boojack, line 76: Blank or non-list lines should reset ordered-list counters. Otherwise separate lists get merged, e.g. `1. first\n\n1. second` becomes `1. first\n\n2. second` when renumbering runs. Keep counters through nested child content, but reset at real list boundaries.

boojack (commented):
Found one merge-blocking edge case in ordered-list renumbering.

web/src/utils/markdown-list-detection.ts
An-n-ya: It is reasonable to define blank lines as list boundaries, but not regular non-list lines. Otherwise, it breaks common list patterns like this:
```
1. first
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Etiam facilisis vel arcu a posuere. Nulla et magna nibh. 
2. second
Ut vehicula ornare enim, dictum efficitur libero pharetra nec. Nullam cursus ipsum ac nisl malesuada blandit. Quisque congue, lectus non molestie commodo, mi erat scelerisque nulla, a tempor lectus eros et odio. 
```

An-n-ya:
The newly released WYSIWYG editor performs exceptionally well with Tab/Shift+Tab in #6030 . Handling these operations in the raw mode seems less worthwhile.

PR #6013, opened 2026-06-05
feat: add optional webhook signing secret (Standard Webhooks HMAC-SHA256)

YigesMx:
## Summary

Add optional HMAC-SHA256 request signing for webhooks, following the [Standard Webhooks](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md) convention. The signing secret is never transmitted on the wire — instead, each outgoing request body is signed and the signature is sent in headers. Receivers verify authenticity and payload integrity using the shared secret.

## Changes

| Layer | Summary |
|---|---|
| `proto/store/user_setting.proto`, `proto/api/v1/user_service.proto` | Add `signing_secret` (INPUT_ONLY) and `has_signing_secret` (OUTPUT_ONLY) fields |
| `internal/webhook/webhook.go` | Sign request body with HMAC-SHA256; decode `whsec_` prefix before signing; set `webhook-id`, `webhook-timestamp`, `webhook-signature` headers |
| `internal/webhook/validate.go` | Validate signing secret: must be empty or `whsec_<base64>` decoding to 24–64 bytes |
| `server/router/api/v1/user_service.go` | Create/Update handlers persist and validate signing secret; converter sets `has_signing_secret` boolean |
| `server/router/api/v1/memo_service.go` | Dispatch passes signing secret to the webhook payload |
| `web/src/components/CreateWebhookDialog.tsx` | Signing secret UI: Status/Pending labels + Clear/Generate/Copy icon buttons; secret input removed (sysgen only); edit support with three-state masking |
| `web/src/components/Settings/WebhookSection.tsx` | Added edit button (pencil icon) to each webhook row |
| `web/src/locales/en.json`, `web/src/locales/zh-Hans.json` | i18n strings for new UI elements |

## Design

- **Signing secret is never returned** — `signing_secret` uses `INPUT_ONLY`; only `has_signing_secret` boolean is visible in responses.
- **No database migration** — webhooks are stored as JSON blobs in `user_setting`.
- **System-generated secrets only** — secrets must follow Standard Webhooks convention (`whsec_<base64>`, 24–64 raw bytes). Manual entry is not supported.
- **Three-state edit flow** — `undefined`: don't update; `""`: clear secret; non-empty: set new secret.
- **Replay protection** — each request carries a unique `webhook-id` and unix `webhook-timestamp`, both bound into the HMAC signature.
- **Status indicator** — UI shows whether a webhook has a configured secret (Status) and what will happen on save (Pending).

## Testing

- `buf generate` + `buf lint` — passed
- `go build ./...` — passed
- `pnpm lint` (tsc + biome) — passed
- `pnpm test` (26 files, 121 tests) — all passed

## Verifying Signatures

When a signing secret is configured, each webhook request carries these headers:

```
webhook-id: msg_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
webhook-timestamp: 1712345678
webhook-signature: v1,<base64-encoded-hmac-SHA256>
```

Receivers verify with:

```javascript
const secretRaw = Buffer.from(secret.replace("whsec_", ""), "base64");

const msgId = req.headers["webhook-id"];
const timestamp = req.headers["webhook-timestamp"];
const body = JSON.stringify(req.body);

const expected = crypto
    .createHmac("sha256", secretRaw)
    .update(`${msgId}.${timestamp}.${body}`)
    .digest("base64");

const signature = req.headers["webhook-signature"].split(",")[1];

if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expected))) {
    throw new Error("signature mismatch");
}
```


boojack:
Thanks for adding webhook authentication support. I think we should adjust the design before merging so it follows the common webhook security pattern used by platforms like GitHub, Linear, Stripe, Slack, and the Standard Webhooks spec.

Instead of storing a `bearer_token` and sending the secret itself on every dispatch as `Authorization: Bearer ...`, this should probably be modeled as a webhook signing secret:

- Store/API field: `signing_secret` or `secret` (input-only, never returned)
- On dispatch, sign the exact request body with HMAC-SHA256
- Send signature metadata in headers, for example:
  - `X-Memos-Signature-256: sha256=<hmac>`
  - `X-Memos-Timestamp: <unix timestamp>` for replay protection

This avoids transmitting the shared secret on every webhook request and lets receivers verify both authenticity and payload integrity. If we want to align with the emerging Standard Webhooks convention even more closely, we could use `webhook-id`, `webhook-timestamp`, and `webhook-signature` headers instead.

References for the pattern:
- GitHub: https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries
- Linear: https://linear.app/developers/webhooks
- Stripe: https://docs.stripe.com/webhooks
- Standard Webhooks: https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md

I’d recommend changing this PR from `bearer_token` to a signing-secret based implementation before merge.

YigesMx:
Thanks for the great suggestion! Totally agree that HMAC-SHA256 signing is the right approach here. I will rework this PR to use the Standard Webhooks convention instead of bearer token — signing the request body and sending the signature in webhook-id / webhook-timestamp / webhook-signature headers. Will push the update once it is ready.

YigesMx:
@coderabbitai resume

boojack:
@codex review

boojack:
@codex review

boojack (approved):
LGTM, thanks for the contribution

YigesMx:
Thanks for the review and merge @boojack! Happy to have contributed to memos. I noticed the current UI does not support editing existing webhooks (URL, name, signing secret management), so I will open a follow-up PR for that shortly.

PR #6051, opened 2026-06-26
fix(tags): include combining marks in tag character class

santhoshtr:
Tags written in scripts that use combining marks were truncated at the first mark. Malayalam #കവിത was recognized only as #കവ, leaving the trailing vowel sign and following letters as untagged text.

`web/src/utils/tag-grammar.ts` defines a tag character as a letter (`\p{L}`),
number (`\p{N}`), or symbol (`\p{S}`), but **not** a mark (`\p{M}`). In many
scripts a word interleaves base letters with combining marks:

| Code point | Char | Category | In old class? |
|-----------|------|----------|---------------|
| U+0D15 | ക | Lo | yes |
| U+0D35 | വ | Lo | yes |
| U+0D3F | ി | **Mc** | **no** |
| U+0D24 | ത | Lo | yes |

The lexer stops at the vowel sign. This affects Indic scripts, Arabic/Persian
harakat, Hebrew niqqud, and decomposed (NFD) Latin/Greek/Cyrillic accents.

Add \p{M} to TAG_CHAR_CLASS. Add unit tests too.

| Before | After |
|-----------|-----------|
|image| image |

### DIFF
--- a/web/src/utils/tag-grammar.ts
+++ b/web/src/utils/tag-grammar.ts
@@ -4,10 +4,14 @@
  * tagMarkdown.ts) and the read-only renderer (utils/remark-plugins/remark-tag.ts)
  * so they can't drift.
  *
- * A tag character is any Unicode letter, number, or symbol, plus `_ - / &`.
+ * A tag character is any Unicode letter, mark, number, or symbol, plus
+ * `_ - / &`. The mark class (`\p{M}`) keeps combining marks — Indic vowel
+ * signs, Arabic harakat, Hebrew niqqud, decomposed accents — attached to the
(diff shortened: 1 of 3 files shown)


boojack:
@codex review

santhoshtr:
The codex review is about the policy : How much inclusive or less restrictive we should be about the characters in a tag? IMHO, unless there is any technical or security reasons, we should be inclusive as much as possible. It is quite possible I will have tag #ു-ചിഹ്നങ്ങൾ where I talk about `u` vowel signs of Malayalam. It is a very valid tag and starts with `\p{M}`.

boojack (approved):
LGTM

PR #6065, opened 2026-07-05
feat(web): add untagged shortcut filter example

archit-goyal:
## Summary

- Add an "Untagged" example to the Shortcut filters guide using `size(tags) == 0`.
- Add `size(tags) == ...` to the supported fields list so users can discover tag-count filters in the UI.

## Why

Discussion #6038 asks how to find memos without tags. A maintainer answered that users can create a Shortcut with `size(tags) == 0`. This PR makes that maintainer-recommended recipe available directly in the in-app Shortcut filters examples.

## Verification

- `cd web && pnpm lint`
- `cd web && pnpm test`
- `git diff --check`

### DIFF
--- a/web/src/pages/Shortcuts.tsx
+++ b/web/src/pages/Shortcuts.tsx
@@ -62,6 +62,12 @@ const shortcutExamples = [
     icon: TagsIcon,
   },
   {
+    title: "Untagged",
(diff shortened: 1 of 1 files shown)


boojack (approved):
LGTM

PR #6068, opened 2026-07-06
feat: add group visibility and group selector for memos

gamesushi:
What does this PR do? This PR introduces a new "Group" visibility feature for memos. It allows users to restrict memo visibility to specific groups, expanding beyond the existing Private/Protected/Public options.

Key Changes:

Added GROUP visibility logic to the backend filter schema (internal/filter/schema.go).
Implemented a GroupSelector component in the frontend (MemoEditor) to allow users to select target groups when the visibility is set to Group (value 4).
Integrated the new selector with the updated CodeMirror editor architecture using useEditorSelector.
Why is this needed? To support community and team-based workflows where memos need to be shared only with specific predefined groups of users, rather than all registered users (Protected) or the public.


gamesushi:
Hi all,

I have addressed all the feedback and review comments:

1. **Database Transaction Atomicity**: Wrapped group creation and owner member insertion atomically within a transaction across SQLite, MySQL, and PostgreSQL drivers.
2. **Cascade Handling on Group Deletion**: Prevented orphan memos by explicitly updating their `group_id` to `NULL` when a group is deleted, and updated the latest database schemas with proper foreign key constraint definitions.
3. **Database Migration Fixes**: Corrected MySQL migration syntax for the reserved keyword `groups` and removed default value constraints on TEXT columns. Also fixed the SQLite migration unit test by providing a complete mocked schema.
4. **Backend Optimization & Security**: Moved validation logic outside of loops in `UpdateMemo`, and replaced in-memory post-filtering with CEL database-level filters for `ListMemos`, `ListAllUserStats`, and `GetUserStats`. This secures access and resolves potential pagination issues.
5. **SSE Concurrency Protection**: Fixed a potential data race in the SSE hub subscription by determining and passing user group IDs at client connection initialization.
6. **Frontend Enhancements & Type Safety**: Resolved TS casting issues (e.g., replaced `4 as any` with the generated `Visibility.GROUP` enum) and changed the `GroupSelector` to single-select mode.

Please let me know if there is anything else that needs adjustment. Thank you!

boojack:
@gamesushi Thanks for the contribution. Spaces (#6228, #6230, #6232) now cover the same core use case, so we’re closing this PR as superseded to avoid overlapping models.

PR #6085, opened 2026-07-12
fix: implement pagination for ListUsers API

Iyanuoluwa007:
## Summary

Implement pagination support for `ListUsers`.

### Changes

* Add support for `page_size` with:

  * default value of `50`
  * maximum value of `1000`
* Add support for `page_token` as an offset-based token.
* Return `next_page_token` when additional pages are available.
* Return `InvalidArgument` for malformed or out-of-range page tokens.
* Return paginated users instead of the full user list.
* Preserve `total_size` in the response.

### Notes

Pagination is currently implemented in memory because `store.FindUser` does not yet support database-level pagination or ordering. This can be improved in a future change.

### Tests Added

* `TestListUsersPagination`
* `TestListUsersInvalidPageToken`
* `TestListUsersPageTokenOutOfRange`

### Verification

```bash
go test ./server/router/api/v1/...
go test ./server/router/api/v1/test/...
gofmt -w server/router/api/v1/user_service.go
gofmt -w server/router/api/v1/test/user_service_test.go
```

### DIFF
--- a/server/router/api/v1/test/user_service_test.go
+++ b/server/router/api/v1/test/user_service_test.go
@@ -1,0 +1,98 @@
+package test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	v1pb "github.com/usememos/memos/proto/gen/api/v1"
+)
+
+func TestListUsersPagination(t *testing.T) {
+	ctx := context.Background()
+
+	ts := NewTestService(t)
+	defer ts.Cleanup()
+
+	// Create admin user.
+	admin, err := ts.CreateHostUser(ctx, "admin")
+	require.NoError(t, err)
+
+	adminCtx := ts.CreateUserContext(ctx, admin.ID)
+
+	// Create additional users.
+	for i := 0; i < 5; i++ {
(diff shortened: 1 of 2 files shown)


boojack:
Thanks for tackling this, @Iyanuoluwa007 — `ListUsers` returning everything was a real gap. I've landed it on `main` in `2193d4f7`, aligned with how `ListMemos` already paginates so all `ListXxx` endpoints match:

- **Opaque `PageToken`** (shared `common.go` helpers) instead of a raw integer offset.
- **Dropped `total_size`** — it was unimplemented everywhere (hardcoded `0` or `len(page)`), so it's removed from all six list responses; next-page detection uses a `limit + 1` look-ahead, no `COUNT`.
- **DB-level paging** via `store.FindUser.Offset` instead of loading every user into memory, with an `id DESC` tiebreaker so pages stay stable when `created_ts` ties.
- **Frontend** `useListUsers` now pages through `next_page_token`, so the admin Members view still loads everyone.

Iyanuoluwa007:
Thanks for the review and for landing the more complete implementation. I appreciate the explanation around keeping pagination behavior consistent across the `ListXxx` endpoints. I learned a lot from the use of opaque page tokens and DB-level pagination here.

PR #6089, opened 2026-07-13
fix(web): improve username validation error

### LINKED ISSUE #6075: Improve username validation error messages
### Pre-submission Checklist

### Affected Area

Other

### Reproduced On

Latest release

### Memos Version Tested

Version: canary Commit: dc41480ece7653a4c91142211999a3fd284aa6f5

### Previous Version

_No response_

### Environment Details

- Deployment: Docker
- Host OS: Arch Linux

### Reproduction Frequency

Always

### Steps to Reproduce

1. Open the sign-up page.
2. Enter a username such as word.word.
3. Submit the form.

### Current Behavior

The client toast displays:

```txt
[invalid_argument] rpc error: code = InvalidArgument desc = invalid username: word.word
```

### Expected Behavior

The error should explain the validation rule that was violated. For example:

`Username can only contain letters, numbers, and hyphens.`
`Username cannot consist only of numbers.`
`Username must start and end with a letter or number.`

Alternatively, if there are multiple validation rules, the API could return a structured validation error that the frontend can present in a user-friendly way.

### Regression Information

_No response_

### Logs

_No response_

### Additional Context

_No response_

archit-goyal:
## Summary

- Replace the raw Connect RPC username-validation error on sign-up with clear, translation-ready guidance.
- Keep the backend as the source of truth for username validation instead of duplicating its regular expression in the frontend.
- Preserve the existing fallback behavior for unrelated sign-up errors.

## Root cause

The sign-up page passed every server error to the generic error handler. Connect RPC prefixes its public error message with the transport code, so an invalid username such as `word.word` surfaced as `[invalid_argument] invalid username: word.word` instead of explaining the accepted format.

This change recognizes only the backend's invalid-username response and presents the current rule in user-facing language. It does not broaden the accepted username syntax.

## Verification

- `cd web && pnpm lint`
- `cd web && pnpm test` — 48 test files, 207 tests passed
- `cd web && pnpm build`
- `git diff --check`

Fixes #6075

### DIFF
--- a/web/src/locales/en.json
+++ b/web/src/locales/en.json
@@ -351,6 +351,7 @@
     "failed-to-embed-memo": "Failed to embed memo",
     "fill-all": "Please fill in all fields.",
     "fill-all-required-fields": "Please fill all required fields",
+    "invalid-username": "Username must be 1–36 characters, contain only letters, numbers, or hyphens, start and end with a letter or number, and not contain only numbers.",
     "maximum-upload-size-is": "Maximum allowed upload size is {{size}} MiB",
     "memo-not-found": "Memo not found.",
     "new-password-not-match": "New passwords do not match.",
--- a/web/src/lib/error.ts
+++ b/web/src/lib/error.ts
@@ -1,3 +1,9 @@
+import { Code, ConnectError } from "@connectrpc/connect";
+
+export function isInvalidUsernameError(error: unknown): boolean {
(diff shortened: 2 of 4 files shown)


web/src/locales/en.json
codayon, line 1: This is so tuff 🥀

boojack:
refer to https://github.com/usememos/memos/commit/0bfeb91d50fdd4bafd15c13bb02b3e2810d4b2e4

PR #6097, opened 2026-07-17
fix(web): align radio button group indicator

dedyf5:
This PR fixes a visual alignment issue on the custom Radio Button component. Currently, the checked indicator (the center dot) is misplaced. 

**Screenhosts**

| Before | After |
| -------- | -------- |
| setting-member-create-user-before | setting-member-create-user-after |
| setting-storage-before | setting-storage-after |


boojack (approved):
LGTM, good catch

PR #6103, opened 2026-07-20
fix: keep apostrophes inside tags

### LINKED ISSUE #6078: Apostrophe in tag name
### Pre-submission Checklist

### Affected Area

Web UI

### Reproduced On

Latest release

### Memos Version Tested

neosmemo/memos:stable (docker), Version: 0.29.1

### Previous Version

_No response_

### Environment Details

Server: Ubuntu 26.04 + nginx + docker + mariadb, client: Chrome 149.0.7827.116 on MacOS M1 26.5.
The problem is that in Ukrainian language apostrophe symbol is widely used, for example: сімʼя (family). And when I type such word as a tag name (i.e. #сім'я) then only part of the word becomes a tag:

Image

Image

Image

Happens both via WebUI and memogram (0.5.0).

### Reproduction Frequency

Always

### Steps to Reproduce

1. Add new memo (or edit existing one)
2. Type some text and put some tags in Ukrainian including apostrophe (#сімʼя): Це буде тест #сім'я
3. Hit Ctrl-Enter to save the note
4. Only '#сім' will become a tag.

### Current Behavior

On saving the content with tag incuding apostrophe only part of the text becomes a tag.
This appears both in the note and in the tags list.

### Expected Behavior

The entire text (i.e. "#сімʼя") should be a tag.

### Regression Information

_No response_

### Logs

```shell

```

### Additional Context

_No response_

mayanksaini18:
## What

Fixes #6078 — an apostrophe inside a tag truncates it. `#сім'я` (Ukrainian for *family*) becomes `#сім`, `#O'Brien` becomes `#O`.

## Why

A tag character is validated by Unicode category. The apostrophe is punctuation — `'` (U+0027) is `Po`, and the curly quotes autocorrect produces (`'` U+2019, `'` U+2018) are `Pf`/`Pi` — none of which pass the letter/mark/number/symbol test, so the tag scan stops at the apostrophe.

(Interestingly, the *modifier-letter* apostrophe `ʼ` U+02BC is category `Lm` and already worked via `unicode.IsLetter`, but no keyboard produces it by default.)

## Change

Allow the ASCII apostrophe and the two curly single quotes in both definitions of the tag grammar, keeping backend and frontend in sync:

- **Backend:** `isValidTagRune` in `internal/markdown/parser/tag.go`
- **Frontend:** `TAG_CHAR_CLASS` in `web/src/utils/tag-grammar.ts` (the single source shared by the editor, renderer, autocomplete, and serialize round-trip)

This is consistent with how the grammar already treats `-`, `_`, `/`, `&`.

### Note for reviewers
This means a trailing/possessive apostrophe is now part of the tag (e.g. `#client's` → `client's`), matching the existing behaviour for the other allowed non-alphanumeric characters. If you'd prefer an interior-only rule (apostrophe kept only *between* two tag characters), I'm happy to switch — it's a bigger change on the frontend since it can't be expressed as a plain character-class member.

## Tests
- Go: added apostrophe cases (ASCII, curly, `O'Brien`, trailing-stops-at-space) plus a U+02BC regression guard in `tag_test.go`
- Frontend: added a renderer test in `remark-tag.test.tsx`
- `go test -race ./internal/...` ✅  ·  `cd web && pnpm lint && pnpm test` (244 tests) ✅

### DIFF
--- a/internal/markdown/parser/tag.go
+++ b/internal/markdown/parser/tag.go
@@ -67,6 +67,14 @@ func isValidTagRune(r rune) bool {
 		return true
 	}
 
+	// Allow apostrophes, which occur inside words and names in many languages
+	// (e.g. Ukrainian сім'я, French aujourd'hui, names like O'Brien). Covers the
+	// ASCII apostrophe as well as the curly single quotes produced by autocorrect.
+	// The modifier-letter apostrophe (U+02BC) is already covered by unicode.IsLetter.
+	if r == '\'' || r == '‘' || r == '’' {
+		return true
+	}
+
 	return false
 }
 
--- a/internal/markdown/parser/tag_test.go
+++ b/internal/markdown/parser/tag_test.go
@@ -186,6 +186,36 @@ func TestTagParser(t *testing.T) {
 			expectedTag: "family👨‍👩‍👧‍👦",
 			shouldParse: true,
 		},
+		{
+			name:        "apostrophe inside tag (ASCII ', Ukrainian сім'я)",
+			input:       "#сім'я",
+			expectedTag: "сім'я",
+			shouldParse: true,
+		},
+		{
+			name:        "apostrophe inside tag (curly ’, autocorrected)",
+			input:       "#сім’я",
+			expectedTag: "сім’я",
+			shouldParse: true,
+		},
+		{
+			name:        "apostrophe in name (O'Brien)",
+			input:       "#O'Brien",
+			expectedTag: "O'Brien",
+			shouldParse: true,
+		},
+		{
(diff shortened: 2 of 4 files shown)


mayanksaini18:
Hi @boojack — when you have a moment, could you give this a review? It's a small, focused fix for #6078 (apostrophes truncating tags), with tests on both the Go parser and the shared frontend grammar.

The Backend/Frontend Tests workflows are currently held at `action_required` (the first-time-contributor gate), so they'll need a maintainer's approval to run. CodeRabbit's automated review has already passed. Thanks!

sjorge:
I think this now got covered by https://github.com/usememos/memos/pull/6135 ?

PR #6108, opened 2026-07-23
fix(mcp): accept canonical resource names in path params and keep error results schema-valid

### LINKED ISSUE #6107: MCP: tools with `{memo}` path param reject canonical `memos/{id}` resource names (404); error results violate the declared outputSchema
### Pre-submission Checklist

### Affected Area

API (built-in MCP server)

### Reproduced On

Demo site (v0.30.0-rc.1) and self-hosted v0.30.0-rc.2

### Memos Version Tested

v0.30.0-rc.1 (demo.usememos.com) and v0.30.0-rc.2 (self-hosted Docker)

### Previous Version

_No response_

### Environment Details

This is the rewritten OpenAPI-driven MCP server from #6026. Reproduced unauthenticated against `https://demo.usememos.com/mcp` (Streamable HTTP) with plain curl — no MCP client involved.

**Related issues:** distinct from #6022 / #6037 / #6049 (list tools returning bare arrays), which are fixed in this build.

### Reproduction Frequency

Always

### Steps to Reproduce

```bash
SID=$(curl -sS -D - -o /dev/null -X POST https://demo.usememos.com/mcp \
  -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"repro","version":"1.0"}}}' \
  | grep -i mcp-session-id | tr -d '\r' | awk '{print $2}')

mcp() { curl -sS -X POST https://demo.usememos.com/mcp \
  -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
  -H "Mcp-Session-Id: $SID" -d "$1"; }

# 1. list_memos returns canonical resource names:
mcp '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"memo_list_memos","arguments":{"pageSize":1}}}'
#    -> "name": "memos/webclipper0001"

# 2. Feed that name back to get_memo:
mcp '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"memo_get_memo","arguments":{"memo":"memos/webclipper0001"}}}'
#    -> isError: true, structuredContent: {"error":{"message":"404 Not Found: Not Found"}}

# 3. Same call with the bare ID succeeds:
mcp '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"memo_get_memo","arguments":{"memo":"webclipper0001"}}}'
```

The same full-name → 404 / bare-ID → success behavior applies to `memo_update_memo`, `memo_delete_memo`, `memo_create_memo_comment`, and `memo_list_memo_comments` (verified update/delete on a self-hosted v0.30.0-rc.2 instance with a PAT; for `memo_update_memo` the 404 means the PATCH is **silently never applied**).

### Current Behavior

**1. Resource-name inconsistency.** Every tool that maps to a `/api/v1/memos/{memo}` path substitutes the `memo` argument directly into the URL, so only bare IDs work (`memos/<id>` becomes `/api/v1/memos/memos/<id>` → 404). But the API returns `name: "memos/<id>"` in every memo object, and the tool's input schema describes the param only as `"The memo id."`. An MCP client (human or LLM) that round-trips the `name` it received from `memo_create_memo`/`memo_list_memos` gets a 404 on every follow-up operation. For `memo_update_memo` this is destructive-adjacent: the edit is silently lost.

**2. Error results violate the declared outputSchema.** On REST failure the tool returns `structuredContent: {"error": {"message": "..."}}`, while e.g. `memo_get_memo` declares an `outputSchema` with `required: ["state","content","visibility"]`. Spec-strict clients (Claude Code, LiteLLM, ChatGPT connector) reject the response with a schema-validation error (`data must have required property 'state'…`), masking the underlying 404 and making the tools appear completely broken.

### Expected Behavior

- Tools accept the canonical `memos/{id}` resource name (ideally both forms: strip a `memos/` prefix before path substitution), matching the `name` values the API returns.
- On failure, tools either omit `structuredContent` (relying on `isError` + text content, per MCP spec) or declare an output schema that admits the error shape — so strict clients surface the real error message.

**Workarounds:** pass bare IDs (strip the `memos/` prefix) for any `{memo}` path parameter; read via `memo_list_memos` when a get fails validation.

harpchad:
Fixes #6107

### Problem

The MCP tools substitute path parameters directly into the REST route, so `memo_get_memo` with `memo: "memos/abc123"` requests `/api/v1/memos/memos/abc123` and 404s. But `memos/abc123` is exactly what the API returns as `name` in every memo object — an MCP client that round-trips the name it received from `memo_create_memo`/`memo_list_memos` gets a 404 on every follow-up call, and for `memo_update_memo` the edit is silently lost.

The failure is then hidden by a second issue: on REST errors the tools return `structuredContent: {"error": ...}`, which doesn't conform to the tool's declared `outputSchema` (e.g. required `state`/`content`/`visibility`). Spec-strict clients (Claude Code, LiteLLM, the ChatGPT connector) reject the response with a schema-validation error, masking the underlying 404.

### Changes

- `substitutePathParameters` now routes values through `trimResourceNamePrefix`: when a path parameter's placeholder directly follows its collection segment (`/api/v1/memos/{memo}`) and the value carries that collection prefix (`memos/abc123`), the prefix is stripped. Bare IDs pass through unchanged; foreign prefixes and multi-segment values are left untouched. Works generically for `{memo}`, `{attachment}`, and any future path parameter.
- `newToolErrorResult` no longer attaches `structuredContent`. Error results report through `isError: true` and text content, per the [spec's structured-content guidance](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content) — `structuredContent` is only meaningful when it conforms to the declared output schema. As a bonus, clients now see the real error message (`404 Not Found: memo not found`) instead of a validation failure.

### Testing

- New table-driven test `TestBuildAPIRequestAcceptsResourceNamesForPathParameters` covering canonical names, bare IDs, nested routes, attachment names, and the two left-untouched cases; updated error-shape assertions in `adapter_test.go` and `service_test.go`.
- `go test ./server/router/mcp/...` passes; `go build ./...` clean; `golangci-lint run` reports 0 issues.
- Verified end-to-end against a locally built server over the streamable HTTP endpoint: create → get/update/comment/list-comments/delete all succeed using full `memos/<id>` names (update verified persisted), and a nonexistent-memo call returns `isError: true` with no `structuredContent` and the real 404 message.

### DIFF
--- a/server/router/mcp/adapter.go
+++ b/server/router/mcp/adapter.go
@@ -102,9 +102,34 @@ func substitutePathParameters(operation *openAPIOperation, arguments map[string]
 		if !ok || value == nil || valueToString(value) == "" {
 			return "", errors.Errorf(`missing required path parameter "%s"`, parameter.Name)
 		}
-		path = strings.ReplaceAll(path, "{"+parameter.Name+"}", url.PathEscape(valueToString(value)))
+		id := trimResourceNamePrefix(operation.Path, parameter.Name, valueToString(value))
+		path = strings.ReplaceAll(path, "{"+parameter.Name+"}", url.PathEscape(id))
 	}
 	return path, nil
+}
+
+// trimResourceNamePrefix accepts canonical resource names for path parameters.
+// The API returns names like "memos/abc123", but the REST paths take the bare
+// ID ("/api/v1/memos/{memo}"), so clients that round-trip a returned name
+// would otherwise request "/api/v1/memos/memos/abc123" and get a 404. When the
+// placeholder directly follows its collection segment and the value carries
+// that collection prefix, strip the prefix; bare IDs pass through unchanged.
+func trimResourceNamePrefix(path, parameterName, value string) string {
+	placeholder := "/{" + parameterName + "}"
+	index := strings.Index(path, placeholder)
+	if index < 0 {
+		return value
+	}
+	head := path[:index]
+	collection := head[strings.LastIndex(head, "/")+1:]
+	if collection == "" {
+		return value
+	}
+	id, ok := strings.CutPrefix(value, collection+"/")
+	if !ok || id == "" || strings.Contains(id, "/") {
+		return value
+	}
+	return id
 }
 
 func valueToString(value any) string {
--- a/server/router/mcp/adapter_test.go
+++ b/server/router/mcp/adapter_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/labstack/echo/v5"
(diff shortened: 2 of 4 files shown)


boojack (approved):
LGTM

Issue #6154, opened 2026-08-06
Minor issues after: feat(web): unify navigation in a route-aware sidebar #6144


sjorge:
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this has not already been reported
- [x] I have reproduced this bug on the latest Memos release, current `main`, or the demo site, or this report is specifically about upgrading to the latest release
- [x] This is a bug report, not a question or configuration request

### Affected Area

Web UI

### Reproduced On

Current main branch

### Memos Version Tested

main@5192732cee40020f84c1e57dbcc53ea3d9eaa42f

### Previous Version

v0.30.0

### Environment Details

- Deployment: source
- Database: SQLite
- Browser: Firefox
- Server OS: FreeBSD
- Reverse Proxy: Apache HTTPd
- Related settings: none

### Reproduction Frequency

Always

### Steps to Reproduce

I found some more small issues and instead of bumping the already merged PR with a comment I though I'd file a separate issue to track then.

1. Opening the share menu makes a scroll bar appear

<img width="305" height="174" alt="Image" src="https://github.com/user-attachments/assets/f8435a9b-2a46-4b5f-85df-5cdba82fe666" />

2. The option to auto-collaps the tag tree-mode view is gone (and manual collapsing too)
 
3. navigation based on headers does not handle <ruby> and other tags I'm not actually sure if this is new behavior but it's much more noticeable now with more prominent navigation. (I never noticed the old navigation section 😅 , so this overall is definitely an improvement)

This renders fine in the note:
<img width="349" height="204" alt="Image" src="https://github.com/user-attachments/assets/55333518-05bd-4ad7-adf2-4bff09dde253" />

But not in the navigation:
<img width="297" height="234" alt="Image" src="https://github.com/user-attachments/assets/dfa0ccbd-01a4-48dd-85f7-1fc49630690f" />

```
# <ruby>連濁<rt>れんだく</rt></ruby>: "Why Hito-Bito isn't Hito-Hito"
```

### Current Behavior

See reproducibility

### Expected Behavior

1. since the dropdown menu covers the existing items I don't see why it should add a scrollbar, which triggers a reflow of the content below it.

2. Yes, worked previously and handy when multiple nested levels are used.

3. Ideally it's just rendered correctly, having it satanized is also fine.

### Regression Information

1. new UI so not a regression

2. I'm not sure, it probably always existed looking at the diff in the PR so I don't think it's a regression just more noticeable 

### Logs

```shell
n/a
```

[...]

PR #6159, opened 2026-08-06
fix: align memo outline links with rendered headings

Iams4kura:
## Summary

- Fixes: For the ruby-markup heading reported in item 3 of #6154, the memo outline displays raw <ruby>/<rt> tags and generates a fragment that does not match the rendered heading ID, so outline navigation cannot reach the heading.
- Root cause: extractHeadings recursively treats mdast html marker nodes as visible text, while the rendered HAST heading excludes those markers before rehypeHeadingId computes its ID. The outline therefore derives both its label and slug from a different string than the renderer.

## Regression evidence

- Before: `npx --yes pnpm@11.0.1 test -- tests/markdown-manipulation.test.ts` exited `1`

- After: `npx --yes pnpm@11.0.1 test -- tests/markdown-manipulation.test.ts` exited `0`

## Verification

- `npx --yes pnpm@11.0.1 exec vitest run tests/markdown-manipulation.test.ts`
- `npx --yes pnpm@11.0.1 test`
- `npx --yes pnpm@11.0.1 lint`
- `npx --yes pnpm@11.0.1 build`

## Scope

- 2 files changed, +18 / -0 lines

### DIFF
--- a/web/src/utils/markdown-manipulation.ts
+++ b/web/src/utils/markdown-manipulation.ts
@@ -137,11 +137,13 @@ export function extractHeadings(markdown: string): HeadingItem[] {
 }
 
 interface MdastNode {
+  type?: string;
   value?: string;
   children?: MdastNode[];
 }
 
 function getNodeText(node: MdastNode): string {
+  if (node.type === "html") return "";
   if (node.value) return node.value;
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM, good catch

PR #6235, opened 2026-08-26
Serve raw memo markdown from memo URLs

### LINKED ISSUE #6229: Download markdown directly with .md extension or Accept: text/markdown
### Check before submitting

I went searching for a replacement for Apple Notes, but as I've used Memos more and more I find it more and more useful.

I'd like to be able to just add .md to the end of a URL and be able to give that directly to Ai Agents without the extra step of double-click, copy, paste.

That's all.

Give the raw markdown for .md or Accept: text/markdown.

### Additional context

_No response_

gandazgul:
## Summary
- Serve raw Memo Markdown from `GET /memos/{uid}.md`.
- Serve raw Memo Markdown from `GET /memos/{uid}` when `Accept` explicitly includes `text/markdown` with positive quality.
- Preserve normal React SPA fallback for browser/default and wildcard Accept requests.
- Reuse existing memo read access facts and policy; share tokens do not authorize this representation.

Closes usememos/memos#6229

## Verification
- `go test -v -race ./server/router/api/v1/...` passed.
- `go test -v -race ./server/...` passed.
- `golangci-lint run` passed with v2.13.1 built by Go 1.27.0. The preinstalled local binary was v2.12.2 built with Go 1.26.2 and could not load this repo's Go 1.27 config.
- `go test ./...` failed under default package parallelism because a process was killed during the broad run. `go test -p 1 ./...` passed: 2037 passed, 0 failed.

## Tests added
- Added `server/router/api/v1/memo_markdown_test.go` with 7 new `TestMemoMarkdownRoute*` tests.
- No tests were removed or replaced.


gandazgul:
@bluedbird I'm addressing the feedback. I'll leave the .gitignore changes if you dont mind, else I risk committing machinery that shouldn't be in git.

gandazgul:
Addressed the review feedback in local commit `9aaa1d71`:

- tightened `Accept: text/markdown` parsing so invalid or out-of-range quality values such as `q=2`, `q=NaN`, `q=+Inf`, `q=0`, and malformed values do not select Markdown;
- split authentication handling so invalid credentials resolve to anonymous access, while datastore/auth infrastructure failures propagate and the Markdown route returns a sanitized 500;
- added regression coverage for the Accept quality cases and auth failure/status mapping;
- added comments for the touched functions flagged by the documentation review;
- preserved the `.gitignore` change and removed the temporary plan document.

Verification passed:

- `go test -v -race ./server/auth/...`
- `go test -v -race ./server/router/api/v1 -run 'TestMemoMarkdownRoute'`
- `go test -v -race ./server/router/api/v1/...`
- `go test -v -race ./server/...`
- `go test ./store/db/...`
- `$(go env GOPATH)/bin/golangci-lint run`

gandazgul:
Addressed the CodeRabbit follow-up comments in commit 9f5adbe6:

- Quoted the GOPATH-derived golangci-lint path in the backend workflow and AGENTS.md commands.
- Added the buf format drift check to the RunWield verification command after buf lint.
- Wrapped the memo Markdown current-user auth cause with github.com/pkg/errors.

Validation run:
- ruby YAML parse for .github/workflows/backend-tests.yml
- jq empty .wld/settings.json\n- go test ./server/router/api/v1
- "/Users/gandazgul/go/bin/golangci-lint" run --timeout=3m ./server/router/api/v1/...
- cd proto && buf lint && buf format -d check
- git diff --check for changed files

gandazgul:
Addressed the latest CodeRabbit comment in commit 6c0e0622.

Fix:
- Updated `.wld/settings.json` so `buf format -d` output is captured in `buf_format_output`. A formatter execution failure now stops the verification command before the diff-output check runs.
- Preserved the existing unformatted-proto error message and failure behavior.

Validation run:
- `jq empty .wld/settings.json`
- `cd proto && "$(go env GOPATH)/bin/buf" lint && buf_format_output=$("$(go env GOPATH)/bin/buf" format -d) && if [[ -n "$buf_format_output" ]]; then ...; fi`
- `bash -n <(jq -r '.verification_command' .wld/settings.json)`
- `git diff --check -- .wld/settings.json`

PR #6237, opened 2026-08-26
docs: consolidate domain language

gandazgul:
## Summary

- expand `docs/glossary.md` with shared product and domain terms
- keep a compact central subset in `CONTEXT.md`

Updated glossary terms for your consideration.


gandazgul:
- Addressed the CodeRabbit glossary comment by using one canonical `Memos` definition in both `CONTEXT.md` and `docs/glossary.md`.
- The definition now keeps `Memos` at the product/application level and clarifies that one note is a `memo`.
- Verified the docs-only change with `git diff --check`.

gandazgul:
Follow-up: pushed the fix commit to the PR branch. PR #6237 now points at `3ffd945c` (`docs: align Memos glossary definition`).

PR #6246, opened 2026-08-29
fix(attachment): detect heic/heif mime type from filename extension

### LINKED ISSUE #6174: HEIC files uploaded to S3 as OCTET-STREAM
### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this has not already been requested
- [x] This is a feature request, not a bug report or support question

### Area

Web UI

### Problem or Use Case

HEIC still not supported? Is it because of the ability of browser or memos just has no plan to add such feature?

### Who Would Use This?

Apple users. Android users that would collect HEIC format images from others.

### Proposed Solution

may be do some research and figure this issue out?

### Alternatives Considered

_No response_

### Compatibility and Migration

_No response_

### Examples

_No response_

### Additional Context

_No response_

hktitof:
heic/heif photos from an iPhone or Android upload as `application/octet-stream` on the Docker image (issue #6174), which breaks previews and the content type stored in S3

the chain is simple: browsers report an empty mime type for heic/heif files, the Go builtin mime table has no `.heic`/`.heif` extension, `http.DetectContentType` cant sniff the ISO BMFF container, and the Alpine runtime ships no system mime database, so the attachment ends up as `application/octet-stream`

this adds a small curated fallback for `.heic`/`.heif` before the builtin table and content sniffing, wrapped in a `detectAttachmentMimeType` helper, so the resolved type is identical on machines with and without a system mime database

verified:
- `go test -race ./server/router/api/v1/` passes
- new `TestDetectAttachmentMimeType` fails without the fallback (heic resolves to heif/octet-stream)
- golangci-lint clean
- no behavior change for non-heic files

fixes #6174

assisted by an AI coding agent per the repo AGENTS.md workflow

### DIFF
--- a/server/router/api/v1/attachment_mime_test.go
+++ b/server/router/api/v1/attachment_mime_test.go
@@ -1,0 +1,54 @@
+package v1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestDetectAttachmentMimeType(t *testing.T) {
+	tests := []struct {
+		name     string
+		filename string
+		content  []byte
+		want     string
+	}{
+		{
+			name:     "heic image via extension fallback",
+			filename: "photo.heic",
+			want:     "image/heic",
+		},
+		{
+			name:     "heif image via extension fallback",
+			filename: "photo.heif",
+			want:     "image/heif",
+		},
+		{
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM

PR #6251, opened 2026-08-30
refactor(webhook): iterate hostname labels with SplitSeq

elonfliter:
normalizeHostname only consumes each dot-separated label once. 

Use strings.SplitSeq to validate labels lazily while preserving empty-label, maximum-length, and normalization behavior.


boojack (approved):
LGTM

PR #6274, opened 2026-09-04
fix(i18n): align Chinese sensitive content labels

### LINKED ISSUE #6270: Improvement of show sensitive content label
### Check before submitting

Current the content hidden function has been expanded to normal label, and the hidden might not only because of NSFW, but just for hidden.

The label content for english is "Click to show sensitive content", but for some language, for example Chinese, it's "点击显示 NSFW 内容"，not "点击显示敏感内容". By the way, should we just label it "Click to show content"? It's simpler.

### Additional context

_No response_

RubenPari:
## Summary

- replace outdated NSFW wording with generic sensitive-content terminology in Simplified Chinese
- apply the same terminology consistently to Traditional Chinese
- align the related blur-setting labels in both locales

Fixes #6270

## Testing

- `cd web && pnpm lint`
- `cd web && pnpm test` (149 files, 1186 tests)
- manually verified the blurred-memo reveal label in both Chinese locales at 1280x720
- verified that revealing the memo still removes the blur overlay

### DIFF
--- a/web/src/locales/zh-Hans.json
+++ b/web/src/locales/zh-Hans.json
@@ -263,8 +263,8 @@
   "memo": {
     "archived-at": "归档于",
     "back-to": "返回{{source}}",
-    "click-to-hide-sensitive-content": "点击隐藏 NSFW 内容",
-    "click-to-show-sensitive-content": "点击显示 NSFW 内容",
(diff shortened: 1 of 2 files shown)


boojack (approved):
LGTM

PR #6275, opened 2026-09-04
feat(editor): autofocus Home on initial desktop load

### LINKED ISSUE #6265: Option to auto focus the input field.
### Check before submitting

There could be an option for auto focusing on the input field when the page is open if you want to quickly open the url and write something. 

A setting for autofocusing the input field when the page is opened. 

### Additional context

_No response_

johnnyjoygh:
Opening Home on desktop now focuses the memo editor after restoring its saved draft and cursor, so users can start typing immediately without enabling a preference.

Focus is claimed once per initial Home load. Navigation, Space changes, and opening the global composer cancel pending autofocus; an already-focused control also takes precedence. Mobile keeps tap-to-focus.

Validation: `cd web && pnpm lint` and `cd web && pnpm test` passed (149 test files, 1,194 tests), including coverage for draft/cursor restoration and autofocus eligibility.

Closes #6265.

### DIFF
--- a/web/src/components/MemoEditor/hooks/useMemoInit.ts
+++ b/web/src/components/MemoEditor/hooks/useMemoInit.ts
@@ -9,7 +9,7 @@ interface UseMemoInitOptions {
   memo?: Memo;
   cacheKey?: string;
   username: string;
-  autoFocus?: boolean;
+  autoFocus?: boolean | (() => boolean);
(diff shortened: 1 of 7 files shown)


sjorge:
Screenshot 2026-09-05 at 09 11 51

This is handy, but the cursor looks a bit weird on firefox at least.

PR #6276, opened 2026-09-05
feat(search): add expression mode to Quick Find

### LINKED ISSUE #6198: [Feature Request] Support CEL for regular search
### Check before submitting

Search works OK but sometimes you want to quickly filter some memos with a more complex query which isn't possible currently.

Shortcuts support CEL which does the job but creating a shortcut and then deleting it a few minutes later is rather annoying. So having the search box support CEL would be a good solution.

Allow CEL to be used in the search box.

### Additional context

_No response_

johnnyjoygh:
Quick Find now supports temporary expressions without creating a saved view. Users choose Text or Expression, and can reopen or remove the applied expression from its filter chip. Text searches keep their existing word matching behavior.

Searches retain their route and Space scope, group independent filters so an OR expression cannot bypass other constraints, and preserve expression characters through URL navigation. List and grid results distinguish invalid expressions, access failures, and loading failures, with edit, clear, or retry actions. Invalid requests are not retried automatically, and page failures pause automatic pagination.

Validation:
- `cd web && pnpm lint`
- `cd web && pnpm test` — 150 files, 1,224 tests passed
- `cd web && pnpm build`
- Browser checks at desktop and 390px widths, including list/grid layouts, expression editing, validation errors, and URL reloads

Closes #6198

### DIFF
--- a/web/src/components/AppSidebar/AppSidebar.tsx
+++ b/web/src/components/AppSidebar/AppSidebar.tsx
@@ -35,7 +35,7 @@ import { type AttachmentSection, type InboxFilter, useAppSidebar } from "@/conte
 import { useAuth } from "@/contexts/AuthContext";
 import { useGlobalMemoEditor } from "@/contexts/GlobalMemoEditorContext";
 import { useInstance } from "@/contexts/InstanceContext";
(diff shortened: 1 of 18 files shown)


sjorge:
Awesome works great!

PR #6285, opened 2026-09-07
feat(memos): support moving memos between spaces

johnnyjoygh:
Authors can now assign, move, or unassign a memo from **More → Move to Space**. The dialog confirms placement and visibility together; unassigning a Space-visible memo defaults to Private. Move and Delete are grouped under More to keep uncommon actions out of the main menu.

Comments retain their own placement and visibility, while existing reactions remain attached to the moved memo. Readable comments retain the existing parent identifier, and inaccessible parents show an inline unavailable state without hiding the comment. Cached denials are revalidated when reopening a memo, and guest parent denials are distinguished from transient failures. No API fields or database migrations are added.

Active reaction creators can withdraw their own reactions after losing memo access or Space membership. Relation previews and cached content are removed when a memo becomes unavailable.

### Validation

- Frontend lint and all 1,290 tests pass; production build passes.
- Store tests pass for SQLite, MySQL, and PostgreSQL.
- Targeted server race tests for moves, comments, reactions, and shares pass.
- Proto lint/format checks and diff whitespace checks pass.
- Verified the move flow and the comment's transition to an unavailable parent with separate authenticated users; checked the mobile placeholder.
- Full server race suite has two failures also reproduced on an unchanged checkout: `TestDetectAttachmentMimeType` (macOS MIME registration) and `TestUserWebhookSigningSecretLifecycle` (webhook DNS resolves to a reserved/private address).

### DIFF
--- a/docs/design/multi-spaces.md
+++ b/docs/design/multi-spaces.md
@@ -93,8 +93,8 @@ Distribution is derived rather than separately configured:
 - A Space feed first requires active membership, then returns readable memos assigned to that Space that do not have a `COMMENT` relation. Membership does not make another author's assigned `PRIVATE` memo readable.
 - Direct memo reads evaluate each memo independently, including comments.
 - A comment or conversation query requires its context memo to be readable, then filters every replying memo by that memo's own audience.
-- A `COMMENT` or `REFERENCE` relation and its snippet are returned only when both endpoints are readable.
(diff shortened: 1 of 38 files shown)


sjorge:
This look great, maybe archive also should end up behind the more menu as I guess it will be used about as much as delete?

A quick test seems and no problems so far.

Issue #6309, opened 2026-09-09
[Feature Request] Native Single-Memo and Bulk Export/Import for Easy Instance Migration


Novapixel1010:
### Check before submitting

- [x] I searched existing issues and did not find a duplicate

### What problem would this solve?

Migrating a Memos workspace or restoring data currently requires manual database management (such as running pg_dump / pg_restore or uncompressing server archives via CLI). For users without direct database access, non-technical administrators, or those moving across different database backends (e.g., SQLite to PostgreSQL), migrating data is cumbersome and error-prone.

### What would you like to happen?

1. Single Memo Export: Add an export option (e.g., "Export as Markdown") in the dropdown menu of individual memos.
2. Bulk Export: Add a button in Settings > System/Workspace to export all user memos into a standardized archive (e.g., a .zip containing .md files with frontmatter, or a clean .json export).
3. Bulk Import: Add a corresponding import tool in the UI to seamlessly ingest exported memos into a fresh Memos instance without touching CLI database tools.

### Additional context

Adding UI-level export/import tools would greatly lower the friction for backup retention, data ownership, and instance migrations.

PR #6313, opened 2026-09-11
feat: export personal memos and attachments as a ZIP archive

maximilliangrand:
Adds **My Account → Export Memos → Download ZIP**, so a user can keep a local copy of their notes and attached files without querying each API page or copying server storage.

The archive contains original, unchanged Markdown, a versioned JSON manifest with timestamps/state/visibility/pinning/tags/location and attachment metadata, and an offline Markdown index linking to the exported files. It includes the caller's readable archived notes and comments. Database, local-file, and S3 attachment bytes are included; legacy external attachments retain their URLs in the manifest.

This follows the one-shot export direction discussed with @clnhlzmn in #5998. Refs #6309 for the bulk-export portion; import and single-memo export remain separate work.

### Implementation

- An authenticated native HTTP download route reuses the existing file-server authenticator and S3 resolution. Creator filtering and the store's memo access predicate apply together, including current Space membership. Share tokens and administrator status do not expand the export.
- The server finishes a private temporary ZIP before responding. Missing, unreadable, or size-mismatched attachments fail the request. Cancellation and all exit paths clean up the temporary file; one concurrent export is allowed per server process.
- Note content and attachment blobs are loaded one at a time. Archive paths handle traversal characters, Windows device names, long Unicode filenames, and case-insensitive filesystem collisions; original names remain in the manifest.
- The UI refreshes credentials, disables duplicate submissions, reports failures, and aborts the request when leaving settings.

### Scope

Markdown URLs are preserved verbatim. The offline index and manifest locate exported attachments; external URLs are not fetched. Other users' notes and unlinked uploads are excluded, and comment parent IDs are recorded only when the parent is also exported. This is a user export, not a transactional instance backup or an import format. Server temporary storage and the browser's completed download must accommodate the archive.

### Validation

- `pnpm lint`, `pnpm test`: **1,373 tests passed**, and `pnpm build` passed.
- `golangci-lint v2.13.1 run` passed; `go mod tidy -go=1.27.0` leaves both module files unchanged.
- `go test -v -race ./server/...` found one pre-existing failure: `TestDetectAttachmentMimeType/unknown_extension_falls_back_to_content_sniffing` expects `.xyz` to be unknown, but this macOS installation resolves it to `chemical/x-xyz`. The same test fails identically in an untouched worktree at `7330627`.
- The full server race suite passes with only that subtest excluded using `-skip '^TestDetectAttachmentMimeType$/^unknown_extension_falls_back_to_content_sniffing$'`. The two optional MinIO container tests are skipped by the local environment; the fake-S3 HTTP integration tests run.
- New tests cover archive content and metadata, more than a page of notes/attachments, ownership and Space access, browser-cookie and bearer authentication, comments, external URLs, filename collisions, failed exports, cleanup, cancellation, and concurrent requests. UI tests cover downloads, duplicate clicks, HTTP errors, and aborting on unmount.
- Inspected account settings in the browser. A live local server export was downloaded over HTTP and checked for note content, attachment bytes, manifest, and ZIP CRCs.

### DIFF
--- a/server/router/fileserver/export.go
+++ b/server/router/fileserver/export.go
@@ -1,0 +1,300 @@
+package fileserver
+
+import (
+	"archive/zip"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"slices"
+	"strings"
+	"time"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/labstack/echo/v5"
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/encoding/protojson"
+
+	storepb "github.com/usememos/memos/proto/gen/store"
+	"github.com/usememos/memos/store"
+)
+
+const exportReadme = `# Memos export
+
+Open index.md to browse your notes and their attached files offline.
+
+- memos/<id>-<uid>.md contains the original Markdown, unchanged.
+- attachments/<id>-<uid>/<filename> contains the original attachment bytes.
+- manifest.json (format version 1) records UTC timestamps, state, visibility,
+  pinning, tags, location, and the archive paths and original names of attachments.
+
+This is an export of your own readable memos, including archived memos and comments.
+Other users' notes, unlinked uploads, and notes you can no longer read are excluded.
+Markdown URLs are preserved as written; use index.md or the manifest to find local
+attachments. Legacy external attachments retain their URL in the manifest and
+are not downloaded. Only comment context within this export is recorded. This is
+not an instance backup or an import format. Changes made while an export is being prepared may be reflected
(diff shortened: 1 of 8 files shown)


clnhlzmn:
This is cool.

Since you mentioned my previous PR I'll also point out that since that PR was closed as out of scope for memos core I made an extension app that does the same thing (continuous file system mirror of memos) via the memos API: https://github.com/clnhlzmn/memos-markdown-exporter

I've been using this for a few months now and it's been pretty great.

maximilliangrand:
Closing this as superseded by #6334, which now provides the personal ZIP export workflow together with import and a documented format. Keeping one canonical format and one Settings flow is the right outcome for users. Thanks for reviewing this proposal.

PR #6321, opened 2026-09-13
feat(cli): accept -V and --version alongside the version subcommand

### LINKED ISSUE #6289: Add -V (posix / bsd) and --version (gnu) version flags for old-timers and Ai
### Check before submitting

- `-V` is what BSD / POSIX users expect
- `--version` is what GNU users expect
- `version` is what Go and Rust users expect

What's the harm in having all three? It makes it easier for everyone's muscle memory at no cost. Also useful for Ai, which picks one at random.

Do the same thing for all three.

### Additional context

_No response_

anandghegde:
Closes #6289.

`memos version` was the only way to print the version. This adds `-V` (POSIX/BSD) and `--version` (GNU), all three printing exactly the same string.

```
$ memos version     -> 0.26.1
$ memos --version   -> 0.26.1
$ memos -V          -> 0.26.1
```

## Why it is written this way

Two details that are easy to get wrong with Cobra:

**The flag is registered before Cobra's own.** `InitDefaultVersionFlag` adds a `--version` flag with a lowercase **`-v`** shorthand — not the `-V` the issue asks for. Registering the flag first means Cobra finds it already present and leaves it alone, while still handling it ahead of `RunE`, so `--version` short-circuits instead of starting a server. `-v` stays unclaimed:

```
$ memos -v
ERROR memos failed error="unknown shorthand flag: 'v' in -v"
```

**The version template is set to the bare value.** Cobra's default renders `memos version 0.26.1`, which would not match the subcommand's output. `SetVersionTemplate("{{.Version}}\n")` makes all three identical, which is the point of the issue.

Help text:

```
  -V, --version    print the current Memos version
```

## Verification

Two tests in `cmd/memos/main_test.go`:

- `TestVersionFlagUsesPosixShorthand` — pins the shorthand as `V` and asserts `-v` is unclaimed, so a future change that lets Cobra register its own default is caught.
- `TestVersionFormsPrintTheSameString` — executes `rootCmd` with `--version` and `-V` and compares against `version.GetCurrentVersion()`, the same value the subcommand prints.

Removing the three added lines fails both tests. With them:

```
ok  github.com/usememos/memos/cmd/memos
```

`go test ./cmd/... ./internal/...` passes, `gofmt -l` is empty and `go vet ./cmd/memos/...` is clean. I also checked a version-injected build (`-ldflags -X ...version.Version=0.26.1`) to confirm all three forms report the injected value rather than `dev`.

I did not have `golangci-lint` available locally, so that one command from `AGENTS.md` is unverified on my side; the change is three lines in `init()` plus tests, so I would not expect lint findings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XWr9UtfXxdTdfHMydFh4V8


boojack (approved):
LGTM

PR #6324, opened 2026-09-14
fix(web): reveal metadata row controls on hover

### LINKED ISSUE #6316: [bug] no longer possible to delete attachments
### Check before submitting

### Describe the bug

It's currently no longer possible to delete an attachment

### Steps to reproduce

1. create a new memo
2. add a pdf as attachment
3. save the memo
4. edit the memo
5. try to remove the attached pdf

### Memos version

main@751005bc190e9f68e92bfa0d850f3015bec7febc

### Additional context

Image

anandghegde:
The remove button for attachments in the memo editor never shows up on desktop, so it looks like attachments can no longer be removed.

**Cause**

98d8322 moved the editor rows onto shared helpers in `MetadataSection.tsx`. Their hover and focus classes are built at runtime by `whenEngaged`:

```ts
const whenEngaged = (value: string) => `md:group-hover/row:${value} md:group-focus-within/row:${value}`;
```

Tailwind only generates classes it finds written out in the source, so `md:group-hover/row:opacity-100` and the related classes are never generated. The built CSS has `md:opacity-0` but no `group-hover/row` or `group-focus-within/row` rules. The trailing controls (move up/down, remove) stay invisible on hover and on keyboard focus. The same problem affects the relation and location rows, the file download hint, and the detail rail that should fade out on hover.

**Fix**

Write the class names out in full and remove the helper. No behaviour changes apart from the classes now being generated.

**Verification**

- `pnpm build`: before the fix the CSS had no `group-hover/row` rules. After it, `group-hover/row:opacity-{0,75,100}` and `group-focus-within/row:opacity-{0,75,100}` are all present.
- Headless Chromium against a local backend and `pnpm dev` (fresh data dir and dev server for each run). I created a memo with a PDF attachment, opened it with Edit and hovered the attachment row. The computed opacity of the controls was `0` before the fix and `1` after. After clicking remove and then Update, the memo reloads without the attachment.
- `pnpm lint` and `pnpm test` pass (172 files, 1371 tests).

Before (row hovered):
Before

After (row hovered):
After

Fixes #6316


sjorge:
This does indeed fix the issue for me

boojack (approved):
LGTM

PR #6341, opened 2026-09-17
fix(web): render tel: and sms: markdown links

### LINKED ISSUE #6340: tel: and sms: URI schemes not supported in links.
### Check before submitting

### Describe the bug

Markdown links using the `tel:` and `sms:` URI schemes do not render correctly.

### Steps to reproduce

Write the following markdown in a memo:

```
[phone me](tel:+440000000000)
```

I'd expect this to render as something like:

```
<a href="tel:+440000000000">phone me</a>
```

But it renders as

```
<a>phone me</a>
```

The same bug also affects the `sms:` URI scheme.

### Memos version

v0.31.0-rc.1

### Additional context

_No response_

johnnyjoygh:
Fixes #6340

## Problem

`[phone me](tel:+440000000000)` rendered as `<a>phone me</a>` with no `href`. The same happened for `sms:`.

Two independent allowlists stripped the target:

- react-markdown's default `urlTransform` only accepts `http`, `https`, `irc`, `ircs`, `mailto`, and `xmpp`.
- The rehype-sanitize schema inherited the default `href` protocol list, which has the same set.

## Fix

- Add `tel` and `sms` to the sanitizer's `href` protocols and export a `memoUrlTransform` that passes those two schemes through and defers everything else to `defaultUrlTransform`. Both are driven by one constant so they cannot drift.
- Pass the transform to `ReactMarkdown` in the memo renderer.
- Link previews now only trigger for `http(s)` targets, so a bare `<tel:…>` line renders as a plain link instead of attempting a metadata fetch.

Both schemes hand off to the dialer or messaging app and cannot execute script, so they are in the same class as the already-permitted `mailto:`. `javascript:`, `data:`, and `vbscript:` remain blocked. Scheme matching stays exact-case, mirroring how the sanitizer already treats `https:` vs `HTTPS:`.

## Tests

- `memo-content-security.test.tsx`: `tel:` and `sms:` hrefs survive; script-capable schemes are still stripped.
- `memo-content-paragraph.test.tsx`: handoff links do not get preview cards.

Full web vitest suite passes (178 files, 1416 tests).


felixwatts:
Great work and fast, thanks!

PR #6346, opened 2026-09-18
fix(web): toggle numbered task list items correctly

kalelooz:
Clicking a checkbox in a numbered list currently leaves the task unchanged, although bulk task actions already support numbered lists.

This fixes individual toggling for both `1.` and `1)` markers while preserving numbering, spacing, and line endings. Existing parser-based task selection keeps code blocks untouched.

Validation:
- Added regression tests that fail before the fix and pass afterward.
- All 1,426 frontend tests pass.
- Type checking, lint, production build, and `git diff --check` pass.
- Verified checking, unchecking, and persistence after reload in the local app.

Memos was one of the first apps I self-hosted, and I’ve really enjoyed using it. Thanks for building it and for all the effort you put into maintaining it. It’s great to be able to contribute a small fix back!


boojack (approved):
LGTM

